PR #340, opened 2022-09-24
ComboGeist - randomly juxtapose two notes

bfollington:
Before trying to port my genetic algorithm over I figured it would be smart to take some baby steps. 

The intention of this PR is to introduce the `ComboGeist`, which randomly picks two entries and presents them together in the feed (screenshot attached).

I want to add a `Synthesize` action that would create a new entry _pre-populated_ with the `/slashlinks` to both the entries shown in the feed. I'm not quite sure how to do this as the current flow requires the user to enter a slug for the entry before reaching the editor. That's not a problem in-and-of-itself but we'd need to thread some context through to support the pre-populated editor content.

Keen to hear any suggestions for how to achieve this 😁 

- [x] Register geist
- [x] Display combination of entries
- [x] Synthesize new entry from combination
- [x] Finalise naming scheme

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
@@ -27,6 +27,8 @@ struct StoryComboView: View {
             .frame(height: AppTheme.unit * 11)
             Divider()
             VStack(alignment: .leading, spacing: AppTheme.unit4) {
+                Text(story.prompt)
+                
                 Button(
                     action: {
                         action(story.entryA.link, story.entryA.linkableTitle)
@@ -73,6 +75,7 @@ struct StoryComboView_Previews: PreviewProvider {
     static var previews: some View {
         StoryComboView(
             story: StoryCombo(
+                prompt: "How are these similar?",
                 entryA: EntryStub(
                     SubtextFile(
                         slug: Slug("meme")!,
--- a/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
@@ -29,7 +29,9 @@ struct StoryComboView: View {
             \(first.toSlashlink()) \(second.toSlashlink())
             """
         
-        guard let slug = Slug(formatting: "\(first) \(second)") else { return }
+        guard let slug = Slug(formatting: "\(first) \(second)") else {
+            return
+        }
         let link = EntryLink.init(slug: slug)
         
         action(link, content)
--- a/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
@@ -12,13 +12,24 @@ struct StoryComboView: View {
     var story: StoryCombo
     var action: (EntryLink, String) -> Void
     
+    /// Sort entries by slug
+    /// (this should be factored out eventually)
+    func orderedSlugPair() -> (Slug, Slug) {
+        var entries = [story.entryA.slug, story.entryB.slug]
+        entries.sort(by: { $0.description < $1.description })
+        
+        return (entries[0], entries[1])
+    }
+    
+    /// Construct combined slug and default content, then open editor
     func synthesize() {
-        let slug = "combo/\(story.entryA.slug.description)/\(story.entryB.slug.description)"
+        let (first, second) = orderedSlugPair()
+        let slug = "\(first.description)-\(second.description)"
         let content =
             """
             \(story.prompt)
             
-            \(story.entryA.slug.toSlashlink()) \(story.entryB.slug.toSlashlink())
+            \(first.toSlashlink()) \(second.toSlashlink())
             """
         
         guard let link = EntryLink.init(title: slug) else { return }
--- a/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
@@ -10,8 +10,7 @@ import SwiftUI
 /// A story is a single update within the FeedView
 struct StoryComboView: View {
     var story: StoryCombo
-    var viewAction: (EntryLink) -> Void
-    var synthesizeAction: (EntryLink, EntryLink) -> Void
+    var action: (EntryLink, String) -> Void
 
     var body: some View {
         VStack(alignment: .leading, spacing: 0) {
@@ -30,7 +29,7 @@ struct StoryComboView: View {
             VStack(alignment: .leading, spacing: AppTheme.unit4) {
                 Button(
                     action: {
-                        viewAction(story.entryA.link)
+                        action(story.entryA.link, story.entryA.linkableTitle)
                     },
                     label: {
                         TranscludeView(entry: story.entryA)
@@ -40,7 +39,7 @@ struct StoryComboView: View {
                 
                 Button(
                     action: {
-                        viewAction(story.entryB.link)
+                        action(story.entryB.link, story.entryB.linkableTitle)
                     },
                     label: {
                         TranscludeView(entry: story.entryB)
@@ -55,7 +54,7 @@ struct StoryComboView: View {
                 // pop open a new note with the other two notes already transcluded in it
                 Button(
                     action: {
-                        synthesizeAction(story.entryA.link, story.entryB.link)
+                        action(story.entryA.link, story.entryA.linkableTitle)
                     },
                     label: {
                         Text("Synthesize")
@@ -101,8 +100,7 @@ struct StoryComboView_Previews: PreviewProvider {
                     )
                 )
             ),
-            viewAction: { entryA in },
-            synthesizeAction: { entryA, entryB in }
+            action: { link, fallback in }
         )
     }
 }
--- a/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
@@ -59,7 +59,7 @@ struct StoryComboView: View {
                         action(story.entryA.link, story.entryA.linkableTitle)
                     },
                     label: {
-                        Text("Open")
+                        Text("Create")
                     }
                 )
                 Spacer()
--- a/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
@@ -59,7 +59,7 @@ struct StoryComboView: View {
                         action(story.entryA.link, story.entryA.linkableTitle)
                     },
                     label: {
-                        Text("Synthesize")
+                        Text("Open")
                     }
                 )
                 Spacer()
--- a/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
@@ -24,7 +24,6 @@ struct StoryComboView: View {
     /// Construct combined slug and default content, then open editor
     func synthesize() {
         let (first, second) = orderedSlugPair()
-        let slug = "\(first.description)-\(second.description)"
         let content =
             """
             \(story.prompt)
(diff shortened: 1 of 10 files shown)


gordonbrander:
Amazing! This is exactly how I've been thinking about Geists as well.

There are two main CTAs I think Geists will be able to choose from:

- Open: opens the the note that is the focus of attention for editing
- Create (called synthesize here): creates a new note with a contents specified by the Geist

The distinction between open/create is actually looser than it might appear, since everything in Subconscious operates along a search-or-create flow. So a note will be either re-opened, or created depending upon whether it exists already or not.

Action `DetailAction.loadAndPresentDetail(slug:fallback:autofocus:)` will take care of looking up a note by slug, creating it if it doesn't exist already using `fallback` content, and then open up the editor, optionally autofocusing it. We dispatch this action down to detail from `FeedAction.openStory(EntryLink)` for the open action. https://github.com/subconsciousnetwork/subconscious/blob/main/xcode/Subconscious/Shared/Components/Feed.swift#L189.

We can use the same action for the create action. However, right now entries use the title of the note as fallback content if the note does not exist. For Geists to generate content, we probably want to have the option to include title AND further fallback content. I picture changing the action signature to something like `FeedAction.openStory(link: EntryLink, fallback: String)`.

This might also require a little bit of re-plumbing detail to support both EntryLink and fallback instead of just Slug and fallback.

@bfollington Let me know if you want to find some time on Hangouts to pair program on this. I anticipate it will be a minor change, but may touch a few files.

xcode/Subconscious/Shared/Services/ComboGeist.swift
```diff
@@ -0,0 +1,35 @@
...
+            return nil
+        }
+        return Story.combo(StoryCombo(entryA: stubA, entryB: stubB))
+    }
+}
+
+extension ComboGeist {
+    init(
+        database: DatabaseService,
+        data: Data
```
gordonbrander, line 31: FYI, this form exists to allow us to easily load a JSON Tracery grammar from the app package. If we don't want to include Tracery-generated prompts with the two transcludes, then we don't need it.

However, I would lean toward including a small (possibly Tracery-generated) prompt with the story, just to frame up what you're supposed to do with these two juxtaposed stories.

Possible prompts:

- How might these be connected?
- What similarities can I see between these? Differences? 
- Can I combine these to create something new?
- Can I find three ways in which these might be connected?
- ...other good questions?

gordonbrander:
@bfollington I put together a PR that updates the relevant actions to allow for custom fallback content to be specified https://github.com/subconsciousnetwork/subconscious/pull/341

After merging this PR, `StoryView`'s `action` parameter will support fallback content. Action will dispatch a `DetailAction.loadAndPresentDetail(link:fallback:autofocus:)` which will either open or create a note with custom fallback content, as necessary.

I'm going to hold off merging it into main just yet, since I have to cut a tag for the TestFlight release shortly, but I can merge it into main it shortly after we do the test flight release this coming week. In the mean time, feel free to merge it into your branch.

xcode/Subconscious/Shared/Services/ComboGeist.swift
bfollington: Ah, that makes sense! It did feel like the structure was evaporating once I removed the grammar dependency. I agree that a small prompt is a nice touch, I'll re-introduce tracery and wire that up.

bfollington:
@gordonbrander I rebased on your WIP PR and introduced some basic prompts that appear above the two entries. Thanks for the clarifications, I have a much better understanding of how you're thinking about actions for Geists now.

I said as much on Discord but I'd be keen to pair on getting the rest working, I think I could stumble my way there alone but programming is better together!

xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
```diff
@@ -0,0 +1,109 @@
...
+            Divider()
+            HStack {
+                // TODO: make this actually do something
+                // pop open a new note with the other two notes already transcluded in it
+                Button(
+                    action: {
+                        action(story.entryA.link, story.entryA.linkableTitle)
+                    },
+                    label: {
+                        Text("Open")
```
gordonbrander, line 62: Presumably, we'll synthesize a unique link for the new content?

If so, my sense is that "Create" would be a good CTA here, even though it does the same thing under the hood.

bfollington: Yep! My intention was to create a new link (directly at the action dispatch site? Might make sense as shared utility function later?). I wasn't sure what label you'd prefer so I reverted to "Open" and I left it in a "draw the rest of the owl" state.

I'll push things along when I find a few spare hours this week, and/or if we line up a pairing session we'll be able to smash it out then 🙂

gordonbrander:
I have merged the changes to `DetailAction.loadAndPresentDetail` into main.

bfollington:
@gordonbrander I'm going to try and pick this up on Friday morning (Australia time) but if you have the time / inclination in the meantime feel free to branch off and move things forward.

bfollington:
I took a stab at a naming scheme and format for the fallback content. Any feedback @gordonbrander?

_Moving this from draft to ready for review._

Screen Shot 2022-09-29 at 9 05 43 pm
Screen Shot 2022-09-29 at 9 05 57 pm

gordonbrander:
Geist -> editor flow is looking good.

Something Chris and I have been talking about... When we build the block-based editor, the approach we have planned for transcludes looks like:

- `[[wikilinks]]` are rendered as text links, no transclude
- Bare `/slashlink` or `@bfollington/slashlink` are rendered as a transclude beneath the text block in which it appears.

The rationale is that wikilinks already have a textual representation, whereas slashlinks are more like a pointer.

So in future, the block containing `/subconscious /combogeist` would have rendered underneath it a transclude for `/subconscious`, followed by a transclude for `/combogeist`, in order. I think this is perfect!

xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
```diff
@@ -0,0 +1,119 @@
...
+
+import SwiftUI
+
+/// A story is a single update within the FeedView
+struct StoryComboView: View {
+    var story: StoryCombo
+    var action: (EntryLink, String) -> Void
+    
+    func synthesize() {
+        let slug = "combo/\(story.entryA.slug.description)/\(story.entryB.slug.description)"
```
gordonbrander, line 16: We should probably come up with a convention for Geist-generated content like this. A couple ideas on possible conventions, without anchoring on any specific one:

- I like what you're doing here with `/combo/slug/slug`. OTOH, by default I lean toward flat naming schemes, since they create more chance of name collision (good, since it causes you to return to old notes and add to them https://subconscious.substack.com/p/knowledge-gardening-is-recursive).
- I like the "A-B" conjoining you're doing here. It makes sense that if we already created a combo, we might want to see that content instead of generating new content? How do we feel about `slug-slug` rather than `combo/slug/slug`?
- For Geist-generated content where we want a low chance of collision, maybe we default to `YYYY-MM-DD-name`? Not sure we need that here, though.

bfollington: I feel similarly torn. Nesting under `/combo` avoids polluting the root namespace and helps trace the origin of the entry, but the argument of maximising name collision is a strong one. 

If tracing the origin is actually useful we could also include `source: [[ComboGeist]]` in the content instead of  prefixing with `/combo`? I can see good arguments for `A-B` naming but longer slugs might be difficult to visually parse this way.

Also occurs to me that the same two entries (A and B) could be combined in A-B order or B-A order and I feel intuitively that those should point to the same combined slug? Perhaps instead of blindly concatenating A and B we should sort them them alphanumerically by slug beforehand (or maybe creation date?) I'll push a commit to try this out.

Having a defined ordering on slugs exposes some additional meaning from the `/combo/entry-a/entry-b` format, since viewing `/combo/entry-a` would list all the entries where you've combined `entry-a` with something else. 

e.g.
image

What's our orienting principle here? If we're maximising naming collisions then I vote to drop `combo/` and flatten to `slug-slug` but if we're maximising explorability I would vote to keep it.

gordonbrander: @bfollington let's go with the following:

- Order slugs by alpha
- combine as `slug-slug` (using `Slug(formatting: "\(a) \(b)")` should do the trick)

xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
```diff
@@ -0,0 +1,130 @@
...
+    func synthesize() {
+        let (first, second) = orderedSlugPair()
+        let content =
+            """
+            \(story.prompt)
+            
+            \(first.toSlashlink()) \(second.toSlashlink())
+            """
+        
+        guard let slug = Slug(formatting: "\(first) \(second)") else { return }
```
gordonbrander, line 34: Nit: formatting

```swift
guard let slug = Slug(formatting: "\(first) \(second)") else {
    return
}
```

Not blocking merge though.

xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
```diff
@@ -0,0 +1,130 @@
...
+        let link = EntryLink.init(slug: slug)
+        
+        action(link, content)
+    }
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            HStack(spacing: AppTheme.unit) {
+                Text("@bfollington")
```
gordonbrander, line 43: Note: I hardcoded these a while back just to get something on screen. Before we toggle Geists on we'll want to have some kind of dynamic header... probably the Geist's name.

This is just an FYI. No need to change.

xcode/Subconscious/Shared/Data/Geists/zettelkasten.json
```diff
@@ -21,6 +21,13 @@
       "What questions does this answer?",
       "What new questions does this trigger?"
   ],
+  "combo_followup": [
```
gordonbrander, line 24: @bfollington let's create a separate `combo.json` grammar for the combo geist.

xcode/Subconscious/Shared/SubconsciousApp.swift
```diff
@@ -677,6 +677,12 @@ struct AppEnvironment {
             data: zettelkastenGrammar
         )
         self.feed.register(name: "zettelkasten", geist: zettelkastenGeist)
+        
+        let comboGeist = try! ComboGeist(
```
gordonbrander, line 681: Note: in future, we should probably not force unwrap, and instead only register geist if it returns some value.

FYI only, no need to change.

gordonbrander (changes requested):
Looks great. Just a couple of minor changes requested.

xcode/Subconscious/Shared/Data/Geists/zettelkasten.json
bfollington: Done!

xcode/Subconscious/Shared/SubconsciousApp.swift
bfollington: I took a swing at this, lmk what you think. I'm still learning the languages features of swift so I am likely ignorant of many patterns.

xcode/Subconscious/Shared/Components/Common/StoryComboView.swift
bfollington: Fixed

xcode/Subconscious/Shared/Data/Geists/combo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "combo_followup": [
+      "How might these be connected?",
+      "What similarities can I see between these? Differences?",
+      "Can I combine these to create something new?",
+      "Can I find three ways in which these might be connected?",
+      "What could I add to these to form a set?"
```
gordonbrander, line 7: I love this prompt @bfollington

xcode/Subconscious/Shared/SubconsciousApp.swift
```diff
@@ -668,15 +668,38 @@ struct AppEnvironment {
...
-            resource: Config.default.traceryZettelkasten,
-            withExtension: "json"
-        )
-        let zettelkastenGeist = try! RandomPromptGeist(
-            database: database,
-            data: zettelkastenGrammar
-        )
-        self.feed.register(name: "zettelkasten", geist: zettelkastenGeist)
+        // MARK: zettelkasten geist
+        do {
```
gordonbrander, line 672: TY!

gordonbrander (approved):
Let's goooooooooooooooo!!! 🚀🌔

PR #349, opened 2022-09-30
"On This Day" Geist

bfollington:
I couldn't resist trying my hand at another geist after seeing https://github.com/subconsciousnetwork/subconscious/issues/335, still very WIP. It's hardcoded to show notes from the last 24hrs at the moment.

- [x] Create Geist
- [x] Render output in feed
- [x] Query entries based on date range
- [x] Randomly select date range from list of options (6mo, 1yr etc.)
- [x] Brainstorm list of story variants
- [x] Finalise phrasing for prompt text


xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -825,6 +825,34 @@ struct DatabaseService {
             )
         }
     }
+    
+    func readRandomEntryInDateRange(startDate: Date, endDate: Date) -> EntryStub? {
+        try? database.execute(
+            sql: """
+            SELECT slug, body
+            FROM entry
+            WHERE entry.modified BETWEEN date(?) AND date(?)
```
bfollington, line 834: I tried a few different ways of achieving this, but this felt the most generally flexible for future use.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -825,6 +825,34 @@ struct DatabaseService {
...
+        .compactMap({ row in
+            guard
+                let slugString: String = row.get(0),
+                let slug = Slug(slugString),
+                let body: String = row.get(1)
+            else {
+                return nil
+            }
+            let entry = SubtextFile(slug: slug, content: body)
+            return EntryStub(entry)
```
bfollington, line 852: Kind of ugly to repeat this chunk but I prefer not to hastily abstract mapping code.

xcode/Subconscious/Shared/Services/OnThisDayGeist.swift
```diff
@@ -0,0 +1,26 @@
...
+
+struct OnThisDayGeist: Geist {
+    private let database: DatabaseService
+
+    init(database: DatabaseService) {
+        self.database = database
+    }
+
+    func ask(query: String) -> Story? {
+        guard let yesterday = Calendar.current.date(byAdding: .day, value: -1, to: Date.now) else { return nil }
```
bfollington, line 18: Hardcoded to just the last day at the moment, needs to change.

gordonbrander:
Amazing! I'm so excited about this!!

gordonbrander:
@bfollington if you're still excited to move this PR forward, would love to merge when ready. I am also happy to help and/or adopt it.

bfollington:
@gordonbrander I've had a disrupted past week or so with work, still keen to finish this off. I'll have time tomorrow.

That said, I would like input from you on the time intervals, querying approach and phrasing as the next step anyway!

xcode/Subconscious/Shared/Components/Common/StoryOnThisDayView.swift
```diff
@@ -0,0 +1,105 @@
...
+        case .sixMonthsAgo:
+            return "6 months ago"
+        case .aMonthAgo:
+            return "One month ago"
+        case .inTheLastWeek:
+            return "Modified in the last week"
+        case .inTheLastDay:
+            return "Modified in the last day"
+        }
+    }
```
bfollington, line 29: I had a go at phrasing these in a feed-friendly way but they seem a bit clunky still.

xcode/Subconscious/Shared/Services/OnThisDayGeist.swift
```diff
@@ -0,0 +1,94 @@
...
+    case sixMonthsAgo
+    case aMonthAgo
+    
+    static func random() -> OnThisDayVariant {
+        // https://stackoverflow.com/questions/63189870/return-a-random-case-from-an-enum-in-swift
+        return allCases.randomElement()!
+    }
+}
+
+struct OnThisDayGeist: Geist {
```
bfollington, line 24: This _might_ be better named `MementoGeist` or something similar. `OnThisDay` is clear but the `inTheLastDay` and `inTheLastWeek` cases feel discordant with the name. 

_Maybe_ I'm overthinking it.

bfollington:
@gordonbrander I did a cleanup pass over this, keen to get your review and get another geist merged 🤖

xcode/Subconscious/Shared/Models/StoryOnThisDay.swift
```diff
@@ -0,0 +1,19 @@
...
+//  StoryOnThisDay.swift
+//  Subconscious (iOS)
+//
+//  Created by Ben Follington on 30/9/2022.
+//
+
+import Foundation
+
+/// "on this day" story model
+struct StoryOnThisDay: Hashable, Identifiable, CustomStringConvertible {
```
gordonbrander, line 11: We could use StoryPrompt here and construct the prompt using `OnThisDayVariant` logic within the Geist service. That way we wouldn't need a new view and enum case.

xcode/Subconscious/Shared/Services/OnThisDayGeist.swift
gordonbrander: Memento
Recollect
Remember

I like this idea. Choose your fav!

xcode/Subconscious/Shared/Services/OnThisDayGeist.swift
```diff
@@ -0,0 +1,94 @@
...
+    
+    func addTo(date: Date, component: Calendar.Component, value: Int) -> Date? {
+        guard let d = Calendar.current.date(byAdding: component, value: value, to: date) else {
+            return nil
+        }
+        
+        return d
+    }
+    
+    func dateFromVariant(variant: OnThisDayVariant) -> (Date, Date)? {
```
gordonbrander, line 39: I suggest creating a struct called DateRange. Reason: I've had issues in the past with the Swift compiler crashing because it gets confused trying to track anonymous tuple lifetimes.

```
struct DateRange: Hashable {
    var beginning: Date
    var end: Date
}
```

gordonbrander (changes requested):
Offering the below. I think we can get away with re-using StoryPrompt, and should.

Also we should avoid using an anonymous tuple due to Swift compiler limitations.

Other than these two changes, LGTM!

xcode/Subconscious/Shared/Services/OnThisDayGeist.swift
bfollington: Done and done :)

xcode/Subconscious/Shared/Models/StoryOnThisDay.swift
bfollington: Yep, I was going to ask about this copy-pasting views pattern. Makes perfect sense to use `StoryPrompt` here since it's visually identical, will make the switch and delete the unused code.

gordonbrander (approved):
Looks great. Thanks!

PR #390, opened 2023-01-11
Multiplayer UI components

gordonbrander:
This PR introduces UI components related to multiplayer features.

At this point, the goal is to build out the components, rather than immediately integrating with the app.

- [x] Add audience selector button style
- [x] Profile pic
- [x] Profile pic small (used in backlinks and transcludes)
- [x] Profile pic medium (used in app header) 
- [x] Byline (pfp + name + path)
- [x] Byline sm
- [x] Backlink reacts
- [x] Multiplayer transclude
- [x] Story audience info bar
- [x] Metadata table (with links to IPFS, etc)


gordonbrander:
Stray thought while building these out: there are several views in the source that take complex models as view properties. This makes sense for stateful and effectful components, however it introduces a hard coupling between model and view. 

It may be better to take simple properties, especially for stateless common views. This decouples the view and makes it easier to mock, as well as to reuse without refactoring data layer.

gordonbrander:
Another thought is that my views are overly generic. We should create more task-specific views that come with all their necessary styles. AudienceDisclosureButton is an example.

PR #402, opened 2023-02-20
Add instructions for development on-device

bfollington:
There are a few hoops to jump through when setting up for local device development, may as well document them for future devs.


gordonbrander (approved):
LGTM

Issue #399, opened 2023-02-17
ValidatedTextField error badge uses light BG in dark mode


bfollington:
Spotted during the onboarding flow on-device. 

![image](https://user-images.githubusercontent.com/5009316/219800644-e337e41c-becf-44a6-9239-65bc071dee60.png)

**Expected behaviour**: The error badge icon background should adapt based on the current user theme
**Actual outcome**: The error badge icon background is always white

PR #401, opened 2023-02-20
Fix ValidatedTextField badge background in dark theme

bfollington:
Ensures that the validation error icon will mask the text in the light and dark theme.

Extends https://github.com/subconsciousnetwork/subconscious/pull/398

Fixes #399

PR #403, opened 2023-02-21
Fix ValidatedTextField background (again)

bfollington:
I forgot to update the base from `2023-02-17-color` to `main` before merging https://github.com/subconsciousnetwork/subconscious/pull/401, so it never made it into `main`. I'll blame the early starts for my brain not working.

This is the exact same change again, sorry for the runaround.

Fixes https://github.com/subconsciousnetwork/subconscious/issues/399


gordonbrander (approved):
LGTM

PR #397, opened 2023-02-02
Basic user-facing Noosphere integration

gordonbrander:
This PR gets Noosphere integration up-to-par for end-user use.

## Changes

- Primary ID used throughout app is now `MemoAddress` instead of `Slug`. `MemoAddress` is a slug plus an `Audience`.
- Update database schema
    - memo primary key is now a string representation of `MemoAddress`. This allows us to index and search across both sphere-managed and local content in the same table.
    - In addition, we have a `slug` and `address` key to allow for querying these properties.
- Introduce NoosphereService, a threadsafe manager of `Noosphere` and `SphereFS` instances.
- New DataService methods for
     - Getting sphere identity and version
     - Syncing sphere to gateway
     - Writing sphere state to database
- Improved first run experience
- Introduce `AppDefaults` which offer type-safe getters/setters to `UserDefaults`.
    - We use this to persist sphere ID and some other book-keeping details.
- Update and clean up First Run Experience
- Get settings pane working

## App behavior updates:

- When launching app:
    - Sync sphere with gateway
        - Then sync sphere to database
    - Sync local files with database
 - When pull-to-refresh
     - Sync sphere with gateway
        - Then sync sphere to database
    - Sync local files with database

## QA Checklist

Tested the following after the refactor:

- [x] Fresh start
- [x] Bad Noosphere gateway
- [x] No Noosphere gateway
- [x] Rebuild DB
- [x] Launch First Run Experience
- [x] Relaunch First Run Experience
- [x] Noosphere sync
- [x] File system sync
- [x] Noosphere pull-to-refresh
    - [x] File system pull-to-refresh
- [x] List notes
- [x] Count notes
- [x] Create note
    - [x] Lists refresh
- [x] Open note
- [x] Update note
    - [x] Lists refresh
- [x] Retitle note
    - [x] Lists refresh
- [x] Move note
    - [x] Lists refresh
- [x] Merge note
    - [x] Lists refresh
- [x] Remove note
    - [x] Lists refresh
- [x] Change note audience
    - [x] Lists refresh
- [x] Search notes
- [x] Click link for note
- [x] Click link for note that does not exist
- [x] Search links
- [x] Add link

PR #404, opened 2023-02-21
Update Noosphere package

gordonbrander:
This picks up changes from Noosphere@v0.6.3 which fix the saving/rollback bug we encountered in #397.

The v0.6.3 release branch Package.swift was incorrect, and points to the v0.6.2 build artifact. Therefore, we stay on Noosphere@main, which picks up the v0.6.3 build artifact.

Tested and confirmed that it resolves the saving problem.


cdata (approved):
Nice, I recognize that SHA!

PR #410, opened 2023-02-22
Update backlink style

gordonbrander:
Update backlink style to bring them closer to the Figma design files.

- Introduces Transclude2View, a Transclude style that roughly matches TranscludeView but without the PFP. Until we integrate PFPs with the app and Noosphere, Transclude2View makes a good-enough approximation.
- Also removed BacklinkButtonStyle, since it is no longer used.
- Set transclude borders to 0.5. This matches separators in iOS generally.


xcode/Subconscious/Shared/Components/Common/TranscludeView2.swift
```diff
@@ -0,0 +1,57 @@
...
+//
+//  Created by Gordon Brander on 2/22/23.
+//
+
+import SwiftUI
+
+/// A transitional transclude form that does not have a PFP.
+/// We'll replace uses of Transclude2View with TranscludeView once we integrate
+/// pfps throughout the app.
+struct Transclude2View: View {
```
bfollington, line 13: nitpick: filename does not match component name (`TranscludeView2` vs `Transclude2View`)

bfollington (approved):
Keen to try and render one of these within TextKit 2

xcode/Subconscious/Shared/Components/Common/TranscludeView2.swift
gordonbrander: Fixed. Thanks!

PR #411, opened 2023-02-22
Reset gateway when valid gateway URL is submitted in settings

### LINKED ISSUE #408: Stale Gateway URL is still used for sync after update
Ran into this while dogfooding sync.

# Reproduction 
- Fresh install of the app
- Complete initial setup
- Open settings
- Update gateway URL
- Hit "Sync with Gateway"
- Sync fails
- Logs show the old Gateway URL being used

image

```
[app] Syncing with gateway: http://127.0.0.1:4433
```

# Expected Behaviour
Use the new URL for sync immediately after updating the setting field

# Workaround
Relaunching the app loads the updated Gateway URL

gordonbrander:
Fixes #408 

Changes:

- Call `NoosphereService.resetGateway()` when a valid gateway URL is submitted in settings
- Configure gateway TextField for URLs
    - Autocapitalize off
    - Autocomplete off
    - URL keyboard
- Introduce URL.isHTTP() extension, with tests.
- Limit labeled content in settings (including gateway URL field) to one line

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -431,7 +431,7 @@ struct AppModel: ModelProtocol {
         text: String
     ) -> Update<AppModel> {
         let url = URL(string: text)
-        var isGatewayURLTextFieldValid = url?.isHTTP() ?? false
+        let isGatewayURLTextFieldValid = url?.isHTTP() ?? false
 
         var model = state
         model.gatewayURLTextField = text
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -425,21 +425,17 @@ struct AppModel: ModelProtocol {
         return Update(state: model)
     }
     
-    static func isValidWebURL(_ string: String) -> Bool {
-        guard let url = URL(string: string) else {
-            return false
-        }
-        return url.scheme == "http" || url.scheme == "https"
-    }
-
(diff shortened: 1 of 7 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -448,18 +444,30 @@ struct AppModel: ModelProtocol {
...
+            return Update(state: fallback)
         }
+
         var model = state
         model.gatewayURL = gatewayURL
         model.gatewayURLTextField = gatewayURL
         model.isGatewayURLTextFieldValid = true
+
+        // Persist to UserDefaults
         AppDefaults.standard.gatewayURL = gatewayURL
```
bfollington, line 467: Similar to my clipboard comment the other day, I find myself wondering if `AppDefaults` should also be available via the `AppEnvironment`

bfollington (approved):
LGTM, tested and confirmed the fix

xcode/Subconscious/Shared/Components/AppView.swift
gordonbrander: Fair point. Xcode does expose UserDefaults for unit testing, so we don't need to mock. However, there is a good argument to be made to make this part of environment anyway.

AppDefaults is used in several places. This change would touch a number of other files, so I suggest following up in another PR.

gordonbrander: https://github.com/subconsciousnetwork/subconscious/issues/413

Issue #214, opened 2022-03-15
Block-based editor


gordonbrander:
Basic functionality:
- [x] #881 
- [x] #981 
- [ ] #1042
- [ ] #982
- [x] #983
- [x] #987
- [x] https://github.com/subconsciousnetwork/subconscious/pull/1010
- [x] #1041 
- [ ] #1009 
- [x] #1043

Possible feature enhancements:

- [ ] https://github.com/subconsciousnetwork/subconscious/issues/1039
- [ ] https://github.com/subconsciousnetwork/subconscious/issues/1038

gordonbrander:
Looks like iOS16 has TextKit2 https://developer.apple.com/documentation/uikit/uitextview/4018418-textlayoutmanager

bfollington:
Currently doing discovery on TextKit 2 integration in https://github.com/subconsciousnetwork/ObservableStoreDemo

Once we have found an approach that's viable I'll start integrating this into the main app.

gordonbrander:
Approaching this with a stack of UITextViews in UIKit. See https://github.com/subconsciousnetwork/subconscious/pull/881

PR #421, opened 2023-02-28
TextKit 2 - Basic Integration

bfollington:
The goal of this PR is to integrate the TextKit 2 research I've been doing into the main Subconscious app. For this first step (what I'm calling "Basic Integration") I'm aiming for drop-in replacement of the existing editor with one important difference: per-layout-fragment rendering / formatting.

This should, theoretically, improve the experience of working on large documents quite a bit!

Lays the foundation for #214


xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -123,16 +123,34 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
                 )
             )
         }
+        
+        // This allows us to intercept touches on embedded transclude blocks
+        override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+            guard let touch = touches.first else { return }
+            let tapPoint = touch.location(in: self)
+            
+            let textContentStorage = self.textLayoutManager?.textContentManager as! NSTextContentStorage
```
bfollington, line 132: Might be worth passing a reference to `NSTextContentStorage` explicitly, instead of this assert-y, cast-y approach.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -123,16 +123,34 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
     //  MARK: Coordinator
-    class Coordinator: NSObject, UITextViewDelegate, NSTextStorageDelegate {
+    class Coordinator: NSObject, UITextViewDelegate, NSTextContentStorageDelegate, NSTextContentManagerDelegate, NSTextLayoutManagerDelegate {
         /// Is event happening during updateUIView?
         /// Used to avoid setting properties in events during view updates, as
         /// that would cause feedback cycles where an update triggers an event,
         /// which triggers an update, which triggers an event, etc.
         var isUIViewUpdating: Bool
         var representable: MarkupTextViewRepresentable
+        var renderTranscludeBlocks: Bool = false
```
bfollington, line 153: Gate rendering of embedded transclude blocks, they work / render mostly as expected but there's too many rough edges to fix for this initial PR.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -141,37 +159,6 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
-        /// NSTextStorageDelegate method
-        /// Handle markup rendering, just before processEditing is fired.
-        /// It is important that we render markup in `willProcessEditing`
-        /// because it happens BEFORE font substitution. Rendering before font
-        /// substitution gives the OS a chance to replace fonts for things like
-        /// Emoji or Unicode characters when your font does not support them.
-        /// See:
-        /// https://github.com/gordonbrander/subconscious/wiki/nstextstorage-font-substitution-and-missing-text
-        ///
-        /// 2022-03-17 Gordon Brander
```
bfollington, line 153: I _believe_ we no longer need to work via `NSTextStorage` at all anymore. We can render markup ephemerally without altering the underlying storage.

This comment had me worried that we'd lose support for proper unicode / emoji rendering but it seems to be behaving, please advise of any other edgecases to check.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -243,6 +230,39 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
+        
+        // MARK: - NSTextLayoutManagerDelegate
+                                
+        func textLayoutManager(_ textLayoutManager: NSTextLayoutManager,
+                               textLayoutFragmentFor location: NSTextLocation,
+                               in textElement: NSTextElement) -> NSTextLayoutFragment {
+            
+            // TODO: might be better to hold a ref to textContentStorage somewhere
+            // textContentStorage is a concrete implementation of textContentManager
+            let textContentStorage = textLayoutManager.textContentManager as! NSTextContentStorage
```
bfollington, line 242: Same as before, might be nice to have an explicit reference to `textContentStorage` in scope instead of this

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -243,6 +230,39 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
+        
+        // MARK: - NSTextContentStorageDelegate
+        
+        func textContentStorage(_ textContentStorage: NSTextContentStorage, textParagraphWith range: NSRange) -> NSTextParagraph? {
+            // In this method, we'll inject some attributes for display, without modifying the text storage directly.
+            let originalText = textContentStorage.textStorage!.attributedSubstring(from: range)
+            let textWithDisplayAttributes = NSMutableAttributedString(attributedString: originalText)
+            
+            self.representable.renderAttributesOf(textWithDisplayAttributes)
+            return NSTextParagraph(attributedString: textWithDisplayAttributes)
```
bfollington, line 264: This callback should be invoked only for `NSTextParagraph`s that have changed, massively speeding up rendering the attibutes 🎊 

(`NSTextParagraphs` roughly correlate to `NSTextLayoutFragments` 1:1, with some exceptions)

xcode/Subconscious/Shared/Components/Common/Text/TranscludeBlockLayoutFragment.swift
```diff
@@ -0,0 +1,155 @@
...
+            layer.render(in: rendererContext.cgContext)
+            
+            rendererContext.cgContext.restoreGState()
+        }
+        
+        return img
+    }
+}
+
+class TranscludeBlockLayoutFragment: NSTextLayoutFragment {
```
bfollington, line 67: Ported verbatim from my other repo

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
gordonbrander: This may be a TK1 specific issue. The best thing to do is playtest on-device. Watch for issues with autocompletion, autocorrection, emoji, and emoji substitution.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
gordonbrander: Yes, Let's avoid `!`, since they crash the app. Better to have a guard clause and no-op + `logger.warning` if we can't unwwrap.

```swift
guard let textContentStorage = self.textLayoutManager.textContentManager as? NSTextContentStorage else {
    MarkupTextViewRepresentable.logger.warning("Could not access textContentStorage")
    return
}
```

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -243,6 +230,39 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
+                               textLayoutFragmentFor location: NSTextLocation,
+                               in textElement: NSTextElement) -> NSTextLayoutFragment {
+            
+            // TODO: might be better to hold a ref to textContentStorage somewhere
+            // textContentStorage is a concrete implementation of textContentManager
+            let textContentStorage = textLayoutManager.textContentManager as! NSTextContentStorage
+            let content = textContentStorage.attributedString(for: textElement)
+            
+            // Where we decide which layout/rendering implementation to use per TextLayoutFragment
+            if renderTranscludeBlocks && content?.string.contains("/slashlink") ?? false {
```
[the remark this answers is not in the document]
gordonbrander: But let's remove this placeholder logic before landing.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
gordonbrander: Let's log here.

```
MarkupTextViewRepresentable.logger.debug(
    "textContentStorage: render markup attributes"
)
```

MarkupViewRepresentable contains a lot of tricky lifecycle logic, and I like to have logging in key areas such as rendering, so we can easily verify that e.g. rendering happens only once per keystroke.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -243,6 +230,39 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
+            } else {
+                return NSTextLayoutFragment(textElement: textElement, range: textElement.elementRange)
+            }
+        }
+        
+        // MARK: - NSTextContentStorageDelegate
+        
+        func textContentStorage(_ textContentStorage: NSTextContentStorage, textParagraphWith range: NSRange) -> NSTextParagraph? {
+            // In this method, we'll inject some attributes for display, without modifying the text storage directly.
+            let originalText = textContentStorage.textStorage!.attributedSubstring(from: range)
```
gordonbrander, line 260: Let's guard and log:

```swift
guard let originalText = textContentStorage.textStorage?.attributedSubstring(from: range) else {
    MarkupTextViewRepresentable.logger.warning("textContentStorage: could not access attributedSubstring")
    return
}
```

xcode/Subconscious/Shared/Components/Common/Text/TranscludeBlockLayoutFragment.swift
```diff
@@ -0,0 +1,155 @@
+//
```
gordonbrander, line 1: Nit: should we create a `Subconscious/Shared/Components/Common/Transclude/` directory and place this and other transclude code in it?

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -123,16 +123,34 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
         }
+        
+        // This allows us to intercept touches on embedded transclude blocks
+        override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+            guard let touch = touches.first else { return }
+            let tapPoint = touch.location(in: self)
+            
+            let textContentStorage = self.textLayoutManager?.textContentManager as! NSTextContentStorage
+            
+            if let layoutFragment = textLayoutManager!.textLayoutFragment(for: tapPoint) {
```
gordonbrander, line 134: Let's guard and log.

```swift
guard let textLayoutManager = textLayoutManager else {
    MarkupTextViewRepresentable.logger.warning("touchesBegan: Could not access textLayoutManager")
    return
}
if let layoutFragment = textLayoutManager.textLayoutFragment(for: tapPoint) {
}
```

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -123,16 +123,34 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
+                let content = textContentStorage.attributedString(for: layoutFragment.textElement!)
```
gordonbrander, line 135: Let's guard and log.

gordonbrander (changes requested):
Remarkable progress! I'm also surprised by how little we actually had to replace/rewrite. 

Added a few minor change requests. Mostly let's get rid of all `!` and replace with `if let`, or `guard` on failure. Instead of panicking via `!` we should no-op, and log warnings where appropriate.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeBlockLayoutFragment.swift
```diff
@@ -0,0 +1,158 @@
...
+        // Host our SwiftUI view within a UIKit view
+        let hosted = UIHostingController(rootView: v)
+        guard let view = hosted.view else {
+            return
+        }
+        
+        view.translatesAutoresizingMaskIntoConstraints = false
+        
+        // We have to mount the view before it will actually do layout calculations
+        UIApplication.shared.windows.first?.rootViewController?.view.addSubview(hosted.view)
```
bfollington, line 90: Meant to call attention to this line before, without mounting the `UIHostingController` the view cannot be rendered into an image but... this line is gross and yields the a warning: `'windows' was deprecated in iOS 15.0: Use UIWindowScene.windows on a relevant window scene instead`

I don't know what a "relevant window scene" refers to from this context.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -243,6 +239,52 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
...
+            guard let textContentStorage = textLayoutManager.textContentManager as? NSTextContentStorage else {
+                MarkupTextViewRepresentable.logger.warning("Could not access textContentStorage")
+                return baseLayoutFragment
+            }
+            
+            if let text = textContentStorage.attributedString(for: textElement)?.string {
+                let sub = Subtext(markup: text)
+                
+                // Only render transcludes for a single slug in a single block
+                if renderTranscludeBlocks && sub.slugs.count == 1 && sub.blocks.count == 1 {
```
bfollington, line 261: @gordonbrander I'm not sure if this is the usage of `Subtext` you were thinking of? Checking for a single slug and a single block isn't actually a sufficient heuristic for transcludes still, we need the link to be the only text in the block, correct?

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
bfollington: I tested this on-device and everything seemed well behaved to me.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
```diff
@@ -123,16 +123,43 @@ struct MarkupTextViewRepresentable: UIViewRepresentable {
                 )
             )
         }
+        
+        // This allows us to intercept touches on embedded transclude blocks
+        override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
```
gordonbrander, line 128: Just realized we need to be calling super for every logical branch that is failing or not meant to interecpt a transclude tap. That way we fall back to default touchesBegan behavior upon failure.

Perhaps we should remove this logic in this PR, since it is intended to be a no-op for now anyway.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
gordonbrander: I'm realizing two things:

- The way we visualize Transcludes working is that wikilinks do not generate transcludes, while slashlinks DO generate transcludes.
    - Therefore, we want to be able to check for slashlink slugs, specifically. This feature does not currently exist in Subtext. We'll have to add it.
- It would be best to parse Subtext once, when block changes, and keep that instance around for reference in this delegate method, and others.
    - Currently MarkupViewRepresentable doesn't know about Subtext. Instead we have a `renderAttributesOf` delegate, which takes a mutable attributed string and applies attributes. The intention was to make it generic over many markups.
    - However, it may simplify things to complect MarkupTextViewRepresentable with Subtext and call it SubtextTextViewRepresentable instead. This would allow us to make Subtext domain logic available in more places in the implementation. E.g. keep a parsed Subtext instance on the custom textview instance.

All that said, I think we should remove this logic for now, land this PR, then sketch out how we want to approach bringing in transcludes in a follow-up PR.

gordonbrander (changes requested):

gordonbrander (changes requested):

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
bfollington: As discussed on call, this looks like the right approach going forward. I will remove the transclude detection from this PR altogether and we'll land it via #424 instead.

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
bfollington: Removed this code, we'll add it back when we need it 👍

xcode/Subconscious/Shared/Components/Common/MarkupTextViewRepresentable.swift
bfollington: I've removed all the transclude related code from this PR.

gordonbrander (approved):

PR #419, opened 2023-02-24
Multiplayer @slashlink integration

gordonbrander:
This PR integrates `Slashlink` with `MemoAddress.public`.

Slashlink is a type that represents a `@petname/slug` or `/slug` slashlink type. Updating `MemoAddress` in this way allows us to address memos that belong to other spheres.

Prep work for #385. While this PR integrates the `Slashlink` type throughout the app, it does not expose `@petname/foo` markup in the editor yet. We will save that for a followup PR.

## Key changes

- Update `Slashlink` type for parsing correctness, and add numerous tests.
    - Validate via Regular Expression for improved correctness
    - Allow unicode word characters
- Update `Slug` type
    - Validate via Regular Expression for improved correctness
    - Allow unicode word characters
- Introduce `Petname` type, which represents a valid petname/nickname.
- Update `MemoAddress`
    - Public now holds a slashlink e.g. `MemoAddress.public(Slashlink)`
    - `MemoAddress.description` serializes as `public::@foo/bar`, `public::/bar`, `local::/bar`, etc.
- Allow reading from `@petname/foo` addresses for public memos

PR #422, opened 2023-02-28
Parse @petname/slashlinks in Subtext

### LINKED ISSUE #385: Parse @links
Noosphere will support `@handle/link` addressing shortly. We need to support this in the parser.

Implement for slashlinks, but not wikilinks.

gordonbrander:
Introduces parsing for `@petname/slashlinks` in Subtext.

Fixes #385

Followup to #419.

Note that parsed slashlinks are simply identified in markup, not validated for correctness. To validate, parse through `Slashlink`.

### DIFF
--- a/xcode/Subconscious/Shared/Library/CollectionUtilities.swift
+++ b/xcode/Subconscious/Shared/Library/CollectionUtilities.swift
@@ -10,7 +10,10 @@ import Foundation
 extension Collection {
     /// A safe bounds-checked array index lookup
     func get(_ index: Index) -> Element? {
-        indices.contains(index) ? self[index] : nil
+        guard index >= startIndex && index < endIndex else {
+            return nil
+        }
+        return self[index]
     }
 }
 
--- a/xcode/Subconscious/Shared/Parsers/Tape.swift
+++ b/xcode/Subconscious/Shared/Parsers/Tape.swift
@@ -81,7 +81,7 @@ struct Tape {
         return element
     }
 
-    /// Peek forward, and consume if match
+    /// Peek forward, and consume if substrings match
     mutating func consumeMatch(_ subsequence: Substring) -> Bool {
         if let endIndex = offset(by: subsequence.count) {
             if rest[currentIndex..<endIndex] == subsequence {
@@ -90,6 +90,20 @@ struct Tape {
             }
         }
         return false
+    }
+    
+    /// Prefix-match a regular expression, and consume if match.
+    /// - Returns match object
+    mutating func consumeMatch<Pattern, Output>(
+        _ regex: Pattern
+    ) -> Bool
+    where Pattern: RegexComponent<Output>
+    {
+        guard let match = rest.prefixMatch(of: regex) else {
+            return false
+        }
+        self.currentIndex = match.range.upperBound
+        return true
     }
 
     /// Get an item offset by `offset` from the `currentIndex`.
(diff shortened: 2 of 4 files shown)


xcode/Subconscious/SubconsciousTests/Tests_Subtext.swift
```diff
@@ -269,6 +269,54 @@ class Tests_Subtext: XCTestCase {
...
+            XCTFail("Expected slashlink")
+            return
+        }
+        
+        XCTAssertEqual(
+            String(describing: slashlink),
+            "@PETNAME/deep//slashlink/that-is-technically-invalid",
+            "Slashlinks do not parse for valididty of body"
+        )
+    }
```
bfollington, line 318: Could use a test case that looks for an `@petname` that isn't at the start of the string, i.e.

```swift
let markup = "What about @petname/nested/link in the middle."
```

bfollington (changes requested):
Unless there's a good reason I'm missing, I think we should add the extra test cases?

xcode/Subconscious/Shared/Library/CollectionUtilities.swift
```diff
@@ -10,7 +10,10 @@ import Foundation
 extension Collection {
     /// A safe bounds-checked array index lookup
     func get(_ index: Index) -> Element? {
-        indices.contains(index) ? self[index] : nil
+        guard index >= startIndex && index < endIndex else {
```
gordonbrander, line 13: While fixing test code using this bounds-checked arraay getter, I realized we could be doing bounds checking more efficiently.

`indices.contains` does a loop over the array (ostensibly). but since indexes are comparable, all we have to do is check if index is between start (inclusive) and end (exclusive)

xcode/Subconscious/Shared/Parsers/Tape.swift
```diff
@@ -91,6 +91,20 @@ struct Tape {
         }
         return false
     }
+    
+    /// Prefix-match a regular expression, and consume if match.
+    /// - Returns match object
+    mutating func consumeMatch<Pattern, Output>(
```
gordonbrander, line 97: Introduce `Tape.consumeMatch(_ regex:)`, which allows us to look ahead with a regex, and consume tape up until the end of the match if successful.

This single utility lets us factor away a large part of the Subtext parser.

xcode/Subconscious/SubconsciousTests/Tests_Subtext.swift
```diff
@@ -81,7 +81,7 @@ class Tests_Subtext: XCTestCase {
         let markup = "Some text with a http://example.com link"
         let dom = Subtext.parse(markup: markup)
         
-        guard case let .link(link) = dom.blocks[0].inline[0] else {
+        guard case let .link(link) = dom.blocks.get(0)?.inline.get(0) else {
```
gordonbrander, line 84: Fix tests using bounds-checked array getters. Otherwise tests panic on fail.

gordonbrander:
@bfollington while in there adding tests, I realized I could leverage Swift's new Regex library to significantly simplify the Subtext parser.

https://github.com/subconsciousnetwork/subconscious/pull/422/commits/e7f95dd082d5de12e5010088693c4b76aeefc334 introduces `Tape.consumeMatch(_ regex:)` which allows us to radically simplify the parser by taking advantage of Swift's (new) built in regular expression library to lookahead and consume if matching.

Note that parser is still single-pass. We're just using regex to eliminate several arms of the recursive descent parser, which can be expressed as regex instead.

- Added additional Subtext tests you requested.
    - All tests passing.
- Playtested on device

bfollington (approved):

PR #426, opened 2023-03-02
Gateway Sync Status Indicator

bfollington:
There's currently no UI feedback for the gateway sync making it hard to tell if it worked as a user. This PR tracks the status of the sync operation in the `AppModel` and uses that to display a status icon next to the sync button in the `SettingsView`.

The `.inProgress` state is animated, because it's the little things...

Screenshot 2023-03-02 at 12 08 20 pm
Screenshot 2023-03-02 at 12 08 05 pm

https://user-images.githubusercontent.com/5009316/222312511-cdf83d69-9451-4f4f-99f3-a09ae639ab09.mov

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -178,11 +178,11 @@ enum AppDatabaseState {
     case ready
 }
 
-enum GatewaySyncStatus {
+enum GatewaySyncStatus: Equatable {
     case initial
     case inProgress
     case success
-    case failure
+    case failure(String)
 }
 
 //  MARK: Model
@@ -760,7 +760,7 @@ struct AppModel: ModelProtocol {
         logger.log("Sphere failed to sync with gateway: \(error)")
         
         var model = state
-        model.lastGatewaySyncStatus = .failure
+        model.lastGatewaySyncStatus = .failure(error)
         
         return update(
             state: model,
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -178,6 +178,13 @@ enum AppDatabaseState {
     case ready
 }
 
+enum GatewaySyncStatus {
+    case initial
+    case inProgress
+    case success
+    case failure
+}
+
 //  MARK: Model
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -178,6 +178,13 @@ enum AppDatabaseState {
     case ready
 }
 
+enum GatewaySyncStatus {
+    case initial
+    case inProgress
+    case success
+    case failure
```
bfollington, line 185: Might be worth capturing the failure reason here?

xcode/Subconscious/Shared/Components/Settings/SettingsView.swift
```diff
@@ -8,10 +8,39 @@
...
+                    .onAppear() {
+                        self.spin = true
+                    }
+            case .success:
+                Image(systemName: "checkmark.circle")
+            case .failure:
+                Image(systemName: "exclamationmark.arrow.triangle.2.circlepath")
+            }
+            
+            Text("Sync with Gateway")
```
bfollington, line 35: Perhaps the label should also change? Might be a bit visually jarring

gordonbrander: I think the iOS way to do this is usually to have the spinner on the right-hand side of the table cell (secondary icon color).

bfollington: Updated the appearance, it uses secondary color by default but uses the accent color when actively animating.

image

Screenshot 2023-03-03 at 10 53 31 am

Screenshot 2023-03-03 at 10 52 52 am

bfollington: Also tracking the error message now, and rendering a special failure case:

image

bfollington:
> @bfollington looking good.
> 
> 
> 
> Optional:
> 
> 
> 
> The failure case doesn't read to me as a button though. What if we put a bit of text underneath the button row displayed last sync time OR error message?
> 
> 
> 
> Additionally, I might keep the sync button text the same and rely on icon / sync status text to do the work.

Let's follow up with more improvements in a subsequent PR.

PR #427, opened 2023-03-02
Editor draft state

### LINKED ISSUE #362: Omnibox
Implement omnibox-style search UI (see designs)

- Provenance of note
- Search across follows (see #363)
- Anything else?

Epic #330

### LINKED ISSUE #406: Introduce intermediate draft state for new notes
Currently, there is no explicit draft state for new notes. Notes must have a slug and title to be loaded into the editor.

Introducing an intermediate draft state for new notes would allow us to improve a number of interactions within the app:

- "Quick create" could create a note without requiring a search/title first.
- Titles could be derived from body content automatically.
- Publish to Noosphere could become default, because draft state would allow user to toggle local-only before saving.

This might be done as part of a less-ephemeral edit mode, which could allow us to make cleaner change histories via specific snapshots invoked by exiting edit mode (see Figma explorations). Note to self: SwiftUI toolbars do not support animations between toolbar content at this time. However, it is worth noting that Apple Notes doesn't even animate between the view and edit modes of the toolbar.

gordonbrander:
Fixes #406
Fixes #362

## Changes

- Public/private note suggestions
- Notes can be created without address or content. Address is derived from content during first autosave (~20s).
- Omnibox
- Note meta bottom sheet for actions like change link, delete, etc.
    - In future, this will also contain links to note history, and other note details like IPFS hash.
- Removed concept of title. Instead, we derive title from body content using `string.title()` extension.
    - This significantly simplifies a number of concepts in the code and the UX.
    - Title can change (by changing note body) without changing the link
    - Changing the link is a distinct action.
        - Good because it breaks links.
        - This will give us an opportunity in future to re-write slugs in all notes.

## QA checklist

Go through app and make sure the following flows work:

- [x] Create public note with fallback text
- [x] Create private note with fallback text
- [x] Create public note without fallback text
    - [x] Create public note without fallback text, conflicting slug
- [x] Create private note without fallback text
    - [x] Create private note without fallback text, conflicting slug
- [x] Autosave without text (no-op)
- [x] Autosave with text
    - [x] Derives slug from text
    - [x] Sets slug on omnibox
    - [x] Sets slug on bottom sheet
    - [x] Sets slug on rename sheet
- [x] Save on back
- [x] Save on unfocus
- [x] Change note audience
- [x] Change note link
    - [x] Merge note
    - [x] Move note
- [x] Delete note


xcode/Subconscious/Shared/Components/Common/SuggestionLabelStyle.swift
```diff
@@ -14,7 +14,7 @@ struct SuggestionLabelStyle: LabelStyle {
     var iconColor = Color.icon
 
     func makeBody(configuration: Configuration) -> some View {
-        HStack(alignment: .top, spacing: spacing) {
+        HStack(alignment: .center, spacing: spacing) {
```
gordonbrander, line 17: Fix vertical alignment of suggestion rows with a single line of text.

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
```diff
@@ -0,0 +1,307 @@
+//
```
gordonbrander, line 1: This new subcomponent exposes "meta" controls for notes, such as:

- Change audience
- Change link
- Delete

And in future will be a good place to put note history, IPFS hash, sphere ID, etc.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -821,7 +772,10 @@ struct NotebookModel: ModelProtocol {
 
         /// Find all instances of this model in the stack and update them
         model.details = state.details.map({ (detail: DetailOuterModel) in
-            guard detail.address.slug == receipt.to.slug else {
+            guard let address = detail.address else {
```
gordonbrander, line 775: Fixed logic that was causing bug with renaming. We should be comparing addresses, not slugs, for equality.

xcode/Subconscious/Shared/Components/Notebook/NotebookNavigationView.swift
```diff
@@ -55,7 +56,7 @@ struct NotebookNavigationView: View {
                             store.send(.stageDeleteEntry(slug))
                         }
                     ) {
-                        Text("Delete Immediately")
+                        Text("Delete")
```
gordonbrander, line 59: Changed verbiage. "Delete" makes sense for both local and sphere content.

xcode/Subconscious/Shared/Components/Common/Audience/AudienceMenuButtonView.swift
```diff
@@ -13,7 +13,7 @@ struct AudienceMenuButtonView: View {
     var body: some View {
         Menu(
             content: {
-                Section(header: Text("Post Visibility")) {
+                Section(header: Text("Note Audience")) {
```
gordonbrander, line 16: Consistency of terms

xcode/Subconscious/Shared/Components/Common/Audience/MenuButtonView.swift
```diff
@@ -40,7 +42,7 @@ struct MenuButtonView_Previews: PreviewProvider {
     static var previews: some View {
         MenuButtonView(
             icon: Image(systemName: "network"),
-            label: "Public"
+            label: "Everyone"
```
gordonbrander, line 45: Same language Twitter uses. Will be more familiar.

xcode/Subconscious/Shared/Components/Common/EntryRow.swift
```diff
@@ -11,42 +11,29 @@ import SwiftUI
 /// Provides a preview/excerpt of the entry.
 struct EntryRow: View, Equatable {
```
gordonbrander, line 12: What changed: removed title, refactored list style to work best with

- Excerpt
- Slug
- Date

xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
```diff
@@ -0,0 +1,38 @@
...
+            content
+            Spacer()
+        }
+    }
+}
+
+extension View {
+    /// Modify a view so that it expands to fill the whole available
+    /// horizontal space, with content aligned leading.
+    func expandAlignedLeading() -> some View {
```
gordonbrander, line 24: Causes view to fill the whole row, with content aligned leading. I find myself doing this a lot and its tedious. Add a view modifier extension to accomplish.

xcode/Subconscious/Shared/Models/Identified.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Subconscious (iOS)
+//
+//  Created by Gordon Brander on 3/2/23.
+//
+
+import Foundation
+
+/// A wrapping type that is identifiable.
+/// Use to wrap types that can't conform to identifiable.
+struct Identified<Value>: Identifiable, Equatable
```
gordonbrander, line 12: Allows us to add a persistent identity to records that might otherwise not have one.

xcode/Subconscious/Shared/Components/OmniboxView.swift
```diff
@@ -20,23 +20,29 @@ struct OmniboxView: View {
...
                     .foregroundColor(.accentColor)
                 Spacer()
             }
         }
         .padding(.leading, 8)
         .padding(.trailing, 12)
-        .frame(height: 36)
-        .background(Color.secondaryBackground)
-        .cornerRadius(12)
+        .frame(height: 34)
```
gordonbrander, line 30: Visual tweak for better look-and-feel in navigation view title bar.

xcode/Subconscious/Shared/Config.swift
```diff
@@ -24,10 +24,6 @@ struct Config: Equatable, Codable {
 
     var untitled = "Untitled"
 
-    /// Toggle scratch note suggestion feature
-    var scratchSuggestionEnabled = true
```
gordonbrander, line 28: No longer need scratch feature, since you can create notes with no title/slug now.

xcode/Subconscious/Shared/Library/Prose.swift
```diff
@@ -0,0 +1,66 @@
+//
```
gordonbrander, line 1: General utilities for working with prose. Replaces `truncate.swift`, and adds more utilities.

xcode/Subconscious/Shared/Components/Common/RowButtonStyle.swift
```diff
@@ -7,20 +7,105 @@
 
 import SwiftUI
 
+struct RowLabelStyle: LabelStyle {
```
gordonbrander, line 10: RowButtonStyle now also applies a label style, so that icons are displayed to the right of the row by default.

xcode/Subconscious/Shared/Components/Detail/Detail.swift
```diff
@@ -583,19 +505,50 @@ struct DetailEditorCursor: CursorProtocol {
...
+        default:
+            return .metaSheet(action)
+        }
+    }
+}
+
 //  MARK: Model
 struct DetailModel: ModelProtocol {
     var address: MemoAddress?
+    var defaultAudience = Audience.local
```
gordonbrander, line 542: Default audience determines what kind of note will be derived (public/private) if generating an address from body content.

xcode/Subconscious/Shared/Components/Detail/Detail.swift
```diff
@@ -1463,23 +1389,28 @@ struct DetailModel: ModelProtocol {
...
         let to = from.withAudience(audience)
+
         let fx: Fx<DetailAction> = environment.data.moveEntryAsync(
-            from: from.toEntryLink(title: state.headers.title),
-            to: to.toEntryLink(title: state.headers.title)
-        )
-        .map({ receipt in
+            from: from,
+            to: to
+        ).map({ receipt in
```
gordonbrander, line 1398: I'm transitioning to this chaining style piecemeal, because unfortunately Xcode likes to bork the whitespace of the chaining style we used to use.

xcode/Subconscious/Shared/Components/Detail/Detail.swift
```diff
@@ -1696,168 +1697,87 @@ struct DetailModel: ModelProtocol {
...
-
-        return DetailModel.update(
-            state: model,
-            action: .setRenameField(""),
-            environment: environment
-        )
-    }
 
-    /// Set text of slug field
-    static func setRenameField(
```
gordonbrander, line 1751: This all moved into DetailMetaSheetModel.

xcode/Subconscious/Shared/Components/Detail/Detail.swift
```diff
@@ -1922,89 +1844,7 @@ struct DetailModel: ModelProtocol {
         )
         return Update(state: state)
     }
-
-    /// Retitle entry
-    static func retitleEntry(
```
gordonbrander, line 1927: Retitling is no longer a concept, since we derive title from body content.

xcode/Subconscious/Shared/Models/SaveState.swift
```diff
@@ -9,8 +9,7 @@ import Foundation
 
 /// A simple enum representing saved state, including in-progress write.
 enum SaveState: Hashable, Equatable {
-    case draft
-    case modified
+    case unsaved
```
gordonbrander, line 12: Clarifying save state. Draft vs modified was a distinction without a difference in the logic. What we need to know is whether it is unsaved.

xcode/Subconscious/Shared/Models/Suggestion.swift
```diff
@@ -7,22 +7,35 @@
...
 import Foundation
 
-enum Suggestion: Hashable, Equatable, Identifiable {
-    case entry(EntryLink)
-    case search(EntryLink)
-    case scratch(EntryLink)
+enum Suggestion: Hashable {
+    case memo(address: MemoAddress, fallback: String = "")
+    case createLocalMemo(slug: Slug? = nil, fallback: String = "")
+    case createPublicMemo(slug: Slug? = nil, fallback: String = "")
```
gordonbrander, line 13: New suggestion types for local vs public memos.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -550,16 +550,22 @@ extension Subtext {
...
-                return String(block.body())
+                let block = String(block.body())
+                return block.isEmpty ? fallback : block
             }
         }
-        return ""
+        return fallback
+    }
+    
+    static func excerpt(markup: String, fallback: String = "") -> String {
```
gordonbrander, line 566: Cheaper function that derives excerpt from 512 char substring, so we don't have to parse a large subtext document.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -213,7 +213,7 @@ final class DatabaseService {
                 .text(memo.contentType),
                 .date(memo.created),
                 .date(memo.modified),
-                .text(memo.title),
+                .text(memo.title()),
```
gordonbrander, line 216: Derive title from memo body content.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -320,12 +318,12 @@ final class DatabaseService {
...
             throw DatabaseServiceError.notReady
         }
-        let suggestions = try database.execute(
+        let suggestions: [Suggestion] = try database.execute(
             sql: """
             SELECT id, title
             FROM memo
             ORDER BY modified DESC
-            LIMIT 25
+            LIMIT 5
```
gordonbrander, line 326: Intentional. It looked too visually busy with the new note suggestions to have so many recents.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -382,24 +369,16 @@ final class DatabaseService {
         guard self.state == .ready else {
             throw DatabaseServiceError.notReady
         }
-        // If slug is invalid, return empty suggestions
-        guard let queryEntrySlug = Slug(formatting: query) else {
-            return []
-        }
         
-        var suggestions: OrderedDictionary<Slug, Suggestion> = [:]
+        var suggestions: [Suggestion] = []
```
gordonbrander, line 373: We always include "create note" suggestions now.

xcode/Subconscious/SubconsciousTests/Tests_DatabaseService.swift
```diff
@@ -9,167 +9,31 @@ import XCTest
 @testable import Subconscious
 
 class Tests_DatabaseService: XCTestCase {
-    func testCollateRenameSuggestionsRetitle() throws {
```
gordonbrander, line 12: Retitle no longer exists as a concept in the codebase.

xcode/Subconscious/SubconsciousTests/Tests_Detail.swift
```diff
@@ -197,88 +194,16 @@ class Tests_Detail: XCTestCase {
         )
     }
     
-    func testShowRenameSheet() throws {
```
gordonbrander, line 200: Rename sheet no longer part of detail.

xcode/Subconscious/SubconsciousTests/Tests_Detail.swift
```diff
@@ -332,82 +245,6 @@ class Tests_Detail: XCTestCase {
...
             "Does not change address, because addresses don't match"
         )
-        XCTAssertEqual(
-            update.state.headers.title,
-            "Loomings",
-            "Does not change title, because addresses don't match"
-        )
-    }
-    
-    func testSucceedRetitleEntry() throws {
```
gordonbrander, line 342: Retitle no longer a concept

xcode/Subconscious/Shared/Components/Common/RowButtonStyle.swift
```diff
@@ -7,20 +7,105 @@
...
+            configuration.label
+                .labelStyle(RowLabelStyle())
+                .foregroundColor(
+                    configuration.role == .destructive ?
+                    Color.red :
+                    Color.primaryButtonText
+                )
+        }
+        .padding(insets)
+        .frame(minHeight: 44)
```
bfollington, line 43: Another `44`

xcode/Subconscious/Shared/Components/Common/Suggestion/SuggestionLabelView.swift
```diff
@@ -0,0 +1,133 @@
...
+                Text("New note")
+                Text(verbatim: #""\#(preview)""#)
+                    .foregroundColor(.secondary)
+            }
+        } else {
+            HStack {
+                Text("New note")
+            }
+        }
+    }
```
bfollington, line 29: nitpick: I would swap this to prevent the boolean negation

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
```diff
@@ -0,0 +1,328 @@
...
+    @Environment(\.dismiss) private var dismiss
+    var state: DetailMetaSheetModel
+    var send: (DetailMetaSheetAction) -> Void
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            HStack {
+                VStack(alignment: .leading, spacing: AppTheme.unit2) {
+                    HStack {
+                        if let slaslink = state.address?.toSlashlink() {
```
bfollington, line 21: nitpick: typo in "slaslink" should be "slashlink"

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
```diff
@@ -0,0 +1,328 @@
...
+    
+    var address: MemoAddress?
+    var defaultAudience = Audience.local
+    var audience: Audience {
+        address?.toAudience() ?? defaultAudience
+    }
+    var isRenameSheetPresented = false
+    var renameSearch = RenameSearchModel()
+    
+    /// Is delete confirmation dialog presented?
```
bfollington, line 156: nitpick: probably superfluous given the variable name

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
```diff
@@ -0,0 +1,328 @@
...
+                environment: environment,
+                address: address
+            )
+        case .setDefaultAudience(let audience):
+            return setDefaultAudience(
+                state: state,
+                environment: environment,
+                audience: audience
+            )
+        case .requestUpdateAudience:
```
bfollington, line 210: Might be worth a comment explaining where this is actually handled?

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
bfollington: Maybe I'm missing something, but it looks like we're still checking slugs on the line below?

xcode/Subconscious/Shared/Components/Notebook/NotebookNavigationView.swift
```diff
@@ -41,12 +41,13 @@ struct NotebookNavigationView: View {
                 )
                 .ignoresSafeArea(.keyboard, edges: .bottom)
                 .confirmationDialog(
-                    "Are you sure?",
+                    "Are you sure you want to delete?",
```
bfollington, line 44: Can we fit "Are you sure you want to delete _this note_?" I tend to prefer confirmations to be specific about the result of the action.

xcode/Subconscious/Shared/Components/OmniboxView.swift
```diff
@@ -20,23 +20,29 @@ struct OmniboxView: View {
             HStack(spacing: 0) {
                 Spacer()
                 Text(verbatim: title)
+                    .font(.callout)
                     .foregroundColor(.accentColor)
```
bfollington, line 24: It might be nice to come up with an empty state here. imo it looks like the app is in a broken state when opening a new note right now - like it failed to load the UI until it auto-titles after 15 seconds:

image

Could do something like this:
```swift
 if title.count == 0 {
    Text(verbatim: "Untitled")
        .font(.callout)
        .foregroundColor(.secondary)
} else {
    Text(verbatim: title)
        .font(.callout)
        .foregroundColor(.accentColor)
}
```

Resulting in: 
image

xcode/Subconscious/Shared/Components/OmniboxView.swift
bfollington: More magic numbers, but these don't seem worth adding to `AppTheme` to me.

xcode/Subconscious/Shared/Library/Prose.swift
```diff
@@ -0,0 +1,66 @@
...
+    ) -> String {
+        guard self.count > maxLength else {
+            return self
+        }
+        let truncated = self.prefix(
+            maxLength
+        ).trimmingCharacters(
+            in: .whitespacesAndNewlines
+        )
+        return !truncated.isEmpty ? "\(truncated)\(ellipsis)" : ""
```
bfollington, line 40: If I pass `maxLength` = `256` does this result in a `257` character string due to the ellipsis?

xcode/Subconscious/Shared/Library/Prose.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /// Get first sentence of substring
+    var firstSentence: Substring {
+        self.prefix(while: { character in
+            !character.isSentenceEndingPunctuation
+        })
+    }
```
bfollington, line 48: Elegant ❤️

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -469,22 +446,43 @@ struct DataService {
...
+            // If we can't derive slug from text, give up.
+            guard let slugN = Slug("\(slug.description)-\(n)") else {
+                return nil
+            }
+            if findAddress(slug: slugN) == nil {
+                return slugN.toMemoAddress(audience)
+            }
+        }
+        return nil
+    }
```
bfollington, line 470: Nice work, I can't find any way to break this so far!

xcode/Subconscious/SubconsciousTests/Tests_DataService.swift
```diff
@@ -163,13 +158,11 @@ final class Tests_DataService: XCTestCase {
...
-            title: "Test fallback",
             fallback: "Fallback content"
         )
         
         XCTAssertEqual(detail.entry.address, address)
         XCTAssertEqual(detail.entry.contents.body, memo.body)
-        XCTAssertEqual(detail.entry.contents.title, memo.title)
     }
     
     func testManyWritesThenCloseThenReopen() throws {
```
bfollington, line 168: This test did not change in this PR, but it fails for me locally... Can you confirm if it passes for you?

`testManyWritesThenCloseThenReopen(): XCTAssertEqual failed: ("bafy2bzacecrxijbvgzo3shfqmzfk6wazsyovevdsuur6zhvbtzfz752m2hqqi") is not equal to ("bafy2bzacebu4uahwwzbz3n3jzz2wvos2oypodx2msxasgqpowkzu332uv5qe6")`

`testManyWritesThenCloseThenReopen(): failed: caught error: "fileDoesNotExist("/b")"`

bfollington (changes requested):
Mostly questions and clarifications, a few typos and suggested improvements too.

xcode/Subconscious/Shared/Components/Common/Suggestion/SuggestionLabelView.swift
```diff
@@ -0,0 +1,133 @@
...
+    var empty = String(localized: "Empty note")
+
+    var body: some View {
+        switch suggestion {
+        case let .memo(address, title):
+            Label(
+                title: {
+                    TitleGroupView(
+                        title: Text(
+                            verbatim: !title.isEmpty ? title : empty
```
bfollington, line 43: nitpick: another double negative

xcode/Subconscious/Shared/Components/OmniboxView.swift
gordonbrander: @bfollington Yup. This is a unique case because Omnibox lives in a Navigation toolbar, which is fixed height. The magic number here is what looks visually right for this particular component in this particular context. I don't think its useful elsewhere.

xcode/Subconscious/Shared/Components/Common/Audience/MenuButtonView.swift
```diff
@@ -19,19 +19,21 @@ struct MenuButtonView<Icon: View>: View {
...
+            Spacer()
             Image(systemName: "chevron.down")
                 .font(.system(size: 12))
+                .foregroundColor(Color.secondary)
         }
+        .foregroundColor(Color.accentColor)
         .lineLimit(1)
-        .frame(height: AppTheme.unit * 6)
-        .foregroundColor(Color.primaryButtonText)
+        .frame(height: 30)
```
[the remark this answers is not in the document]
gordonbrander: > Also worth mentioning: increasing the system font size will make the font bigger here but will not increase the size of the frame, leading to a cramped appearance. Non-blocking but just an FYI

@bfollington interesting catch. Unfortunately this item must have a hard-coded width for the transition between Everyone/Local to work right. Otherwise the text reflows and we get a janky animation. Doing better than this will probably require defining some dynamic unit that is responsive to system font size.

Filed an issue to pursue that in a followup PR https://github.com/subconsciousnetwork/subconscious/issues/436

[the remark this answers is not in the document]
gordonbrander: > Does this correspond to a standard size? If not, should we add it to AppTheme as a constant? Both 44 (because Apple HIG), 32 and 30 all seem like good candidates for constants.

I did not specify this as a constant, because I want to discourage casual use of anything smaller than 44 (minimum touch target). This control is a bespoke exception, because it is wider than 44, making it easy to hit. It's based roughly off of similar selector controls seen in Notes app and elsewhere.

xcode/Subconscious/Shared/Components/Common/Buttons/CloseButtonView.swift
```diff
@@ -0,0 +1,37 @@
...
+        Button(
+            action: action,
+            label: {
+                Image(systemName: "xmark")
+                    .font(.system(size: 16, weight: .semibold))
+                    .frame(width: 30, height: 30)
+                    .foregroundColor(.secondary)
+                    .background(Color.secondaryBackground)
+                    .clipShape(Circle())
+                    .frame(width: 44, height: 44)
```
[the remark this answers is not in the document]
gordonbrander: Added `AppTheme.minTouchSize` for 44.

I'm wary of adding constants for anything below 44, since these are only used in bespoke cases, and should only be used carefully in bespoke places.

xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
```diff
@@ -0,0 +1,38 @@
...
+import SwiftUI
+
+/// Modify a view so that it expands to fill the whole available
+/// horizontal space, with content aligned leading.
+struct ExpandAlignLeadingViewModifier: ViewModifier {
+    func body(content: Content) -> some View {
+        HStack(spacing: 0) {
+            content
+            Spacer()
+        }
```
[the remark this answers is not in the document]
gordonbrander: I find that spacer is generally more reliable for filling space and setting alignment. It's also the approach Apple tends to push in their tutorials.

Setting the frame can work in some situations, but it literally just sets the outer frame, and does not necessarily set the alignment of content. As a result there are circumstances where setting frame works, and other circumstances (VStack vs HStack vs other contexts) where you end up with a full-width frame but centered content. Spacer does not have these problems.

xcode/Subconscious/Shared/Components/Common/MetaTableView.swift
```diff
@@ -24,6 +24,25 @@ struct MetaTableView<Rows: View>: View {
     }
 }
 
+struct MetaTableRowButtonStyle: ButtonStyle {
+    func makeBody(configuration: Configuration) -> some View {
+        configuration.label
+            .foregroundColor(.accentColor)
+            .frame(minHeight: 44)
```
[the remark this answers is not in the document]
gordonbrander: Fixed

xcode/Subconscious/Shared/Components/Common/RowButtonStyle.swift
gordonbrander: Fixed

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
gordonbrander: Fixed

xcode/Subconscious/SubconsciousTests/Tests_DataService.swift
gordonbrander: @bfollington it's from a previous PR. Was originally added to troubleshoot the persistence issue we had with Noosphere earlier. Fails because Swift destructors are not guaranteed to run until end of scope. Was testing to see if this was causing the issue. It does cause _a_ issue (hence failing test). However, it was not _the_ issue.

I think we can remove it now that Noosphere flushes on save, or rewrite it so that it passes in this state.

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
gordonbrander: Noted in action definition. https://github.com/subconsciousnetwork/subconscious/pull/427/files/55bcffaa2f77af4ec15e64cdd1ec5989c006d163#diff-cc3a094b36c3ad6fcf4fd7f720a74198f020e94eb390d0cd45d19b14456a64afR125

I prefer to have these comments at the point of action definition, because they often bubble up in Xcode's code helpers when written there.

xcode/Subconscious/Shared/Components/Notebook/NotebookNavigationView.swift
gordonbrander: Good call.

xcode/Subconscious/Shared/Components/OmniboxView.swift
gordonbrander: I suggest following up in a subsequent PR https://github.com/subconsciousnetwork/subconscious/issues/437

xcode/Subconscious/Shared/Library/Prose.swift
gordonbrander: Correct. Since ellipsis can be any string, truncating to x chars INCLUDING ellipsis requires a lot of additional logic to make sure you stay under budget.

I could change the label to `maxLengthNotIncludingEllipsis` or `maxTextLength`?

xcode/Subconscious/Shared/Components/Common/Suggestion/SuggestionLabelView.swift
gordonbrander: These come down to me liking to read the "default" case first and the corner case second. I can see your POV on double negative though.

xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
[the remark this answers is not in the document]
bfollington: I would've thought the `alignment:` parameter on `frame` addresses this concern directly? Either way I don't think this matters for our changes here so happy to leave it.

xcode/Subconscious/Shared/Library/Prose.swift
bfollington: Renaming to `maxLengthNotIncludingEllipsis` works for me me but I would've though that `self.prefix(maxLength - ellipsis.count)` would do the job?

bfollington: ... which raises a question for me of what this function does when passed `maxLength` of `0` or `-1` etc.

xcode/Subconscious/Shared/Library/Prose.swift
```diff
@@ -0,0 +1,66 @@
...
+        Self.sentenceEndingPunctuation.contains(self)
+    }
+}
+
+extension String {
+    /// Truncate string to max length, appending ellipsis if truncated.
+    func truncate(
+        maxLength: Int = 256,
+        ellipsis: String = "…",
+        fallback: String = ""
```
bfollington, line 30: Missed this before, this parameter is unused in the body

xcode/Subconscious/Shared/Components/Common/Audience/MenuButtonView.swift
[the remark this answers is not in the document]
bfollington: All good 👍

bfollington (approved):
Replied to a couple of discussion threads and flagged a potential issue, but LGTM

xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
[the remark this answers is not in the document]
gordonbrander: Surprisingly it doesn't always!

xcode/Subconscious/Shared/Library/Prose.swift
gordonbrander: Actually you're right and its just late for me. I'm too fried for this rn. Fixed and updated tests.

xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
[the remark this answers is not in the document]
bfollington: TIL!

bfollington (approved):
Let's go!

xcode/Subconscious/SubconsciousTests/Tests_DataService.swift
gordonbrander: Follow-up here https://github.com/subconsciousnetwork/subconscious/pull/441

PR #428, opened 2023-03-03
Embedded SwiftUI Transclude Blocks

### LINKED ISSUE #328: As a user, I want links to transclude the linked content
As a user, I want links to transclude the linked content, so I can treat links as embeds for the content.

Related:

- #214
- #324

bfollington:
Fixes #328

- [x] Fix half-pixel trimming on X axis 
     - try wrapping swift view with padding 
- [x] Try UIKit contraint layout instead of naive for loop (https://developer.apple.com/documentation/uikit/uiview/1622623-systemlayoutsizefitting)
- [x] Fix height calculation of textarea (manually count layout fragment heights?)
- [x] Improve dataflow architecture for slashlink previews
- [x] Do not render transcludes unless lookup succeeds
- [ ] Newline handling at end of doc can introduce space between slashlink and transclude

# Current State

https://user-images.githubusercontent.com/5009316/227083975-4abdbc46-6e08-4db5-9d67-be185391a8fd.mov

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -904,6 +904,7 @@ struct AppEnvironment {
 
     var data: DataService
     var feed: FeedService
+    var slashlinks: SlashlinkLookupService
     
     var recoveryPhrase = RecoveryPhraseEnvironment()
     var addressBook: AddressBookEnvironment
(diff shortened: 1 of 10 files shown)


bfollington:
I've had a minor breakthrough on a different approach here.

Instead of rendering the transcludes using TK2 and `TextLayoutFragment` hacks, we can simply _create space_ for the transcludes and render them in a second pass.

I've added code to demonstrate how to layer transcludes on top of our textview but there is work remaining to actually calculate all the relevant rectangles during layout, send them to the store and render them in the overlay.

There is also a double measurement inefficiency here. To lay out the text we need the size of the view, but we need the size of the view to render the view, so we have to calculate layout twice in practice (since SwiftUI will be forced to do the work when we actually mount the component). This might not be avoidable but if the rendering is suitably efficient then it won't matter too much.

PR #420, opened 2023-02-27
Basic AddressBookView - CRUD

bfollington:
This PR introduces a simple Address Book UI + backing model. 
This is the first step towards https://github.com/subconsciousnetwork/subconscious/issues/418, with PRs to follow to read and write friends via Noosphere FFI calls.

- [x] Validated `Did` wrapper type
- [x] Address Book UI
- [x] Add Friend UI
- [x] Remove Friend UI
- [x] Backing Model
- [x] Actions
    - add friend
    - remove friend
    - edit friend(?) // not strictly needed, potentially not supported by FFI
- [x] ~Search Friends~
- [x] Tap to share
- [x] Actually show the real DID of this users's sphere

Adding via QR code has moved to https://github.com/subconsciousnetwork/subconscious/pull/429 as it is non-essential.

PR #429, opened 2023-03-06
Add Friend via QR Code

bfollington:
Builds on #420

This PR introduces support for generating a QR Code encoding the user's DID as well as the ability to scan another user's QR Code during the Add Friend flow to populate their DID. I've tested this on device and it seems well behaved.

# Added
- QR for the user's DID is displayed in the "My DID" section of the address book
- New "Scan QR Code" option added to "Follow User" form

# Tasks
- [x] Add by scanning QR Code
    - looks to be a little involved, reference implementation here https://blog.devgenius.io/camera-preview-and-a-qr-code-scanner-in-swiftui-48b111155c66
    - there's also this writeup https://www.hackingwithswift.com/books/ios-swiftui/scanning-qr-codes-with-swiftui an a library from the author https://github.com/twostraws/CodeScanner
- [x] Generate personal QR code
    - can be deeply further customised, not sure how hard that is to do without using a library like https://github.com/dagronf/QRCode
- [x] Customise QR code for light vs dark theme (can we use the accent colour?)
- [x] Test on-device

# On-Device Demo
https://user-images.githubusercontent.com/5009316/226531915-084a8a60-9d21-421a-b186-f703a7d9b76b.mov

### DIFF
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
@@ -10,23 +10,41 @@ import SwiftUI
 import CodeScanner
 
 struct AddFriendViaQRCodeView: View {
+    @Environment(\.dismiss) var dismiss
+    
     var onScannedDid: (String) -> Void
-    @Environment(\.presentationMode) var presentationMode: Binding<PresentationMode>
+    var onCancel: () -> Void
     
     var body: some View {
-        CodeScannerView(
-            codeTypes: [.qr],
-            simulatedData: "did:key:z6MkmCJAZansQ3p1Qwx6wrF4c64yt2rcM8wMrH5Rh7DGb2K7"
-        ) { res in
-            switch res {
-            case .success(let result):
-                print("Found code: \(result.string)")
-                onScannedDid(result.string)
-            case .failure(let error):
-                print(error.localizedDescription)
+        VStack {
+            // Toolbar is unavailable here, create a minimal one that blends with notch/bezel.
+            HStack {
+                Button("Cancel", role: .cancel) {
+                    dismiss()
+                }
+                .buttonStyle(.borderless)
+
+                Spacer()
             }
+            .padding(AppTheme.unit2)
+            .background(Color.black)
             
-            presentationMode.wrappedValue.dismiss()
+            CodeScannerView(
+                codeTypes: [.qr],
+                showViewfinder: true,
+                simulatedData: "did:key:z6MkmCJAZansQ3p1Qwx6wrF4c64yt2rcM8wMrH5Rh7DGb2K7",
+                shouldVibrateOnSuccess: true
+            ) { res in
+                switch res {
+                case .success(let result):
+                    print("Found code: \(result.string)")
+                    onScannedDid(result.string)
+                case .failure(let error):
+                    print(error.localizedDescription)
+                }
+                
+                dismiss()
+            }
         }
     }
 }
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
@@ -13,8 +13,9 @@ import os
 struct AddFriendViaQRCodeView: View {
     @Environment(\.dismiss) var dismiss
     
-    var onScannedDid: (String) -> Void
+    var onScanResult: (Result<ScanResult, ScanError>) -> Void
     var onCancel: () -> Void
+    var errorMessage: String?
     
     static let logger = Logger(
         subsystem: Config.default.rdns,
@@ -41,16 +42,22 @@ struct AddFriendViaQRCodeView: View {
                 simulatedData: Config.default.fallbackSimulatorQrCodeScanResult,
                 shouldVibrateOnSuccess: true
             ) { res in
-                switch res {
-                case .success(let result):
-                    AddFriendViaQRCodeView.logger.debug("Found code: \(result.string)")
-                    onScannedDid(result.string)
-                case .failure(let error):
-                    AddFriendViaQRCodeView.logger.debug("\(error.localizedDescription)")
-                }
-                
+                onScanResult(res)
                 dismiss()
             }
+        }
+        .alert(
+            isPresented: Binding(
+                get: { errorMessage != nil },
+                set: { _ in }
+            )
+        ) {
+            let errorMessage = errorMessage ?? "Failed to scan"
+            
+            return Alert(
+                title: Text("QR Code Error"),
+                message: Text(errorMessage)
+            )
         }
     }
 }
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
@@ -38,7 +38,7 @@ struct AddFriendViaQRCodeView: View {
             CodeScannerView(
                 codeTypes: [.qr],
                 showViewfinder: true,
-                simulatedData: "did:key:z6MkmCJAZansQ3p1Qwx6wrF4c64yt2rcM8wMrH5Rh7DGb2K7",
+                simulatedData: Config.default.fallbackSimulatorQrCodeScanResult,
                 shouldVibrateOnSuccess: true
             ) { res in
                 switch res {
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
@@ -8,12 +8,18 @@
 import Foundation
 import SwiftUI
 import CodeScanner
+import os
 
 struct AddFriendViaQRCodeView: View {
     @Environment(\.dismiss) var dismiss
     
     var onScannedDid: (String) -> Void
     var onCancel: () -> Void
+    
+    static let logger = Logger(
+        subsystem: Config.default.rdns,
+        category: "AddFriendViaQRCodeView"
+    )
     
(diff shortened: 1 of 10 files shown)


xcode/Subconscious/Shared/Components/AddressBook/DidQrCodeView.swift
```diff
@@ -0,0 +1,66 @@
...
+    let colorInvertFilter = CIFilter.colorInvert()
+    let maskToAlphaFilter = CIFilter.maskToAlpha()
+    let colorReplaceFilter = CIFilter.spotColor()
+    
+    let maskFilter = CIFilter.blendWithMask()
+    
+    func generateQRCode(from string: String) -> UIImage? {
+        filter.message = Data(string.utf8)
+        
+        colorInvertFilter.inputImage = filter.outputImage
```
bfollington, line 26: Base image from the QR Code generator is black-on-white

xcode/Subconscious/Shared/Components/AddressBook/DidQrCodeView.swift
```diff
@@ -0,0 +1,66 @@
...
+        guard let inverted = colorInvertFilter.outputImage else { return nil }
```
bfollington, line 27: Invert to white-on-black to use it as a blending mask (where black is full transparency and white full opacity)

xcode/Subconscious/Shared/Components/AddressBook/DidQrCodeView.swift
```diff
@@ -0,0 +1,66 @@
...
+        // This is the only way I managed to produce a CIColor from .accentColor
+        // https://developer.apple.com/forums/thread/687764
+        // https://stackoverflow.com/questions/56586055/how-to-get-rgb-components-from-color-in-swiftui
+        var r: CGFloat = 0.0
+        var g: CGFloat = 0.0
+        var b: CGFloat = 0.0
+        var a: CGFloat = 0.0
+        UIColor(Color.accentColor).getRed(&r, green: &g, blue: &b, alpha: &a)
+        let col = CIColor(red: r, green: g, blue: b, alpha: a)
+        maskFilter.inputImage = CIImage(color: col)
```
bfollington, line 39: Convert `accentColor` to a `CIColor` and then construct solid-color image to blend with the mask we created above.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -12,14 +12,20 @@ struct ValidatedTextField: View {
     var placeholder: String
     @Binding var text: String
     var caption: String
-    var isValid: Bool = true
+    var hasError: Bool = false
```
bfollington, line 15: This allows for clearer of expression between the different layers of concern. imo a text field should not show an error if it has not been touched by the user yet, even if the underlying value is invalid. It's incorrect to say that the field is valid before it's been touched, technically it's invalid if we tried to submit the form, but it also should not have a visible error.

Flipping the logic keeps the ValidatedTextField dumb and lets FormField expose both isValid and hasError properties and drive the specific expected behaviour.

xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
```diff
@@ -0,0 +1,32 @@
...
+import CodeScanner
+
+struct AddFriendViaQRCodeView: View {
+    var onScannedDid: (String) -> Void
+    @Environment(\.presentationMode) var presentationMode: Binding<PresentationMode>
+    
+    var body: some View {
+        CodeScannerView(
+            codeTypes: [.qr],
+            simulatedData: "did:key:z6MkmCJAZansQ3p1Qwx6wrF4c64yt2rcM8wMrH5Rh7DGb2K7"
```
bfollington, line 19: This value is returned in the simulator, since you can't access a real camera context.

xcode/Subconscious/Shared/Components/AddressBook/AddFriendView.swift
```diff
@@ -7,24 +7,53 @@
...
+        get {
+            validated != nil
+        }
+    }
+    var hasError: Bool {
+        get {
+            validated == nil && touched
+        }
+    }
+}
```
bfollington, line 34: Not trying to venture too far into an form framework here, this seems like enough for ephemeral forms with a few fields. There's much more to explore RE: validation and error messages later.

xcode/Subconscious/Subconscious.xcodeproj/project.pbxproj
```diff
@@ -1783,6 +1795,7 @@
 				INFOPLIST_FILE = iOS/Info.plist;
 				INFOPLIST_KEY_CFBundleDisplayName = Subconscious;
 				INFOPLIST_KEY_LSSupportsOpeningDocumentsInPlace = YES;
+				INFOPLIST_KEY_NSCameraUsageDescription = "Optionally adding friends via scanning a QR Code";
```
bfollington, line 1798: This is required for requesting camera permissions on-device.

xcode/Subconscious/Shared/Config.swift
```diff
@@ -16,6 +16,7 @@ struct Config: Equatable, Codable {
 
     var appTabs = false
     var addressBook = false
+    var addByQRCode = false
```
bfollington, line 19: Flagged off for now

xcode/Subconscious/Shared/Components/AddressBook/DidQrCodeView.swift
```diff
@@ -0,0 +1,65 @@
...
+        // This is the only way I managed to produce a CIColor from .accentColor
+        // https://developer.apple.com/forums/thread/687764
+        // https://stackoverflow.com/questions/56586055/how-to-get-rgb-components-from-color-in-swiftui
+        var r: CGFloat = 0.0
+        var g: CGFloat = 0.0
+        var b: CGFloat = 0.0
+        var a: CGFloat = 0.0
+        UIColor(color).getRed(&r, green: &g, blue: &b, alpha: &a)
+        let col = CIColor(red: r, green: g, blue: b, alpha: a)
+        maskFilter.inputImage = CIImage(color: col)
```
bfollington, line 37: Convert `color` to a `CIColor` and then construct solid-color image to blend with the mask we created above.

xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
```diff
@@ -0,0 +1,56 @@
...
+
+                Spacer()
+            }
+            .padding(AppTheme.unit2)
+            .background(Color.black)
+            
+            CodeScannerView(
+                codeTypes: [.qr],
+                showViewfinder: true,
+                simulatedData: "did:key:z6MkmCJAZansQ3p1Qwx6wrF4c64yt2rcM8wMrH5Rh7DGb2K7",
```
bfollington, line 41: Returned in the simulator when testing, since we can't access the camera in that context.

xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
```diff
@@ -0,0 +1,56 @@
...
+                shouldVibrateOnSuccess: true
```
gordonbrander, line 42: Is vibrate on success a typical pattern for this kind of interaction?

xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
```diff
@@ -0,0 +1,56 @@
...
+            ) { res in
+                switch res {
```
gordonbrander, line 44: I might suggest moving this business logic into `update` by sending this result up via the callback (maybe changing name to `onScanResult`) and implementing the `switch` and logging in `update`.

gordonbrander (changes requested):
Overall looks great. A few minor requests.

xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
bfollington: I can't find any mention in the Apple HIG about this, I was going off of my own personal preference while using it. The view dismisses itself _as soon as_ a code is enters the view which can be a little shocking for the user.

Other option is a little toast notification of some kind saying the scan succeeded?

gordonbrander: SGTM. Let's move forward with what you have here. We can always refine later.

xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
bfollington: I extracted this to a property in `Config` that only returns data when `debug` is true, feels like this view shouldn't care about the value itself and should remain dumb.

xcode/Subconscious/Shared/Components/AddressBook/AddFriendViaQRCodeView.swift
bfollington: I lifted this up to the parent component + added error handling via the store and an `.alert`. This has never come up for me, but seems wise to have an error handling path.

xcode/Subconscious/Shared/Components/AddressBook/FollowUserView.swift
```diff
@@ -15,6 +16,15 @@ struct FollowUserView: View {
     }
     var send: (AddressBookAction) -> Void
     
+    func onQRCodeScanResult(res: Result<ScanResult, ScanError>) {
```
gordonbrander, line 19: Non-blocking: we could send `res` through a `.qrCodeResult(res)` action and do this in the update function instead of in the view.

gordonbrander (approved):

PR #432, opened 2023-03-07
Restore full-document attribute rendering

### LINKED ISSUE #431: Tapping slashlinks in editor does nothing
This is a recent regression.

# Reproduction

1. Create a note
2. Add a `/slashlink` to the body
3. Try and tap it

# Expected Behaviour

Tapping a link navigates to the note it references.

bfollington:
This renders attributes directly to TextStorage (like we used to but with TK2) to resolve #431

- [x] Restore old functionality
- [x] Remove per-fragment rendering (otherwise we're doing it twice)

I've also put together a proof-of-concept for how we could do per-paragraph rendering in #433

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
+++ b/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
@@ -181,7 +181,7 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
     }
 
     //  MARK: Coordinator
-    class Coordinator: NSObject, UITextViewDelegate, NSTextContentStorageDelegate, NSTextContentManagerDelegate, NSTextLayoutManagerDelegate {
+    class Coordinator: NSObject, UITextViewDelegate, NSTextContentStorageDelegate, NSTextStorageDelegate, NSTextContentManagerDelegate, NSTextLayoutManagerDelegate {
         /// Is event happening during updateUIView?
         /// Used to avoid setting properties in events during view updates, as
         /// that would cause feedback cycles where an update triggers an event,
@@ -196,6 +196,39 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
             self.isUIViewUpdating = false
             self.representable = representable
         }
+        
+        /// NSTextStorageDelegate method
+        /// Handle markup rendering, just before processEditing is fired.
+        /// It is important that we render markup in `willProcessEditing`
+        /// because it happens BEFORE font substitution. Rendering before font
+        /// substitution gives the OS a chance to replace fonts for things like
+        /// Emoji or Unicode characters when your font does not support them.
+        /// See:
+        /// https://github.com/gordonbrander/subconscious/wiki/nstextstorage-font-substitution-and-missing-text
+        ///
+        /// 2022-03-17 Gordon Brander
+        func textStorage(
+          _ textStorage: NSTextStorage,
+          willProcessEditing: NSTextStorage.EditActions,
+          range: NSRange,
+          changeInLength: Int
+        ) {
+          SubtextTextViewRepresentable.logger.debug(
+              "textStorage: render markup attributes"
+          )
+          textStorage.setAttributes(
+              [:],
+              range: NSRange(
+                  textStorage.string.startIndex...,
+                  in: textStorage.string
+              )
+          )
+            
+          // Render markup on TextStorage (which is an NSMutableString)
+          // using closure set on view (representable)
+          Subtext.renderAttributesOf(textStorage, url: SubtextTextViewRepresentable.toSubURL)
+        }
+
 
         /// Handle link taps
         func textView(
@@ -350,6 +383,7 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
         
         let view = SubtextTextView(frame: self.frame, textContainer: textContainer)
         view.delegate = context.coordinator
+        view.textStorage.delegate = context.coordinator
 
         // Set inner padding
         view.textContainerInset = self.textContainerInset
--- a/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
+++ b/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
@@ -165,7 +165,14 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
     }
 
     //  MARK: Coordinator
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -164,38 +164,55 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
...
-    class Coordinator: NSObject, UITextViewDelegate, NSTextContentStorageDelegate, NSTextContentManagerDelegate, NSTextLayoutManagerDelegate {
+    class Coordinator: NSObject, UITextViewDelegate, NSTextContentStorageDelegate, NSTextStorageDelegate, NSTextContentManagerDelegate, NSTextLayoutManagerDelegate {
         /// Is event happening during updateUIView?
         /// Used to avoid setting properties in events during view updates, as
         /// that would cause feedback cycles where an update triggers an event,
         /// which triggers an update, which triggers an event, etc.
         var isUIViewUpdating: Bool
         var representable: SubtextTextViewRepresentable
         var renderTranscludeBlocks: Bool = false
+        var subtext: Subtext? = nil
```
bfollington, line 176: I'm caching the parse results so we can use them to detect transclude blocks without repeating work.

xcode/Subconscious/Shared/Library/NSRangeUtilities.swift
```diff
@@ -14,3 +15,26 @@ extension NSRange {
         return range != nil
     }
 }
+
+
+// https://gist.github.com/krzyzanowskim/057676af670f06fa6061473ac28a6c58
+extension NSRange {
```
bfollington, line 21: No license on this gist :( but it's incredibly useful

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -164,38 +164,55 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
...
+        func textStorage(
+          _ textStorage: NSTextStorage,
+          willProcessEditing: NSTextStorage.EditActions,
+          range: NSRange,
+          changeInLength: Int
+        ) {
+            SubtextTextViewRepresentable.logger.debug(
+              "textStorage: render markup attributes"
+            )
+            textStorage.setAttributes(
```
gordonbrander, line 204: What is this code doing (`textStorage.setAttributes`)

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -164,38 +164,55 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
...
-                url: SubtextTextViewRepresentable.toSubURL
-            )
-            self.subtext = subtext
-            super.init(attributedString: renderedAttributedString)
-        }
-    }
-
     //  MARK: Coordinator
-    class Coordinator: NSObject, UITextViewDelegate, NSTextContentStorageDelegate, NSTextContentManagerDelegate, NSTextLayoutManagerDelegate {
+    class Coordinator: NSObject, UITextViewDelegate, NSTextContentStorageDelegate, NSTextStorageDelegate, NSTextContentManagerDelegate, NSTextLayoutManagerDelegate {
```
gordonbrander, line 168: Nit: let's wrap this long line. https://google.github.io/swift/#type-and-extension-declarations

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
gordonbrander, line 204: What is this code doing (textStorage.setAttributes)?

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -299,18 +318,6 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
 
             return layoutFragment
         }
-        
-        // MARK: - NSTextContentStorageDelegate
-        
-        func textContentStorage(_ textContentStorage: NSTextContentStorage, textParagraphWith range: NSRange) -> NSTextParagraph? {
-            guard let attributedSubstring = textContentStorage.textStorage?
```
gordonbrander, line 306: Did the API change in TK2? This form would mean we don't need to bring in the NSRange extension.

xcode/Subconscious/Shared/Library/NSRangeUtilities.swift
gordonbrander: If no license, then we must write our own.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -722,3 +724,21 @@ extension Subtext {
         return shortlinkFor(index: range.lowerBound)
     }
 }
+
+extension Subtext {
```
gordonbrander, line 728: Would prefer to add this extension in the SubtextTextEditor file. That way SubtextTextEditor knows about Subtext and NSTextParagraph, but Subtext doesn't have to know about NSTextParagraph.

gordonbrander (changes requested):
A few questions and clarifications on how this is working

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
gordonbrander, line 204: If I understand correctly, this is clearing all attributes from string.

I think we should always clear all attributes from string before rendering new attributes with `renderAttributesOf`.

Let's move this to `Subtext.renderAttributesOf`.

gordonbrander (changes requested):

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -722,3 +724,21 @@ extension Subtext {
...
+                return false
+            }
+            guard let tcm = forParagraph.textContentManager else {
+                return false
+            }
+            
+            guard let range: Range<String.Index> = Range(NSRange(contentRange, in: tcm), in: base) else {
+                return false
+            }
+            return b.body().range.overlaps(range)
```
bfollington, line 741: Let's check that subtext is using the same underlying strings as textContentManager

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
bfollington: This is the way it _used_ to work before we moved to TK2. Regardless, I've moved this into `renderAttributesOf` now.

gordonbrander (approved):

PR #438, opened 2023-03-09
Noosphere Address Book FFI Stub

### LINKED ISSUE #443: Integrate FFI interface with Address Book UI
Actually call the FFI methods to populate the Address Book UI and mutate its contents.

bfollington:
Resolves #443

This PR integrates the Address Book UI from #420 with the underlying persistence layer via Noosphere.

- [x] Implement actions for NS operations
- [x] Stub methods out
- [x] Model lifecycle including failure and success paths
- [x] Implement cache for address book
- [x] Tests for `FormField` model
- [x] Tests for `AddressBookModel` and/or `AddressBookService`
- [x] Switch the `Follow User` navigation to a `.sheet` instead (we can control presentation from the store directly)
    - Right now we dismiss the form on submission but we need to keep it open to show errors
- [x] Display errors in UI on fail to follow and fail to unfollow

**Note**: Swipe to unfollow can remove the entry from the UI without mutating the underlying store

### DIFF
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
@@ -73,6 +73,7 @@ enum AddressBookAction {
     case populate([AddressBookEntry])
     
     case requestFollow
+    case attemptFollow
     case failFollow(error: String)
     case succeedFollow(did: Did, petname: Petname)
     
@@ -108,48 +109,11 @@ struct AddressBookModel: ModelProtocol {
     ) -> Update<AddressBookModel> {
         switch action {
             
-        case .populate(let follows):
-            var model = state
-            model.follows = follows
-            return Update(state: model)
-            
-        case .presentFollowUserForm(let isPresented):
-            var model = state
-            
-            model.isFollowUserFormPresented = isPresented
-            if isPresented {
-                // Reset form when opened
-                model.followUserForm.did = FormField.update(
-                    state: model.followUserForm.did,
-                    action: .reset,
-                    environment: FormFieldEnvironment()
-                ).state
-                model.followUserForm.petname = FormField.update(
-                    state: model.followUserForm.petname,
-                    action: .reset,
-                    environment: FormFieldEnvironment()
-                ).state
-            }
-            
-            return Update(state: model)
-            
-        case .didField(let action):
-            return DidFieldCursor.update(
-                state: state,
-                action: action,
-                environment: FormFieldEnvironment()
-            )
-            
-        case .petnameField(let action):
-            return PetnameFieldCursor.update(
-                state: state,
-                action: action,
-                environment: FormFieldEnvironment()
-            )
-            
         case .present(let isPresented):
             var model = state
             model.isPresented = isPresented
+            model.failUnfollowErrorMessage = nil
+            model.failFollowErrorMessage = nil
             
             if isPresented {
                 do {
@@ -168,29 +132,59 @@ struct AddressBookModel: ModelProtocol {
             
             return Update(state: model, fx: fx)
             
-        case .requestFollow:
+        case .populate(let follows):
             var model = state
-            // Show errors on any untouched fields, hints at why you cannot submit
-            model.followUserForm.did = FormField.update(
-                state: model.followUserForm.did,
-                action: .markAsTouched,
-                environment: FormFieldEnvironment()
-            ).state
-            model.followUserForm.petname = FormField.update(
-                state: model.followUserForm.petname,
-                action: .markAsTouched,
-                environment: FormFieldEnvironment()
-            ).state
+            model.follows = follows
+            return Update(state: model)
             
+        case .presentFollowUserForm(let isPresented):
+            var model = state
+            
+            model.isFollowUserFormPresented = isPresented
+            
+            return update(
+                state: model,
+                actions: [
+                    .didField(.reset),
+                    .petnameField(.reset)
+                ],
+                environment: environment
+            )
+            
+        case .didField(let action):
+            return DidFieldCursor.update(
+                state: state,
+                action: action,
+                environment: FormFieldEnvironment()
+            )
+            
+        case .petnameField(let action):
+            return PetnameFieldCursor.update(
+                state: state,
+                action: action,
+                environment: FormFieldEnvironment()
+            )
+            
+       
+            
+        case .requestFollow:
+            return update(
+                state: state,
+                actions: [
+                    // Show errors on any untouched fields, hints at why you cannot submit
+                    .didField(.markAsTouched),
+                    .petnameField(.markAsTouched),
+                    .attemptFollow
+                ],
+                environment: environment
+            )
+            
+        case .attemptFollow:
             guard let did = state.followUserForm.did.validated else {
-                return Update(state: model)
+                return Update(state: state)
             }
             guard let petname = state.followUserForm.petname.validated else {
-                return Update(state: model)
-            }
-            // Guard against duplicates
-            guard !state.follows.contains(where: { entry in entry.did == did }) else {
-                return Update(state: model)
+                return Update(state: state)
             }
             
             let fx: Fx<AddressBookAction> = environment.data.addressBook
@@ -205,7 +199,7 @@ struct AddressBookModel: ModelProtocol {
                 })
                 .eraseToAnyPublisher()
             
-            return Update(state: model, fx: fx)
+            return Update(state: state, fx: fx)
             
         case .succeedFollow(did: let did, petname: let petname):
             let entry = AddressBookEntry(pfp: Image("sub_logo_dark"), petname: petname, did: did)
@@ -213,6 +207,7 @@ struct AddressBookModel: ModelProtocol {
             var model = state
             model.isFollowUserFormPresented = false
             model.follows.append(entry)
+            
             return Update(state: model)
             
         case .failFollow(error: let error):
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
@@ -29,14 +29,27 @@ class PlaceholderSphereIdentityProvider: SphereIdentityProtocol {
     }
 }
 
+struct FollowUserFormModel: Equatable {
+    var did: FormField<String, Did> = FormField(value: "", validate: Self.validateDid)
+    var petname: FormField<String, Petname> = FormField(value: "", validate: Self.validatePetname)
+    
+    static func validateDid(key: String) -> Did? {
+        Did(key)
+    }
+
+    static func validatePetname(petname: String) -> Petname? {
+        Petname(petname)
+    }
+}
+
 struct PetnameFieldCursor: CursorProtocol {
     static func get(state: AddressBookModel) -> FormField<String, Petname> {
-        state.petnameField
+        state.followUserForm.petname
     }
     
     static func set(state: AddressBookModel, inner: FormField<String, Petname>) -> AddressBookModel {
         var model = state
-        model.petnameField = inner
+        model.followUserForm.petname = inner
         return model
     }
     
@@ -47,12 +60,12 @@ struct PetnameFieldCursor: CursorProtocol {
 
 struct DidFieldCursor: CursorProtocol {
     static func get(state: AddressBookModel) -> FormField<String, Did> {
-        state.didField
+        state.followUserForm.did
     }
     
     static func set(state: AddressBookModel, inner: FormField<String, Did>) -> AddressBookModel {
         var model = state
-        model.didField = inner
+        model.followUserForm.did = inner
         return model
     }
     
@@ -83,16 +96,7 @@ struct AddressBookModel: ModelProtocol {
     var follows: [AddressBookEntry] = []
     
     var followUserFormIsPresented = false
-    var didField: FormField<String, Did> = FormField(value: "", validate: Self.validateDid)
-    var petnameField: FormField<String, Petname> = FormField(value: "", validate: Self.validatePetname)
-    
-    static func validateDid(key: String) -> Did? {
-        Did(key)
-    }
-
-    static func validatePetname(petname: String) -> Petname? {
-        Petname(petname)
-    }
+    var followUserForm = FollowUserFormModel()
     
     var failFollowErrorMessage: String? = nil
     var failUnfollowErrorMessage: String? = nil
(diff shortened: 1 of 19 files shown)


xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -11,34 +11,38 @@ import SwiftUI
 struct ValidatedTextField: View {
     var placeholder: String
     @Binding var text: String
+    var onFocusChanged: ((Bool) -> Void)?
     var caption: String
-    var isValid: Bool = true
-
+    var hasError: Bool = false
+
```
bfollington, line 17: Validation and error presentation are seperate concerns here, since we should suppress errors on fields the user hasn't yet edited BUT those fields are technically invalid. Once they've been touched (or the user tries to submit the form without entering any values) _then_ we should actually display the error state.

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -9,6 +9,57 @@ import os
...
+    var petname: FormField<String, Petname> = FormField(value: "", validate: Self.validatePetname)
+    
+    static func validateDid(key: String) -> Did? {
+        Did(key)
+    }
+
+    static func validatePetname(petname: String) -> Petname? {
+        Petname(petname)
+    }
+}
```
bfollington, line 62: There's an argument for flattening these fields into the parent `AddressBookModel` instead of nesting.

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -61,25 +166,111 @@ struct AddressBookModel: ModelProtocol {
...
+                    .collect()
+                    .eraseToAnyPublisher()
+                }
+                .catch({ error in
+                    Just([])
+                })
+                .map({ follows in
+                    AddressBookAction.populate(follows)
+                })
+                .eraseToAnyPublisher()
```
bfollington, line 192: We'll be going via a SQLite table instead of directly populating the model, but this logic has to exist somewhere in the pipeline to produce the data format we're after.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -6,39 +6,47 @@
...
 /// A text field that comes with help text and a validation flag
 struct ValidatedTextField: View {
     var placeholder: String
     @Binding var text: String
+    var onFocusChanged: ((Bool) -> Void)?
     var caption: String
-    var isValid: Bool = true
-
+    var hasError: Bool = false
+    @FocusState var focused: Bool
```
bfollington, line 18: Despite technically being a two-way binding, this is only used to subscribe to the focus value.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -0,0 +1,108 @@
...
+//  AddressBookService.swift
+//  Subconscious
+//
+//  Created by Ben Follington on 14/3/2023.
+//
+
+import Foundation
+import Combine
+// temp
+import SwiftUI
```
bfollington, line 11: We can drop this import when I stop using `Image("dog-pfp")` as a placeholder.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -0,0 +1,108 @@
...
+                })
+                .collect()
+                .eraseToAnyPublisher()
+            }
+            .catch { err in
+                Just([])
+            }
+            .map { entries in
+                // Here be dragons (mutation) to cache the result
+                self.addressBook = entries
```
bfollington, line 51: Is using a class + simple mutation like this the "Swift" way to cache in-memory?

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -0,0 +1,108 @@
...
+                        .map { did -> AddressBookEntry? in
+                            guard let did = did else { return nil }
+                            // TODO: hardcoded dog-pfp
+                            return AddressBookEntry(pfp: Image("dog-pfp"), petname: f, did: did)
+                        }
+                        .compactMap { $0 }
+                })
+                .collect()
+                .eraseToAnyPublisher()
+            }
```
bfollington, line 45: I have tested that this code works as expected using dummy data.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -239,4 +239,40 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
...
+        try queue.sync {
+            try self.sphere().listPetnames()
+        }
+    }
+    
+    func getPetnameChanges(sinceCid: String) throws -> [String] {
+        try queue.sync {
+            try self.sphere().getPetnameChanges(sinceCid: sinceCid)
+        }
+    }
```
bfollington, line 277: Not 100% clear if this is the right spot for these, they need to access the underlying `sphere` so it feel reasonable.

xcode/Subconscious/Shared/Services/Noosphere/SphereFS.swift
```diff
@@ -264,6 +264,74 @@ public final class SphereFS: SphereProtocol {
         })
     }
     
+    // TODO: Do these belong on SphereFS?
```
bfollington, line 267: Similar question RE: is this the right place for these methods, they make sense as part of `SphereFS` today but as we know that abstraction is going away in the medium term.

xcode/Subconscious/Shared/Services/AddressBookService.swift
gordonbrander: Yup, this is the way.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
gordonbrander: Yes, this seems right to me.

xcode/Subconscious/Shared/Services/Noosphere/SphereFS.swift
gordonbrander: If I understand @cdata, address book belongs to a sphere, so I believe this is the right place to put these.

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -61,25 +166,93 @@ struct AddressBookModel: ModelProtocol {
...
             
-        case .follow(did: let did, petname: let petname):
+            return Update(state: model, fx: fx)
+            
+        case .requestFollow:
+            var model = state
+            // Show errors on any untouched fields, hints at why you cannot submit
+            model.followUserForm.did = FormField.update(
```
gordonbrander, line 181: Same as above. Should be:

```swift
return update(
    state: model,
    actions: [
        .didField(.markAsTouched),
        .petnameField(.markAsTouched)
    ],
    environment: environment
)
```

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -61,25 +166,93 @@ struct AddressBookModel: ModelProtocol {
...
+                action: .markAsTouched,
+                environment: FormFieldEnvironment()
+            ).state
+            model.followUserForm.petname = FormField.update(
+                state: model.followUserForm.petname,
+                action: .markAsTouched,
+                environment: FormFieldEnvironment()
+            ).state
+            
+            guard let did = state.followUserForm.did.validated else {
```
gordonbrander, line 192: Comment: I generally try not to reach down more than one level to prevent tight coupling. But I think this is fine here. No change needed.

xcode/Subconscious/Shared/Components/AddressBook/AddressBookView.swift
```diff
@@ -60,9 +60,9 @@ struct AddressBookView: View {
                     }
                 }
                 ToolbarItem(placement: .confirmationAction) {
-                    NavigationLink(
-                        destination: {
-                            FollowUserView(state: state, send: send)
+                    Button(
+                        action: {
+                            send(.presentFollowUserForm(true))
```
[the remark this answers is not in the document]
gordonbrander: Perfect. The sheet `presented` state should be owned by the view that owns the sheet, which it is here. No notes.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -6,39 +6,47 @@
...
                 )
+                .focused($focused)
                 .overlay(alignment: .trailing) {
                     VStack {
                         Image(systemName: "exclamationmark.circle")
                             .frame(width: 24, height: 22)
                             .padding(.horizontal, 8)
                             .foregroundColor(.red)
-                            .background(Color.background)
+                            .background(Color.clear)
```
gordonbrander, line 35: We want `Color.background`, because it causes text that would run underneath the icon to be covered with the background, so we don't get text under icon.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -6,39 +6,47 @@
 //
 
 import SwiftUI
+import Combine
 
 /// A text field that comes with help text and a validation flag
 struct ValidatedTextField: View {
     var placeholder: String
     @Binding var text: String
+    var onFocusChanged: ((Bool) -> Void)?
```
gordonbrander, line 15: Comment: I typically prefer to require a closure so I don't have to use unwrapping syntax. It's easy to provide an empty closure for no-op.

No change necessary.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -0,0 +1,108 @@
...
+        return self.listPetnamesAsync()
+            .flatMap { follows in
+                return Publishers.MergeMany(follows.map { f in
+                    return self.getPetnameAsync(petname: f)
+                        .map { did -> AddressBookEntry? in
+                            guard let did = did else { return nil }
+                            // TODO: hardcoded dog-pfp
+                            return AddressBookEntry(pfp: Image("dog-pfp"), petname: f, did: did)
+                        }
+                        .compactMap { $0 }
```
gordonbrander, line 41: Comment: I typically write `.compactMap({ value in value })` just because it is more familiar to people who might not know the Swift sugars.

No change necessary.

xcode/Subconscious/Shared/Components/AddressBook/FollowUserView.swift
```diff
@@ -10,68 +10,82 @@ import ObservableStore
...
+                                placeholder: "DID",
+                                text: Binding(
+                                    get: { form.did.value },
+                                    send: send,
+                                    tag: { v in .didField(.setValue(input: v))}
+                                ),
+                                onFocusChanged: { focused in
+                                    send(.didField(.focusChange(focused: focused)))
+                                },
+                                caption: "e.g. did:key:z6MkmCJAZansQ3p1Qwx6wrF4c64yt2rcM8wMrH5Rh7DGb2K7",
```
gordonbrander, line 36: What does this point to? Anything?

gordonbrander (changes requested):

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
bfollington: Ah, of course. The problem is that in a form field on the dark color scheme the background color of the input isn't `Color.background`. 

Maybe there's another constant I need to read... will look into it.

xcode/Subconscious/Shared/Components/AddressBook/FollowUserView.swift
bfollington: I believe it's my testing sphere address, but I might've changed a couple characters. I intend to make a new one before publishing anything other than test data.

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -49,6 +114,46 @@ struct AddressBookModel: ModelProtocol {
...
+        case .populate(let follows):
+            var model = state
+            model.follows = follows
+            return Update(state: model)
+            
+        case .presentFollowUserForm(let isPresented):
+            var model = state
+            
+            model.followUserFormIsPresented = isPresented
+            if isPresented {
```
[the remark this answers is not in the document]
bfollington: Thankyou! This didn't feel right as I was writing it. Will refactor 👍

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -18,6 +19,7 @@ struct AddressBookEntry: Equatable {
 
 struct AddressBookEnvironment {
     var noosphere: SphereIdentityProtocol
```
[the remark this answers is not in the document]
bfollington: Removed 👍

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -27,16 +29,79 @@ class PlaceholderSphereIdentityProvider: SphereIdentityProtocol {
...
+    case didField(FormFieldAction<String>)
+    case petnameField(FormFieldAction<String>)
 }
 
 struct AddressBookModel: ModelProtocol {
     var isPresented = false
     var did: Did? = nil
     var follows: [AddressBookEntry] = []
+    
+    var followUserFormIsPresented = false
```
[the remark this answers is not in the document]
bfollington: Changed 👍

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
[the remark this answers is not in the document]
bfollington: Refactored, I introduced a new action `.attemptFollow` as an intermediate step. So it flows from:

`.requestFollow` -> `.attemptFollow` then either `.succeedFollow` or `.failFollow`

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
bfollington: Refactored 👍 

I introduced a new action `.attemptFollow` with `update` as an intermediate step to clean up the logic. So it flows from:

`.requestFollow` -> `.attemptFollow` then either `.succeedFollow` or `.failFollow`

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
bfollington: I agree, in a more complex form I think there's value in a `.validated` property (on `followUserForm` in this case) that wraps the entire form payload into a single validated struct. That's often convenient but alsos avoid reaching down two levels.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
bfollington: Okay, so when appearing in a `Form` the background colour of a field should be `UIColor.secondarySystemGroupedBackground`, _but_ during onboarding we also present a textfield outside of a `Form`. 

To accomodate both cases I've added a `.formField()` method to `ValidatedTextField` that will use the appropriate color.

gordonbrander (approved):
LGTM

PR #440, opened 2023-03-09
Improve omnibox with 3p @handles, animations, untitled state

### LINKED ISSUE #437: Better empty state for draft notes
Followup on https://github.com/subconsciousnetwork/subconscious/pull/427#discussion_r1130229235

gordonbrander:
Fixes #437. 

This PR introduces improvements to Omnibox:

- Displays specially formatted `@handles/slashlinks` when petname present
- Animates changes to location with fade animation
- Introduces an untitled state for empty notes

### DIFF
--- a/xcode/Subconscious/Shared/Components/OmniboxView.swift
+++ b/xcode/Subconscious/Shared/Components/OmniboxView.swift
@@ -8,27 +8,32 @@
 import SwiftUI
 
 struct OmniboxView: View {
-    var icon: Image
-    var title: String
+    var address: MemoAddress?
+    var defaultAudience: Audience
 
     var body: some View {
         HStack(spacing: 8) {
-            icon
+            Image(audience: address?.toAudience() ?? defaultAudience)
                 .resizable()
                 .frame(width: 17, height: 17)
-                .foregroundColor(.accentColor)
-            HStack(spacing: 0) {
-                Spacer()
-                Text(verbatim: title)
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Components/OmniboxView.swift
```diff
@@ -37,30 +42,80 @@ struct OmniboxView: View {
...
+        @State private var address: MemoAddress? = nil
+        var defaultAudience = Audience.local
+
+        var body: some View {
+            OmniboxView(address: address, defaultAudience: defaultAudience)
+                .onTapGesture {
+                    withAnimation {
+                        self.address = addresses.randomElement()
+                    }
+                }
```
bfollington, line 83: Question: in general on iOS should there be an "active" state for UI controls? i.e. change in color or opacity while the pointer is down but before it triggers a "tap" event.

There's no indication that I am tapping on the omnibox when I do it but obviously **this** instance is just a preview.

bfollington (approved):
Playtested in simulator and it's looking good! I'm impressed by how smart the transitions are for handling the petname vs link components.

xcode/Subconscious/Shared/Components/OmniboxView.swift
gordonbrander: There is a basic default tap state when run on-device (button dims when pressing). To me this feels right for this control.

PR #441, opened 2023-03-09
Fix failing data service destructor test

gordonbrander:
This test was written while we were troubleshooting the issue with Noosphere not flushing writes (fixed in https://github.com/subconsciousnetwork/noosphere/pull/231).

TLDR: this test was written to repro an issue with Noosphere losing data at a time when we were not sure where the issue was coming from. It should be updated to pass now.

## Background

At the time it was unclear where the issue was cropping up, so I wrote tests on the Subconscious side to repro.

We now know that before https://github.com/subconsciousnetwork/noosphere/pull/231, Noosphere was not flushing writes unless `ns_sphere_fs_free(fs)`, `ns_free(noosphere)` were called. We were calling these cleanup functions in the corresponding Swift wrapper class `deinit`. However, while `deinit` is always called in Swift, it is not guaranteed to be called until end of scope. Additionally, it will not be called in the case of a crash. So while we were doing as much cleanup as possible on the Swift side, we could not guarantee destructors would be called if app crashed. This put users at risk of losing a session's worth of data if the app crashed.

https://github.com/subconsciousnetwork/noosphere/pull/231 fixed the issue by flushing-on-save, instead of flushing on cleanup.

The `deinit` still run and clean up properly during normal app suspensions. Basically we do everything we can do on the Swift side to clean up properly, within the limits of the guarantees that the Swift language makes about calling `deinit`.

## Fix

This fix updates the test so that the relevant portions run within a block. This causes the destructor of the first Noosphere instance to be called before the next instance is opened.

### DIFF
--- a/xcode/Subconscious/SubconsciousTests/Tests_DataService.swift
+++ b/xcode/Subconscious/SubconsciousTests/Tests_DataService.swift
@@ -167,55 +167,117 @@ final class Tests_DataService: XCTestCase {
     
     func testManyWritesThenCloseThenReopen() throws {
         let tmp = try createTmpDir()
-        var data = try createDataService(tmp: tmp)
+
+        let globalStorageURL = tmp.appending(path: "noosphere")
+        let sphereStorageURL = tmp.appending(path: "sphere")
         
-        let versionX = try data.noosphere.version()
-        print("!!! X", versionX)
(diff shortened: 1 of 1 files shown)


jsantell (approved):
Looks great! Thanks for this quick fix

PR #445, opened 2023-03-10
chore: initial work wiring up CI on xcode tests

jsantell:


scripts/run_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+  sudo xcode-select -switch $XCODE_PATH && /usr/bin/xcodebuild -version
+fi
+
+DESTINATION="platform=$PLATFORM"
+DESTINATION="$DESTINATION,name=$DEVICE"
+DESTINATION="$DESTINATION,OS=$OS"
+  
+# Run tests
+xcodebuild test -project xcode/Subconscious/Subconscious.xcodeproj \
+  -scheme "Subconscious (iOS)" \
```
bfollington, line 24: I tried to come up with a Scheme that _doesn't_ launch the simulator to run our tests. They run within Xcode during local dev without opening the simulator... but I had no luck. Maybe @gordonbrander knows more?

We could save a little CI time if we didn't run the `iOS` specific tests like `testLaunchPerformance` but I'm not sure that's worth it.

bfollington (commented):
LGTM, I'll wait for @gordonbrander's input before approving

scripts/run_tests.sh
bfollington: I've disabled `testLaunchPerformance` which trimmed 2 minutes off the CI workflow.

jsantell: Fantastic!

bfollington (approved):
After chatting today we all (@gordonbrander, @jsantell and I) agreed to merge this. There are future improvements to come, but this is already useful.

Issue #448, opened 2023-03-16
Integrate Noosphere Address Book API with AddressBookService


bfollington:
Having completed https://github.com/subconsciousnetwork/subconscious/issues/447 the next step is to actually plug everything in and see it come alive. Now that https://github.com/subconsciousnetwork/noosphere/pull/271 has landed we're ready to update the package in Subconscious and un-stub the calls.

PR #449, opened 2023-03-16
Integrate Address Book API

bfollington:
This PR introduces support for reading and writing address book content from the local sphere. The `AddressBookService` is responsible for providing an abstraction over the underlying petname storage.

# Added
- Confirmation for unfollow
- Human readable errors
- Error alert popups
- Actual FFI calls
- Extracted C array -> Swift array utility

Resolves https://github.com/subconsciousnetwork/subconscious/issues/448

# Tasks
- [x] Confirmation dialog for unfollowing a user
- [x] Error popups
- [x] Implement smarter caching & refetch logic 
- [x] Improve error handling, human readable errors
    - https://www.avanderlee.com/swiftui/error-alert-presenting/

### DIFF
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
@@ -76,10 +76,14 @@ enum AddressBookAction {
     case requestFollow
     case attemptFollow
     case failFollow(error: String)
+    case dismissFailFollowError
     case succeedFollow(did: Did, petname: Petname)
     
     case requestUnfollow(petname: Petname)
+    case confirmUnfollow
+    case cancelUnfollow
     case failUnfollow(error: String)
+    case dismissFailUnfollowError
     case succeedUnfollow(petname: Petname)
     
     case presentFollowUserForm(_ isPresented: Bool)
@@ -97,6 +101,8 @@ struct AddressBookModel: ModelProtocol {
     
     var failFollowErrorMessage: String? = nil
     var failUnfollowErrorMessage: String? = nil
+    
+    var unfollowCandidate: Petname? = nil
 
     static let logger = Logger(
         subsystem: Config.default.rdns,
@@ -219,7 +225,26 @@ struct AddressBookModel: ModelProtocol {
             model.failFollowErrorMessage = error
             return Update(state: model)
             
+        case .dismissFailFollowError:
+            var model = state
+            model.failFollowErrorMessage = nil
+            return Update(state: model)
+            
         case .requestUnfollow(petname: let petname):
+            var model = state
+            model.unfollowCandidate = petname
+            return Update(state: model)
+            
+        case .cancelUnfollow:
+            var model = state
+            model.unfollowCandidate = nil
+            return Update(state: model)
+            
+        case .confirmUnfollow:
+            guard let petname = state.unfollowCandidate else {
+                return Update(state: state)
+            }
+            
             let fx: Fx<AddressBookAction> =
                 environment.data.addressBook
                 .unfollowUserAsync(petname: petname)
@@ -245,6 +270,11 @@ struct AddressBookModel: ModelProtocol {
         case .failUnfollow(error: let error):
             var model = state
             model.failUnfollowErrorMessage = error
+            return Update(state: model)
+            
+        case .dismissFailUnfollowError:
+            var model = state
+            model.failUnfollowErrorMessage = nil
             return Update(state: model)
         }
     }
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
@@ -70,7 +70,7 @@ struct DidFieldCursor: CursorProtocol {
 enum AddressBookAction {
     case present(_ isPresented: Bool)
     
-    case refetch
+    case refreshEntries(forceRefetchFromNoosphere: Bool = false)
     case populate([AddressBookEntry])
     
     case requestFollow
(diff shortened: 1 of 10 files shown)


xcode/Subconscious/Shared/Services/Noosphere/SphereFS.swift
```diff
@@ -449,17 +463,23 @@ public final class SphereFS: SphereProtocol {
...
         }
 
-        let name_count = file_header_names.len
-        guard var pointer = file_header_names.ptr else {
+        return file_header_names.toSwift()
+    }
+}
+
+extension slice_boxed_char_ptr_t {
+    func toSwift() -> [String] {
```
bfollington, line 471: This seems like a bad name but I struggled to find a better one.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -54,6 +56,32 @@ class AddressBookService {
             .eraseToAnyPublisher()
     }
     
+    func followUser(did: Did, petname: Petname) throws {
+        try setPetname(did: did, petname: petname)
+        let version = try self.noosphere.save()
+        try database.writeMetadatadata(key: .sphereVersion, value: version)
```
bfollington, line 62: Unsure if writing directly to the database like this is encouraged, or if we should instead move these methods to `NoosphereService` which already manages `.sphereVersion`?

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -209,17 +230,40 @@ struct AddressBookModel: ModelProtocol {
...
+            
+        case .confirmUnfollow:
+            guard let petname = state.unfollowCandidate else {
+                return Update(state: state)
+            }
+            
             let fx: Fx<AddressBookAction> =
                 environment.data.addressBook
-                .unsetPetnameAsync(petname: petname)
+                .unfollowUserAsync(petname: petname)
```
bfollington, line 266: Higher-level concept: following is user-level and petname is unset is an implementation detail

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -187,9 +200,17 @@ struct AddressBookModel: ModelProtocol {
...
+                    state: state,
+                    action: .failFollow(error: "Already following user"),
+                    environment: environment
+                )
+            }
+            
             let fx: Fx<AddressBookAction> =
                 environment.data.addressBook
-                .setPetnameAsync(did: did, petname: petname)
+                .followUserAsync(did: did, petname: petname)
```
bfollington, line 213: Higher-level concept: `follow` is user-level and `set` is an implementation detail

xcode/Subconscious/Shared/Components/AddressBook/AddressBookView.swift
```diff
@@ -92,6 +80,28 @@ struct AddressBookView: View {
             ) {
                 FollowUserView(state: state, send: send)
             }
+            .alert(
+                state.failUnfollowErrorMessage ?? "",
+                isPresented: Binding(
+                    get: { state.failUnfollowErrorMessage != nil },
+                    set: { _ in send(.dismissFailUnfollowError) }
+                )
+            ) {}
```
gordonbrander, line 89: Q: is this alert doing anything? The view portion of the alert seems to be empty.

xcode/Subconscious/Shared/Services/AddressBookService.swift
gordonbrander: I think this is fine for now. We could look later into whether the relationship between services should be refactored in #462.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -54,6 +56,32 @@ class AddressBookService {
...
+    }
+    
+    /// Associates the passed DID with the passed petname within the sphere, saves the changes and updates the database.
+    func followUserAsync(did: Did, petname: Petname) -> AnyPublisher<Void, Error> {
+        CombineUtilities.async(qos: .utility) {
```
gordonbrander, line 67: Nit: might be better to use a default QOS, since this is a user action.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -54,6 +56,32 @@ class AddressBookService {
...
+    
+    func unfollowUser(petname: Petname) throws {
+        try unsetPetname(petname: petname)
+        let version = try self.noosphere.save()
+        try database.writeMetadatadata(key: .sphereVersion, value: version)
+    }
+    
+    /// Unassociates the passed DID with the passed petname within the sphere, saves the changes and updates the database.
+    func unfollowUserAsync(petname: Petname) -> AnyPublisher<Void, Error> {
+        CombineUtilities.async(qos: .utility) {
```
gordonbrander, line 80: Nit: might be better to use a default QOS, since this is a user action.

xcode/Subconscious/Shared/Services/Noosphere/SphereFS.swift
gordonbrander: `.toStringArray()` ?

gordonbrander (approved):
LGTM

2 questions, one note. Please feel free to address, or not, and merge.

xcode/Subconscious/Shared/Services/Noosphere/SphereFS.swift
bfollington: Much better, thanks!

xcode/Subconscious/Shared/Components/AddressBook/AddressBookView.swift
bfollington: This works fine (renders a simple message + OK button) but I found a better function signature for the job. We can present both a title and message this way too: 

image

PR #450, opened 2023-03-16
Flip FAB light/dark style

gordonbrander:
The red-purple-cyan fab should be displayed in light mode. The pink-violet-cyan fab should be displayed in dark mode.

Also:

- Tweaked text style (brand black for pink-violet-cyan fab)
- Left shadows the same. I think it looks really good to allow the purple color to bleed into the white background.


bfollington (approved):
Whoops, didn't realise I inverted it but it makes sense now I think about it. 

LGTM.

PR #452, opened 2023-03-16
Expand transclude hit area to include empty space

### LINKED ISSUE #446: Backlinks are only tappable on text
Taps on backlinks in the editor sheet are only recognised within the pink area highlighted:

image

Where-as I would expect a tap anywhere in the yellow area to work:

image

gordonbrander:
Additionally, add background wash on tap.

Fixes #446.


bfollington (approved):
LGTM, tested in the simulator.

PR #453, opened 2023-03-16
Fix Subtext perf regression, add Subtext perf test and baseline

### LINKED ISSUE #447: Investigate possible Subtext parser regression
@gordonbrander has noted that after refactoring the Subtext parser to use Regex (in https://github.com/subconsciousnetwork/subconscious/pull/422) there may have been a performance regression, despite it on-paper being faster.

We should perform profiling and compare timings to the old implementation to pinpoint the issue (if any) and address bottlenecks.

gordonbrander:
Fixes #447.

This commit replaces the Regex-based matches with the old faster hand-rolled
recursive descent logic from https://github.com/subconsciousnetwork/subconscious/commit/16709dfa95a99d37c8ce53423e2f1a9e4c3a7c5a.

## Background

Profiling the newer regex-based Subtext parser logic against the old hand-rolled non-regex recursive descent logic revealed that the non-regex approach is 97% faster than regex approach (0.013s vs 0.381s). Not sure why regex is so slow here, but we can put up with the inconvenience of a hand-written recursive descent parser in exchange for the sizable perf win.

## Changes

- Replaces subtext parsing logic with faster recursive descent logic from https://github.com/subconsciousnetwork/subconscious/commit/16709dfa95a99d37c8ce53423e2f1a9e4c3a7c5a
- Adds perf test and baseline

All tests pass, with one updated test. We now accept `@bare-handles-without-paths` because it would be a bit of a pain to do backtracking. The petname parsing code for URLs rejects bare petnames right now, so the user-facing behavior is the same. And we'll want to support petname-only slashlinks at some point anyway.

### DIFF
--- a/xcode/Subconscious/Shared/Parsers/Subtext.swift
+++ b/xcode/Subconscious/Shared/Parsers/Subtext.swift
@@ -188,17 +188,91 @@ struct Subtext: Hashable, Equatable, LosslessStringConvertible {
         }
     }
 
-    /// Check if character is space
-    /// Currently this matches only against a minimal set of space characters.
-    /// For convenience, character is an optional type, since peek
-    /// returns an optional type.
-    private static func isSpace(character: Character?) -> Bool {
-        switch character {
-        case " ", "\n":
-            return true
-        default:
-            return false
+    /// Consume a well-formed bracket link, or else backtrack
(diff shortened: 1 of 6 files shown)


bfollington (approved):
LGTM, 10x performance is always a welcome change

PR #458, opened 2023-03-16
Introduce scalable units

### LINKED ISSUE #435: Detail editor does not respect system font size settings
Increasing (or decreasing) the base font on iOS causes the in-app text to resize correctly in all components _except_ the TextView. We're hardcoding the font-size as an attribute which seems to prevent dynamic resize, even though we use specific font sizes elsewhere in SwiftUI and _those_ adapt just fine.

# Default
image

## Large Text
image

### LINKED ISSUE #436: Define dynamic unit that is responsive to system font size
There are cases where we need to hard-code dimensions, yet have the item scale with system font size. See https://github.com/subconsciousnetwork/subconscious/pull/427#discussion_r1130213274

Currently we only have `AppTheme.unit`, which is hard-coded. We should additionally define a dynamic unit, based on system font size, sort of like `rem`.

gordonbrander:
Fixes #436
Fixes #435

### DIFF
--- a/xcode/Subconscious/Shared/AppTheme.swift
+++ b/xcode/Subconscious/Shared/AppTheme.swift
@@ -7,15 +7,13 @@
 
 import SwiftUI
 
-enum AppTheme {}
-
 //  MARK: Theme units
-extension AppTheme {
+enum Unit {
     static let unit: CGFloat = 4
-    static let unitHalf: CGFloat = unit / 2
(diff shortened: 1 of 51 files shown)


gordonbrander:
Turns out this approach is a dud. `@ScaledMetric` can only be used within a view.

We should keep units as-is, and used `@ScaledMetric` to adjust them on a per-view basis.

PR #461, opened 2023-03-17
Scale editor text and audience menu button with preferred font size

### LINKED ISSUE #435: Detail editor does not respect system font size settings

### LINKED ISSUE #436: Define dynamic unit that is responsive to system font size

gordonbrander:
Fixes #436
Fixes #435

Uses https://developer.apple.com/documentation/swiftui/scaledmetric property wrapper to define scalable units. 

## Approach

`@ScaledMetric` is a view property wrapper. We keep the same `AppTheme.unit` constants, but use the property wrapper in particular views to scale these units as needed.

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
+++ b/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
@@ -173,7 +173,6 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
         NSTextContentManagerDelegate,
         NSTextLayoutManagerDelegate
     {
-        static let renderer = SubtextAttributedStringRenderer()
         /// Is event happening during updateUIView?
         /// Used to avoid setting properties in events during view updates, as
         /// that would cause feedback cycles where an update triggers an event,
@@ -181,6 +180,8 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
         var isUIViewUpdating: Bool
         var representable: SubtextTextViewRepresentable
         var renderTranscludeBlocks: Bool = false
+        /// Subtext renderer instance
+        var renderer: SubtextAttributedStringRenderer
         var subtext: Subtext? = nil
 
         init(
@@ -188,6 +189,9 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
         ) {
             self.isUIViewUpdating = false
             self.representable = representable
+            self.renderer = SubtextAttributedStringRenderer(
+                bodySize: representable.bodySize
+            )
         }
         
         /// NSTextStorageDelegate method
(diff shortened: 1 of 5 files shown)


xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -173,21 +173,25 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
...
-        static let renderer = SubtextAttributedStringRenderer()
         /// Is event happening during updateUIView?
         /// Used to avoid setting properties in events during view updates, as
         /// that would cause feedback cycles where an update triggers an event,
         /// which triggers an update, which triggers an event, etc.
         var isUIViewUpdating: Bool
         var representable: SubtextTextViewRepresentable
         var renderTranscludeBlocks: Bool = false
+        /// Subtext renderer instance
+        var renderer: SubtextAttributedStringRenderer
```
gordonbrander, line 184: Make renderer an instance variable. We can change its body size dynamically in `updateUIView` now.

xcode/Subconscious/SubconsciousTests/Tests_SubtextAttributedStringRenderer.swift
```diff
@@ -8,7 +8,7 @@
 import XCTest
 @testable import Subconscious
 
-final class Tests_SubtextAttributedStringRendererToURL: XCTestCase {
+final class Tests_SubtextAttributedStringRenderer: XCTestCase {
```
gordonbrander, line 11: Fix class name

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -389,9 +395,17 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
...
+        // preferences changed.
+        if (
+            view.text != state.text ||
+            self.bodySize != context.coordinator.renderer.bodySize
+        ) {
             SubtextTextViewRepresentable.logger.debug("updateUIView: set text")
+            // Set body size on renderer. This costs nothing, and makes sure
+            // that the next text render will have the new body size if the
+            // body size preferenced changed.
+            context.coordinator.renderer.bodySize = bodySize
```
bfollington, line 408: Oof, of course we have to do this. I guess it's better to have solved this now before we get any further into the the transclude block work.

bfollington (approved):
Nice, pleasantly surprised by how clean `@ScaledMetric` is

Issue #472, opened 2023-03-27
`main` should be an evergreen branch


gordonbrander:
## TODO

- [x] Flag for Noosphere #471 fixed by #473
- [x] Lock SHA for reproducible builds https://github.com/subconsciousnetwork/subconscious/pull/476
- [x] Add branch protection to enforce linear history
- [x] CI tests https://github.com/subconsciousnetwork/subconscious/pull/445

## Background motivation

Over the weekend, our TestFlight build expired. I was able to bump the build number and redeploy. However, the event exposed a number of issues with the way we manage `main` branch.

### Postmortem

- Database rebuild logic was broken in the last deployed tag [0.0.7](https://github.com/subconsciousnetwork/subconscious/releases/tag/0.0.7%2B1). We use this logic to migrate the database to new schemas.
- We fixed the database rebuild logic in https://github.com/subconsciousnetwork/subconscious/commit/5897c0cb4158201f33472ad4a890d2fca396e30a
    - Root cause was a bug related to `FileManager.default.removeItem(atPath:)`, and was fixed by using `FileManager.default.removeItem(at:)` instead. [See code for fix](https://github.com/subconsciousnetwork/subconscious/commit/5897c0cb4158201f33472ad4a890d2fca396e30a#diff-ab7ffc52a85ac83cd4ac02e124560b9d1000c363c30d6738e3cb520f0792a887R717).
- However https://github.com/subconsciousnetwork/subconscious/commit/5897c0cb4158201f33472ad4a890d2fca396e30a was also the PR where we introduced basic Noosphere features.
- We do not yet wish to roll out Noosphere to users, meaning we could not tag and roll out the PR with the database rebuild fix, or subsequent PRs.
- Redeploying 0.0.7 worked because there was no database rebuild to apply (same database version). However, this situation easily could have put us into a position where we needed to create a hotfix branch. Bad! 

In summary, our `main` branch violated the not-rocket-science rule. https://graydon2.dreamwidth.org/1597.html

> The Not Rocket Science Rule Of Software Engineering:
> automatically maintain a repository of code that always passes all the tests

### Other learnings

- TestFlight revokes the app when the build expires, but it DOES NOT delete the user's `Files/Subconscious` directory where local files are installed (good).
- Users are able to re-install the TestFlight without a new invite (it re-appears in your TestFlight app).
- Re-deploying same build goes through an expedited TestFlight review/deploy process. The build was available seconds after I bumped the build number.

### Solution going forward

[...]

PR #473, opened 2023-03-27
Add developer setting to toggle Noosphere enabled

### LINKED ISSUE #471: Allow Noosphere to be flagged off
Allowing Noosphere to be flagged off would let us tag and deploy to Testflight from `main`.

Part of umbrella issue #472

gordonbrander:
Fixes #471
Progress toward umbrella issue #472

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -229,12 +229,19 @@ struct AppModel: ModelProtocol {
     /// Is database connected and migrated?
     var databaseState = AppDatabaseState.initial
 
+    /// Has first run completed.
+    /// We assign user default to model property at startup.
+    /// This property is changed via `persistFirstRunComplete`, which will
+    /// update both the model property (triggering a view re-render)
+    /// and persist the new value to UserDefaults.
+    var isFirstRunComplete = AppDefaults.standard.firstRunComplete
+
     /// Should first run show?
     var shouldPresentFirstRun: Bool {
         guard AppDefaults.standard.noosphereEnabled else {
             return false
         }
-        return !AppDefaults.standard.firstRunComplete
+        return !isFirstRunComplete
     }
 
     var nickname = AppDefaults.standard.nickname
@@ -479,6 +486,12 @@ struct AppModel: ModelProtocol {
         logger.debug(
             "Sphere ID: \(sphereIdentity)"
         )
+        logger.debug(
+            "Sphere ID: \(sphereIdentity)"
+        )
+        logger.debug(
+            "Noosphere enabled? \(AppDefaults.standard.noosphereEnabled)"
+        )
         return update(
             state: state,
             actions: [
@@ -625,7 +638,9 @@ struct AppModel: ModelProtocol {
     ) -> Update<AppModel> {
         // Persist value
         AppDefaults.standard.firstRunComplete = isComplete
-        return Update(state: state).animation(.default)
+        var model = state
+        model.isFirstRunComplete = isComplete
+        return Update(state: model).animation(.default)
     }
 
     /// Reset NoosphereService managed instances of `Noosphere` and `SphereFS`.
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -237,7 +237,7 @@ struct AppModel: ModelProtocol {
     /// This property is changed via `persistNoosphereEnabled`, which will
     /// update both the model property (triggering a view re-render)
     /// and persist the new value to UserDefaults.
-    var isNoosphereEnabled = AppDefaults.standard.noosphereEnabled
+    var isNoosphereEnabled = AppDefaults.standard.isNoosphereEnabled
 
     /// Has first run completed?
     /// We assign UserDefault property to model property at startup.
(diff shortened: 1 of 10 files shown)


xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
```diff
@@ -29,13 +29,15 @@ struct DetailMetaSheet: View {
                         }
                     }
                     .font(.callout)
-                    AudienceMenuButtonView(
-                        audience: Binding(
-                            get: { state.audience },
-                            send: send,
-                            tag: DetailMetaSheetAction.requestUpdateAudience
+                    if AppDefaults.standard.noosphereEnabled {
```
bfollington, line 32: Shouldn't this be "isNoosphereEnabled" from the state?

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -348,7 +348,9 @@ final class DatabaseService {
         var special: [Suggestion] = []
         
         // Insert quick-create suggestion
-        special.append(.createPublicMemo(fallback: ""))
+        if AppDefaults.standard.noosphereEnabled {
```
bfollington, line 351: Shouldn't this also check the store value instead of the config flag?

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -373,7 +375,9 @@ final class DatabaseService {
         var suggestions: [Suggestion] = []
         
         // Create suggestions for the literal query
-        suggestions.append(.createPublicMemo(fallback: query))
+        if AppDefaults.standard.noosphereEnabled {
```
bfollington, line 378: Same again?

bfollington (changes requested):
I think we should pick one preferred way of checking Noosphere, either going via AppDefaults or the store as the way to read the value. It seems like it might be tricky to read from the state in a service though, so I can see a compromise there... but it doesn't seem ideal.

The store will cause a rerender when it changes which I expect is not true of the AppDefaults? But once again that's only relevant to views, not services.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -609,6 +646,21 @@ struct AppModel: ModelProtocol {
...
+    /// Persist Noosphere enabled state.
+    /// Note that this updates the model state (triggering a re-render),
+    /// and ALSO perists the state to UserDefaults.
+    static func persistNoosphereEnabled(
+        state: AppModel,
+        environment: AppEnvironment,
+        isEnabled: Bool
+    ) -> Update<AppModel> {
+        // Persist setting to UserDefaults
+        AppDefaults.standard.noosphereEnabled = isEnabled
```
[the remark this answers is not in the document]
gordonbrander: Rationale here is that UserDefaults is a synchronous API, and AppDefaults is already a service (of sorts).

Ideally we would move AppDefaults into Environment. However, doing so would require a change to ObservableStore that allows initial store state be generated from a factory with access to the environment. See https://github.com/subconsciousnetwork/ObservableStore/issues/28.

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
gordonbrander: Fixed

xcode/Subconscious/Shared/Library/UserDefaultsProperty.swift
```diff
@@ -27,14 +32,23 @@ import Foundation
...
+        // did not actually update the equatable value.
+        self.subject.removeDuplicates().eraseToAnyPublisher()
+    }
+
     var wrappedValue: Value {
         get {
             scope.value(forKey: key) as? Value ?? `default`
         }
-        set {
+        nonmutating set {
```
bfollington, line 49: TIL this keyword exist

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
```diff
@@ -29,13 +29,15 @@ struct DetailMetaSheet: View {
                         }
                     }
                     .font(.callout)
-                    AudienceMenuButtonView(
-                        audience: Binding(
-                            get: { state.audience },
-                            send: send,
-                            tag: DetailMetaSheetAction.requestUpdateAudience
+                    if AppDefaults.standard.isNoosphereEnabled {
```
bfollington, line 32: Even though there's no way for a user to toggle this setting while the sheet is up, I wanted to check if this re-renders when the `AppDefault` changes... and I was surprised to find it does re-render!?

Is this a byproduct of adding a publisher to these properties or do `AppDefaults` changes trigger magic renders?

bfollington (approved):
LGTM

xcode/Subconscious/Shared/Components/Detail/DetailMetaSheet.swift
gordonbrander: That's the publisher doing its thing. The subscription to `AppDefaults.standard.$isNoosphereEnabled` means we get an event sent to the store whenever the value changes. This then replays the value change to the store and updates the store's copy of the property, triggering re-render.

xcode/Subconscious/Shared/Library/UserDefaultsProperty.swift
gordonbrander: Yeah another Swift oddity. This lets us use setter syntax for let-bound variables that point to this property.

PR #475, opened 2023-03-28
Guard against following yourself

### LINKED ISSUE #474: You can follow yourself
This causes infinite recursive problems, and should be disallowed. 

> cdata: This is bringing to mind that we don't stop you from following yourself, and that this could cause an infinite loop as you update your address book to point to your latest sphere version, which causes a yet newer version to be published to the name system, etc.

When trying to follow yourself an error should be presented: "You cannot follow yourself"

Related: https://github.com/subconsciousnetwork/noosphere/issues/293

bfollington:
Fixes #474.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
@@ -211,7 +211,7 @@ struct AddressBookModel: ModelProtocol {
             guard !state.follows.contains(where: { f in f.did == did }) else {
                 return update(
                     state: state,
-                    action: .failFollow(error: "Already following user"),
+                    action: .failFollow(error: AddressBookError.alreadyFollowing.localizedDescription),
                     environment: environment
                 )
             }
--- a/xcode/Subconscious/Shared/Services/AddressBookService.swift
+++ b/xcode/Subconscious/Shared/Services/AddressBookService.swift
@@ -10,6 +10,25 @@ import Combine
 // temp
 import SwiftUI
 
+enum AddressBookError: Error {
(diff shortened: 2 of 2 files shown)


xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -211,7 +211,7 @@ struct AddressBookModel: ModelProtocol {
             guard !state.follows.contains(where: { f in f.did == did }) else {
                 return update(
                     state: state,
-                    action: .failFollow(error: "Already following user"),
+                    action: .failFollow(error: AddressBookError.alreadyFollowing.localizedDescription),
```
gordonbrander, line 214: Nice

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -10,6 +10,25 @@ import Combine
...
+    case cannotFollowYourself
+    case alreadyFollowing
+    case other(String)
+}
+
+extension AddressBookError: LocalizedError {
+    var errorDescription: String? {
+        switch self {
+        case .cannotFollowYourself:
+            return NSLocalizedString("You cannot follow yourself.", comment: "Address Book error description")
```
gordonbrander, line 23: We should use [`String(localized:)`](https://developer.apple.com/documentation/swift/string/init(localized:)) here instead. It's the new Swift-native API for this.

gordonbrander (changes requested):
Let's use `String(localized:)`. Otherwise, good to go.

gordonbrander (approved):

Issue #483, opened 2023-03-31
Implement sphere traversal FFI functions


gordonbrander:
We should update the changed function names and also wire up the bindings for traversal (from https://github.com/subconsciousnetwork/noosphere/issues/288) as part of this.

Followup to #477

PR #484, opened 2023-04-03
Integrate traversal FFI calls into NoosphereService

bfollington:
# Changes

- Added `traverse(petname: String) throws -> Sphere` to `SphereProtocol`
- Add private `Sphere.init` that constructs a `Sphere` from an `ns_sphere_t`
- Add `testSphereTraversal` (skipped test) to demonstrate usage

Resolves https://github.com/subconsciousnetwork/subconscious/issues/483

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -275,7 +275,7 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
         }
     }
 
-    func traverse(petname: String) throws -> Sphere? {
+    func traverse(petname: String) throws -> Sphere {
         try queue.sync {
             try self.sphere().traverse(petname: petname)
         }
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -141,19 +141,17 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
             throw NoosphereServiceError.defaultSphereNotFound
         }
         
-        return try open(identity: identity)
+        return try sphere(identity: identity)
     }
     
     /// Open a sphere using its identity
-    func open(identity: String) throws -> Sphere {
+    func sphere(identity: String) throws -> Sphere {
         let noosphere = try noosphere()
         logger.debug("init Sphere with identity: \(identity)")
-        let sphere = try Sphere(
+        return try Sphere(
             noosphere: noosphere,
             identity: identity
         )
-        self._sphere = sphere
-        return sphere
     }
     
     func identity() throws -> String {
@@ -282,15 +280,14 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
         }
     }
     
+    /// Attempt to retrieve the sphere of a recorded petname, this can be chained to walk
+    /// over multiple spheres:
+    ///
+    /// `sphere().traverse(petname: "alice")?.traverse(petname: "bob")?.traverse(petname: "alice)` etc.
+    ///
     func traverse(petname: String) throws -> Sphere? {
         try queue.sync {
-            try self.sphere().traverseByPetname(petname: petname)
-        }
-    }
-    
-    func traverse(sphere: Sphere, petname: String) throws -> Sphere? {
-        try queue.sync {
-            try sphere.traverseByPetname(petname: petname)
+            try self.sphere().traverse(petname: petname)
         }
     }
 }
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -143,10 +143,12 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
         
         let noosphere = try noosphere()
         logger.debug("init Sphere with identity: \(identity)")
-        return try Sphere(
+        let sphere = try Sphere(
             noosphere: noosphere,
             identity: identity
         )
+        self._sphere = sphere
+        return sphere
     }
     
     func identity() throws -> String {
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -141,11 +141,6 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
             throw NoosphereServiceError.defaultSphereNotFound
         }
         
-        return try sphere(identity: identity)
-    }
-    
-    /// Open a sphere using its identity
-    func sphere(identity: String) throws -> Sphere {
         let noosphere = try noosphere()
         logger.debug("init Sphere with identity: \(identity)")
         return try Sphere(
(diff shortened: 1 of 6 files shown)


xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -140,6 +140,12 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
         guard let identity = self._sphereIdentity else {
             throw NoosphereServiceError.defaultSphereNotFound
         }
+        
+        return try open(identity: identity)
+    }
+    
+    /// Open a sphere using its identity
+    func open(identity: String) throws -> Sphere {
```
bfollington, line 148: Unsure if this actually makes sense as an API. I want to be able to manage multiple `Sphere`s so this seems like a logical inclusion.

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -348,6 +354,26 @@ public final class Sphere: SphereProtocol {
...
+            sphere,
+            petname
+        )
+        
+        guard let sphere = sphere else {
+            logger.warning("Failed to traverse sphere for \(petname)")
+            return nil
+        }
+        
+        return Sphere(noosphere: noosphere, sphere: sphere, identity: identity)
```
bfollington, line 374: Given that we get a `ns_sphere_t` back from this call we may as well use it to initialize our own `Sphere` and skip another call.

xcode/Subconscious/SubconsciousTests/TestUtilities.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+//  Created by Ben Follington on 3/4/2023.
+//
+
+import XCTest
+import Combine
+import ObservableStore
+@testable import Subconscious
+
+struct TestUtilities {
```
bfollington, line 13: Extract common test utilities

xcode/Subconscious/SubconsciousTests/Tests_NoosphereService.swift
```diff
@@ -21,7 +21,34 @@ final class Tests_NoosphereService: XCTestCase {
...
+        
+        try bobSphere.setPetname(did: aliceReceipt.identity, petname: "alice")
+        try aliceSphere.setPetname(did: bobReceipt.identity, petname: "bob")
+        try noosphere.save()
+        
+        let aliceDid = try bobSphere.getPetname(petname: "alice")
+        XCTAssertEqual(aliceDid, aliceReceipt.identity)
+        
+//        let result = try noosphere.sync() 
+        let newSphere = try noosphere.traverse(sphere: bobSphere, petname: "alice")
```
bfollington, line 44: Am I right in thinking I need to actually `sync` before this will work?

jsantell: Bob's sphere records Alice's sphere's identity in Bob's address book, but a gateway is needed to resolve the identity into a Link (CID) -- some ideas:

* Compartmentalize some tests that require a running gateway (integration tests), and run a gateway locally as part of the test (is `orb` a part of the build step anywhere?)
* Expose a way (possibly in FFI) to manually set the Link of a sphere, though the hydration would still require IPFS (or seeding into a store via side-channel)

We have a `test_kubo` feature in Noosphere that require running ipfs-desktop in conjunction for tests that require full IPFS node running when testing e.g. the gateway/name system. I suspect there'll be some heavier workflows worth testing in this way

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -75,6 +75,12 @@ public final class Sphere: SphereProtocol {
     public let sphere: OpaquePointer
     public let identity: String
     
+    init(noosphere: Noosphere, sphere: OpaquePointer, identity: String) {
```
bfollington, line 78: Should be private?

xcode/Subconscious/SubconsciousTests/Tests_NoosphereService.swift
```diff
@@ -21,7 +21,45 @@ final class Tests_NoosphereService: XCTestCase {
         )
         return url
     }
+    
+    func testSphereTraversal() throws {
+        throw XCTSkip("Sphere Traversal relies on running a gateway, this test will not pass currently.")
```
bfollington, line 26: Test is skipped for now, but this demonstrates the intended usage for `traverse`

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -348,6 +361,25 @@ public final class Sphere: SphereProtocol {
...
+        let sphere = try Noosphere.callWithError(
+            ns_sphere_traverse_by_petname,
+            noosphere.noosphere,
+            sphere,
+            petname
+        )
+        
+        guard let sphere = sphere else {
+            logger.warning("Failed to traverse sphere for \(petname)")
+            return nil
```
gordonbrander, line 377: Should we throw a custom error rather than returning nil? This would make the signature

```swift
func traverse(petname: String) throws -> Sphere
```

I find that it is often more ergonomic to use one or the other style of fallibility, rather than both.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -140,14 +140,18 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
         guard let identity = self._sphereIdentity else {
             throw NoosphereServiceError.defaultSphereNotFound
         }
+        
+        return try sphere(identity: identity)
+    }
+    
+    /// Open a sphere using its identity
+    func sphere(identity: String) throws -> Sphere {
```
gordonbrander, line 148: Why do we need this method when we have traverse?

If we do need this method, I feel we should consider making it a synonym of `traverse`, or pursue one of the other directions below.

```swift
func traverse(identity: String) throws -> Sphere
```

Rationale: `.sphere()` is meant to return a fallible memoized instance of the _user's_ sphere — it's essentially like a property, but allowing for throwing failure.

Adding a synonym to return someone else's sphere by identity confuses the concept.

Approaches to consider:

- Synonym for `traverse`
- Move to `Noosphere` as `Noosphere.sphere(identity:)`
    - Noosphere is lower-level, not a service, so doesn't have the concept of "user's sphere" to worry about. Probably a better place for this.

gordonbrander (changes requested):
Minor API changes requested

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
bfollington: Cool, I was wondering about this. I've switched it to throw on a failed traversal instead.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
bfollington: We don't actually need it at all, I was just thinking forward to code that we don't need to write yet. 

In the event we truly require a method like this one I quite like `Noosphere.sphere(identity:)` as a location.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -140,14 +140,13 @@ final class NoosphereService: SphereProtocol, SphereIdentityProtocol {
...
         }
+        
         let noosphere = try noosphere()
         logger.debug("init Sphere with identity: \(identity)")
-        let sphere = try Sphere(
+        return try Sphere(
             noosphere: noosphere,
             identity: identity
         )
-        self._sphere = sphere
```
gordonbrander, line 149: I think this removal was unintended? We need to cache the user's sphere to `self._sphere` here.

gordonbrander (changes requested):

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
bfollington: Good catch, restored.

gordonbrander (approved):

PR #486, opened 2023-04-05
Implement basic 3p sphere content view

### LINKED ISSUE #451: Implement viewer detail for 3p sphere content
We currently support parsing and dispatching actions for 3P `@handle/note` slashlinks, but the 3P content viewer is not yet implemented, and is config-flagged off.

Now that the FFI has landed for 3P sphere content https://github.com/subconsciousnetwork/noosphere/pull/271, we should implement this.

Followup to #442 and #422. Part of #330.

gordonbrander:
Progress toward #451.

This PR lands a basic 3P content viewer we can use to traverse the Noosphere.

I ran into some minor problems with `Text` and `AttributedString` not rendering links correctly. Rather than debug, I stubbed in a read-only version of `SubtextTextViewRepresentable`. Additionally, the view has basic handling for loading and not found states, and a very basic meta sheet.

Further refinement is needed before flagging on, but I would like to land this progress behind a flag. That way we can flip the flag to allow for traversal in dev.


gordonbrander:
Attempted to render Subtext NSMutableAttributedString to AttributedString, and pass to `Text()`. It mostly works. However, links don't seem to work correctly.

Getting to the bottom of this will require some investigation. Going to use a read-only SubtextTextViewRepresentable for now.

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -43,6 +43,7 @@ enum SubtextTextAction: Hashable, CustomLogStringConvertible {
     case setSelection(range: NSRange, text: String)
     /// Set selection at the end of the text
     case setSelectionAtEnd
+    case setEditable(Bool)
```
gordonbrander, line 46: Expose UITextView editable setting on component

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetailMetaSheetView.swift
```diff
@@ -8,10 +8,10 @@
 import SwiftUI
 import ObservableStore
 
-struct DetailMetaSheet: View {
+struct MemoEditorDetailMetaSheetView: View {
```
gordonbrander, line 11: Change name of component to differentiate between e.g. viewer meta sheets vs editor meta sheets, which have different features.

xcode/Subconscious/Shared/Models/LoadingState.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  NetworkResourceState.swift
+//  Subconscious (iOS)
+//
+//  Created by Gordon Brander on 4/6/23.
+//
+
+import Foundation
+
+enum LoadingState: Hashable {
```
gordonbrander, line 10: General enum for representing loading states. Useful in cases where you want a flag on a struct.

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
```diff
@@ -0,0 +1,144 @@
...
+                            }
+                        ) {
+                            Label(
+                                "Copy link",
+                                systemImage: "doc.on.doc"
+                            )
+                        }
+                        .disabled(state.address == nil)
+                        .buttonStyle(RowButtonStyle())
+                    }
```
bfollington, line 55: To me, it seems cleaner to use `ShareLink` here instead of copying to the clipboard. I think it's more idiomatic in iOS in general to go via the share sheet since it offers copy to pasteboard within it anyway & it's more directly useful for sending a link in an iMessage etc. 

This apple link looked promising for guidance but I can't load it right now because of an SSL error(!!): https://developer.apple.com/design/human-interface-guidelines/patterns/collaboration-and-sharing/

We could expose a computed `shareableLink` property from the model to keep the logic from the `.copyLink` handler in the model, then use the `MetaTableItemShareLinkView` I implemented recently like so:

```swift
MetaTableItemShareLinkView(label: "Share Link", item: state.shareableLink)
```

Happy to merge this PR as-is, but I would vote to prefer `ShareLink` over manual pasteboard interaction as the default choice.

xcode/Subconscious/Shared/Models/LoadingState.swift
bfollington: I expect we'll expand this to include other failure cases in the future? No reason to add them until they come up though.

bfollington (approved):
LGTM, I'm going to merge this then drop a follow-up PR to change the pasteboard use to `ShareLink` & we can discuss whether that's worth it as part of that change.

PR #493, opened 2023-04-07
Read initial UserDefaultsProperty value from UserDefaults

### LINKED ISSUE #487: "Enable Noosphere" setting spontaneously turns itself off at app startup
# Reproduction Steps

- Open the app
- Enable Noosphere in Developer Settings
- Quit the app
- Open the app
- Note that Noosphere is disabled again

# Investigation

Seems to be due to `.notifyNoosphereEnabled` being dispatched on app start with a value of `false` despite the `AppStore` being initialised with `true` (as read from `AppDefaults).

gordonbrander:
Fixes #487.

Previously, we were setting the initial value to the wrapped `default` value. What we should have been doing is reading from UserDefaults at construction, and passing that to our `CurrentValueSubject`. This PR fixes the problem.

This is what was causing our Noosphere flag to spontaneously turn itself off between runs.

### DIFF
--- a/xcode/Subconscious/Shared/Library/UserDefaultsProperty.swift
+++ b/xcode/Subconscious/Shared/Library/UserDefaultsProperty.swift
@@ -7,22 +7,36 @@
 import Combine
 import Foundation
 
-/// Creates a type-safe getter and setter for a UserDefaults key.
-/// Properties decorated with `@UserDefaultsProperty` will get and set their
-/// value from UserDefaults.
+/// Property wrapper that creates a type-safe getter/setter for a
+/// UserDefaults key.
 ///
-/// The property's default value will be used if no value is found in the
-/// UserDefaults for that key.
+/// Properties decorated with `@UserDefaultsProperty` will get and set their
+/// value from UserDefaults. The property's default value will be used if
+/// no value is found in the UserDefaults for that key.
 ///
 /// You can subscribe to changes in this default via the projected value, using
 /// `$name`, where name is the name of your property. Projected value is a
 /// publisher that emits any time the wrapped value changes.
 @propertyWrapper struct UserDefaultsProperty<Value: Equatable> {
+    /// Create a type-safe accessor for a UserDefaults key.
+    private struct UserDefaultsKey<Value: Equatable> {
+        var key: String
+        var `default`: Value
+        var scope = UserDefaults.standard
+        
+        var value: Value {
+            get {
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Library/UserDefaultsProperty.swift
```diff
@@ -7,22 +7,36 @@
...
+/// Properties decorated with `@UserDefaultsProperty` will get and set their
+/// value from UserDefaults. The property's default value will be used if
+/// no value is found in the UserDefaults for that key.
 ///
 /// You can subscribe to changes in this default via the projected value, using
 /// `$name`, where name is the name of your property. Projected value is a
 /// publisher that emits any time the wrapped value changes.
 @propertyWrapper struct UserDefaultsProperty<Value: Equatable> {
+    /// Create a type-safe accessor for a UserDefaults key.
+    private struct UserDefaultsKey<Value: Equatable> {
```
gordonbrander, line 22: Added this inner accessor type to offer a typesafe wrapper we could use in both the constructor and the consumer-facing property wrapper accessor.

xcode/Subconscious/Shared/Library/UserDefaultsProperty.swift
```diff
@@ -31,8 +45,12 @@ import Foundation
...
     ) {
         self.key = key
-        self.default = wrappedValue
-        self.subject = CurrentValueSubject(wrappedValue)
+        self.store = UserDefaultsKey(
+            key: key,
+            default: wrappedValue,
+            scope: scope
+        )
+        self.subject = CurrentValueSubject(store.value)
```
gordonbrander, line 53: Set initial value from backing store (UserDefaults)

xcode/Subconscious/SubconsciousTests/Tests_UserDefaultProperty.swift
```diff
@@ -46,6 +46,6 @@ final class Tests_UserDefaultProperty: XCTestCase {
         test.isToggled = false
         test.isToggled = false
         test.isToggled = false
-        XCTAssertEqual(count, 2)
+        XCTAssertEqual(count, 1)
```
gordonbrander, line 49: Turns out the reason we were getting 2 counts is because the value wasn't being read from the backing store on construction. Test updated. 😨

bfollington (approved):
LGTM!

PR #494, opened 2023-04-10
Update to Noosphere v0.8.2

gordonbrander:
- Lock in new commit
- Update `Sphere` wrapper class to defer to Noosphere traversal logic instead of attempting to implement on Swift side.

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
@@ -194,16 +194,16 @@ public final class Sphere: SphereProtocol {
         return Self.readFileHeaderNames(file: file)
     }
     
-    /// Read the value of a memo from this Sphere
+    /// Read the value of a memo from this, or another sphere
     /// - Returns: `MemoData`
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -194,16 +194,16 @@ public final class Sphere: SphereProtocol {
         return Self.readFileHeaderNames(file: file)
     }
     
-    /// Read the value of a memo from this Sphere
+    /// Read the value of a memo from this, or another sphere
     /// - Returns: `MemoData`
-    private func read(slug: Slug) throws -> MemoData {
+    public func read(slashlink: Slashlink) throws -> MemoData {
         guard let file = try Noosphere.callWithError(
```
cdata, line 200: Sort of an aside, but I wonder if it is possible in Swift to curry this so that you can invoke the FFI more idiomatically like a straight-up function call (but also benefit from the semantics of `callWithError`). (no action required)

cdata (approved):
Awesome!

PR #496, opened 2023-04-11
Use ShareLink for sharing links

bfollington:
This PR picks up on the discussion I started in https://github.com/subconsciousnetwork/subconscious/pull/486#discussion_r1162195697.

> To me, it seems cleaner to use ShareLink here instead of copying to the clipboard. I think it's more idiomatic in iOS in general to go via the share sheet since it offers copy to pasteboard within it anyway & it's more directly useful for sending a link in an iMessage etc.

> This apple link looked promising for guidance but I can't load it right now because of an SSL error(!!): https://developer.apple.com/design/human-interface-guidelines/patterns/collaboration-and-sharing/

> We could expose a computed shareableLink property from the model to keep the logic from the .copyLink handler in the model, then use the MetaTableItemShareLinkView I implemented recently...

This PR shows what it would look like to make this switch.

# Changes

- Convert .copyLink action to ShareLink control
- Remove unused pasteboard environment

### DIFF
--- a/xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
+++ b/xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
@@ -62,7 +62,7 @@ enum MemoViewerDetailMetaSheetAction: Hashable {
 
 struct MemoViewerDetailMetaSheetModel: ModelProtocol {
     typealias Action = MemoViewerDetailMetaSheetAction
-    typealias Environment = PasteboardProtocol
+    typealias Environment = ()
     
     static let logger = Logger(
         subsystem: Config.default.rdns,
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
```diff
@@ -82,6 +74,22 @@ struct MemoViewerDetailMetaSheetModel: ModelProtocol {
     var noteVersion: String?
     var authorKey: String?
     
+    var shareableLink: String? {
+        guard let address = address else {
+            Self.logger.log("Copy link: (nil)")
```
gordonbrander, line 79: We should remove the logging. I had this in the update function since clipboard seemed like a relevant event. However share link does not do the same modification of global persistent state.

gordonbrander (changes requested):
Would remove logging from `shareableLink`.

Otherwise, LGTM

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
bfollington: Done!

gordonbrander (approved):

Issue #462, opened 2023-03-17
Services Architecture


gordonbrander:
We are currently pretty unprincipled about the way services are structured, and communicate with each other. This issue tracks ways we could clean that relationship up.

## Requirements

- Services should be single-purpose
- Services depend upon other services sometimes.
- Services should be mockable.
- Services are stateful. They are a facade over some source of truth.

Swift UI app considerations:

- Services are generally long-lived, usually for the lifetime of the app.
- Services must be able to initialize without failing (no throw, no nil)
    - Since services are long-lived, it is not practical to cope with failures once only at startup.
    - However methods of services may fail in any way they choose.

## A: Services with delegates

Outline:

- Services are single-purpose classes, structs, or actors.
    - Since services will be shared to many places, perhaps require them to be actors to ensure thread safety.
- Dependencies on other services are handled through mutable delegate properties.
    - Delegates can be configured during initialization, or by assigning to property.
- Delegate properties are **always** protocols, never concrete types.
    - This allows for mocking.
    - Very minimal protocols could be defined that only contain the dependencies needed for that service.

Sketch:

```swift
class FooService {
    var count: Int = 0
    var bar: BarServiceProtocol
    var baz: BazServiceProtocol

    func count() -> Int {
        let barCount = bar.count()
        let bazCount = baz.count()
        return count + barCount + bazCount
    }
}
```

This is very slight riff on what we already do. The main difference is using protocols for dependencies to allow for mocking and replacement.

Decent blog post about roughly this approach https://www.cocoawithlove.com/blog/separated-services-layer.html#lets-try-to-add-tests-to-cwlfeedreader

## B: Services with encapsulated dependencies

A riff on **A**. This approach would hide (encapsulate) service dependencies, to enforce the single-purpose-ness of each service.

Outline:

- Services are single-purpose classes, structs, or actors.
    - Since services will be shared to many places, perhaps require them to be actors to ensure thread safety.
- Dependencies on other services are handled through initializer arguments which are assigned to `private` properties of the service.
- Dependencies are **always** protocols, never concrete types.
    - This allows for mocking.
    - Very minimal protocols could be defined that only contain the dependencies needed for that service.

Sketch, similar to above but showing dependency hiding, and actor-forward approach:

```swift
actor FooService {
    var count: Int = 0
    private var bar: BarServiceProtocol
    private var baz: BazServiceProtocol

    init(bar: BarServiceProtocol, baz: BazServiceProtocol) {
[...]

PR #501, opened 2023-04-12
Make all calls to Noosphere non-blocking on main thread

### LINKED ISSUE #439: App becomes non-responsive during sync with invalid gateway
This _may_ be a simulator only issue (I'm currently unable to test this on-device due to #376). 

# Reproduction

1. App setup (create a sphere)
2. Configure your gateway URL with Tailscale running 
3. Sync with gateway successfully (unsure if we need this step)
4. Quit app
5. Kill Tailscale (so gateway is unreachable)
6. Re-open app
7. Tap settings icon
8. Watch the app hang until the the request to the gateway times out (60 seconds for me)
9. Settings sheet finally opens, showing error

If you manually run sync again, close the sheet and try to re-open it the app will hang until the timeout error occurs again.

Screenshot 2023-03-09 at 12 13 04 pm

# Expected Behaviour

Sync would run and fail in the background, letting me use the app in the meantime.

# Notes

My gateway URL is `http://bfollington.sphere.test.subconscious.cloud` and I suspect that 50% of the issue stems from the fact that pinging that URL (without Tailscale) takes 60 seconds to actually timeout. Using a localhost gateway like `http://127.0.0.1:4433` instantly times out instead of waiting.

Of course, the UI should still not be blocking during that waiting period.

Here's a video of the issue:

https://user-images.githubusercontent.com/5009316/223899842-e6b237cb-b563-40d8-ba85-6a8012ad480a.mov

gordonbrander:
Fixes #439 

There are a few places where we have inadvertently called Noosphere synchronously on the main thread, causing the UI to block on certain long operations, like syncing. This should never happen.

This PR fixes the issue, and makes services Swift `actor`s to make it difficult to trip over this footgun. Actors are isolated, and can only be accessed via async/await. This means you cannot accidentally call blocking synchronous APIs on an actor (as I found we were doing in a number of subtle places). You must await, or else wrap in some kind of callback or publisher. This prevents us from accidentally causing UI hangs.

## Notes

- `NoosphereService`, `Noosphere` and `Sphere` are now Swift `actor`s. That means they always run off of the main thread, and are isolated (can only be accessed from the outside via `async`/`await` or via thread-safe things like publishers).
    - Swift manages a cooperative threadpool of non-main-thread threads for all actors. The thread pool is concurrent.
- A new `SpherePublisherProtocol` exposes nonisolated publishers for Noosphere actions.
    - `NoosphereService` conforms to `SpherePublisherProtocol`
- Services have been "flattened". App Environment holds a bag of services, each of which do one job.
    - Services may depend upon each other by passing in references during construction
    - Services keep their dependencies private. This means there is no calling `data.noosphere` anymore. If you need `noosphere`, you should just call `noosphere`.
    - Services are expected to handle their own thread-safety/atomicity if needed
- This gets us most of the way to the design discussed in #462 
    - I tried exposing only protocols for services, so that components could specify only the requirements they needed. However, this had a lot of sharp edges. Swift's type system is just not up to the task of generics with protocol constraints. I ended up going back to environments made up of concrete types.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
@@ -18,9 +18,14 @@ struct AddressBookEntry: Equatable {
 }
 
 struct AddressBookEnvironment {
+    var logger = Logger(
+        subsystem: Config.default.rdns,
+        category: "AddressBook"
+    )
+    var noosphere: NoosphereService
     var data: DataService
+    var addressBook: AddressBookService
 }
-
 
 struct FollowUserFormModel: Equatable {
     var did: FormField<String, Did> = FormField(value: "", validate: Self.validateDid)
@@ -36,39 +41,49 @@ struct FollowUserFormModel: Equatable {
 }
 
 struct PetnameFieldCursor: CursorProtocol {
-    static func get(state: AddressBookModel) -> FormField<String, Petname> {
+    typealias Model = AddressBookModel
+    typealias ViewModel = FormField<String, Petname>
+
+    static func get(state: Model) -> ViewModel {
         state.followUserForm.petname
     }
     
-    static func set(state: AddressBookModel, inner: FormField<String, Petname>) -> AddressBookModel {
+    static func set(state: Model, inner: ViewModel) -> Model {
         var model = state
         model.followUserForm.petname = inner
         return model
     }
     
-    static func tag(_ action: FormFieldAction<String>) -> AddressBookAction {
+    static func tag(_ action: ViewModel.Action) -> Model.Action {
         AddressBookAction.petnameField(action)
     }
 }
 
 struct DidFieldCursor: CursorProtocol {
-    static func get(state: AddressBookModel) -> FormField<String, Did> {
+    typealias Model = AddressBookModel
+    typealias ViewModel = FormField<String, Did>
+
+    static func get(state: Model) -> ViewModel {
         state.followUserForm.did
     }
     
-    static func set(state: AddressBookModel, inner: FormField<String, Did>) -> AddressBookModel {
+    static func set(state: Model, inner: ViewModel) -> Model {
         var model = state
         model.followUserForm.did = inner
         return model
     }
     
-    static func tag(_ action: FormFieldAction<String>) -> AddressBookAction {
+    static func tag(_ action: ViewModel.Action) -> Model.Action {
         AddressBookAction.didField(action)
     }
 }
 
 enum AddressBookAction {
     case present(_ isPresented: Bool)
+    
+    case refreshDid
+    case succeedRefreshDid(Did)
+    case failRefreshDid(_ error: String)
     
     case refreshEntries(forceRefetchFromNoosphere: Bool = false)
     case populate([AddressBookEntry])
@@ -110,12 +125,7 @@ struct AddressBookModel: ModelProtocol {
     var unfollowCandidate: Petname? = nil
     
     var isQrCodeScannerPresented = false
-
-    static let logger = Logger(
-        subsystem: Config.default.rdns,
-        category: "AddressBookModel"
-    )
-
+    
     static func update(
         state: AddressBookModel,
         action: AddressBookAction,
@@ -124,28 +134,37 @@ struct AddressBookModel: ModelProtocol {
         switch action {
             
         case .present(let isPresented):
-            var model = state
-            model.isPresented = isPresented
-            model.failUnfollowErrorMessage = nil
-            model.failFollowErrorMessage = nil
-            
-            if isPresented {
-                do {
-                    model.did = try Did(environment.data.noosphere.identity())
-                } catch {
-                    model.did = nil
-                }
-            }
-            
-            return update(
-                state: model,
-                action: .refreshEntries(forceRefetchFromNoosphere: false),
-                environment: environment
+            return present(
+                state: state,
+                environment: environment,
+                isPresented: isPresented
             )
+            
+        case .refreshDid:
+            let fx: Fx<AddressBookAction> = environment.noosphere
+                .identityPublisher()
+                .tryMap({ identity in
+                    let did = try Did(identity).unwrap()
+                    return .succeedRefreshDid(did)
+                })
+                .recover({error in
+                        .failRefreshDid(error.localizedDescription)
+                })
+                .eraseToAnyPublisher()
+            return Update(state: state, fx: fx)
+            
+        case .succeedRefreshDid(let did):
+            var model = state
+            model.did = did
+            return Update(state: model)
+            
+        case .failRefreshDid(let error):
+            environment.logger.log("Failed to refresh sphere did: \(error)")
+            return Update(state: state)
             
         case .refreshEntries(let forceRefreshFromNoosphere):
             let fx: Fx<AddressBookAction> =
-                environment.data.addressBook.listEntries(refetch: forceRefreshFromNoosphere)
+            environment.addressBook.listEntries(refetch: forceRefreshFromNoosphere)
                 .map({ follows in
                     AddressBookAction.populate(follows)
                 })
@@ -198,7 +217,7 @@ struct AddressBookModel: ModelProtocol {
                 ],
                 environment: environment
             )
-
+            
             
         case .attemptFollow:
             guard let did = state.followUserForm.did.validated else {
@@ -217,7 +236,7 @@ struct AddressBookModel: ModelProtocol {
             }
             
             let fx: Fx<AddressBookAction> =
-                environment.data.addressBook
+            environment.addressBook
                 .followUserAsync(did: did, petname: petname)
                 .map({ _ in
                     AddressBookAction.succeedFollow(did: did, petname: petname)
@@ -270,7 +289,7 @@ struct AddressBookModel: ModelProtocol {
             }
             
             let fx: Fx<AddressBookAction> =
-                environment.data.addressBook
+            environment.addressBook
                 .unfollowUserAsync(petname: petname)
                 .map({ _ in
                     AddressBookAction.succeedUnfollow(petname: petname)
@@ -304,13 +323,13 @@ struct AddressBookModel: ModelProtocol {
             var model = state
             model.failUnfollowErrorMessage = nil
             return Update(state: model)
-        
+            
         case .presentQRCodeScanner(let isPresented):
             var model = state
             model.failQRCodeScanErrorMessage = nil
             model.isQrCodeScannerPresented = isPresented
             return Update(state: model)
-        
+            
         case .qrCodeScanned(scannedContent: let content):
             return update(
                 state: state,
@@ -328,5 +347,33 @@ struct AddressBookModel: ModelProtocol {
             
         }
         
+    }
+    
+    static func present(
+        state: Self,
+        environment: Self.Environment,
+        isPresented: Bool
+    ) -> Update<Self> {
+        var model = state
+        model.isPresented = isPresented
+        model.failUnfollowErrorMessage = nil
+        model.failFollowErrorMessage = nil
+        
+        guard isPresented else {
+            return update(
+                state: model,
+                action: .refreshEntries(forceRefetchFromNoosphere: false),
+                environment: environment
+            )
+        }
+
+        return update(
+            state: model,
+            actions: [
+                .refreshDid,
+                .refreshEntries(forceRefetchFromNoosphere: false)
+            ],
+            environment: environment
+        )
     }
 }
--- a/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
+++ b/xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
@@ -18,10 +18,12 @@ struct AddressBookEntry: Equatable {
 }
 
 struct AddressBookEnvironment {
+    // Default logger for environment
     static let logger = Logger(
         subsystem: Config.default.rdns,
         category: "AddressBook"
     )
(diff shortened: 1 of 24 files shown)


xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -64,8 +63,66 @@ public protocol SphereProtocol {
     func traverse(petname: Petname) throws -> Sphere
 }
 
-protocol SphereIdentityProtocol {
```
gordonbrander, line 67: No longer used anywhere, so I removed it. Would prefer to make this part of SphereProtocol anyway.

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -64,8 +63,66 @@ public protocol SphereProtocol {
...
-    func identity() throws -> String
+/// Describes a Sphere using a Combine Publisher-based API.
+/// See `Sphere` for concrete implementation.
+public protocol SpherePublisherProtocol {
```
gordonbrander, line 68: An asynchronous publisher-based protocol that mirrors the synchronous one.

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -97,6 +154,10 @@ public final class Sphere: SphereProtocol {
     public let sphere: OpaquePointer
     public let identity: String
     
+    private var queue: DispatchQueue {
```
gordonbrander, line 157: Use same queue as the Noosphere instance

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1171,6 +1205,11 @@ struct AppEnvironment {
     var documentURL: URL
     var applicationSupportURL: URL
 
+    private var _noosphere: NoosphereService
+    var noosphere: some NoosphereServiceProtocol { _noosphere }
+    /// Exposes sphere publisher protocol for Noosphere
+    var sphere: some SpherePublisherProtocol { _noosphere }
```
gordonbrander, line 1211: We use `some` to expose only certain protocols from the caller side.

`some` does caller-side type-erasure, so that the type remains concrete in `AppEnvironement`, but the caller may only interact with the protocol, and not the underlying type.

The protocols we expose are "UI-thread-friendly". That is, if they are blocking, they are very fast, and if they aren't very fast, they produce Combine publishers and run on a separate thread.

xcode/Subconscious/Shared/Library/CombineUtilities.swift
```diff
@@ -41,3 +41,74 @@ struct CombineUtilities {
         .eraseToAnyPublisher()
     }
 }
+
+extension DispatchQueue {
+    /// Run a closure async on this queue, returning a Combine Future.
+    func future<T>(
```
gordonbrander, line 47: Create a Combine future from an async work item on a GDC queue. This helper is similar in principle to the `CombineUtilities.async` helper, but gives us more control over which GDC queue things are run on.

In future, I would like to move all of our uses of `CombineUtilities.async` to `DispatchQueue.future`.

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -85,9 +85,7 @@ struct DataService {
 
     /// Sync local state to gateway
     func syncSphereWithGateway() -> AnyPublisher<String, Error> {
-        CombineUtilities.async(qos: .utility) {
-            try noosphere.sync()
-        }
+        noosphere.syncPublisher()
```
gordonbrander, line 88: Noosphere handles its own async publisher stuff now.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -27,25 +28,37 @@ enum NoosphereServiceError: Error, LocalizedError {
     }
 }
 
+protocol NoosphereServiceProtocol {
```
gordonbrander, line 31: Exposes basic controls for interacting with memoized Noosphere/Sphere instance.

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -36,40 +41,50 @@ struct FollowUserFormModel: Equatable {
 }
 
 struct PetnameFieldCursor: CursorProtocol {
-    static func get(state: AddressBookModel) -> FormField<String, Petname> {
+    typealias Model = AddressBookModel
```
gordonbrander, line 44: These changes were byproducts of experimenting with type-erased generics. I ended up going back to concrete types for services, but the typealiases may as well stay. It's cleaner.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -391,7 +396,7 @@ struct AppModel: ModelProtocol {
             return AppAddressBookCursor.update(
                 state: state,
                 action: action,
-                environment: environment.addressBook
+                environment: AddressBookEnvironment(environment)
```
gordonbrander, line 399: Constructing AddressBookEnvironment on the fly instead.

The way I've got this set up at the moment is environments hold services. Services are classes (or maybe actors in future), but environments are just structs to hold these services. This makes them cheap to create and throw away, provided the services are just passed down by reference. I've created a convenience initializer for `AddressBookEnvironment` that takes an `AppEnvironment`.

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -38,10 +38,10 @@ struct MoveReceipt: Hashable {
...
 /// access methods for the app.
 struct DataService {
-    var addressBook: AddressBookService
-    var noosphere: NoosphereService
-    var database: DatabaseService
-    var local: HeaderSubtextMemoStore
+    private var addressBook: AddressBookService
+    private var noosphere: NoosphereService
+    private var database: DatabaseService
+    private var local: HeaderSubtextMemoStore
```
gordonbrander, line 44: Dependencies of NoosphereService should be private.

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -14,6 +15,8 @@ import os
 public protocol SphereProtocol {
     associatedtype Memo
     
+    func identity() throws -> String
```
gordonbrander, line 18: The protocol allows for throwing identity so that we can implement this signature in `NoosphereService`, where it might fail.

gordonbrander:
Unfortunately I discovered another UI blocker (back button save) and am tracking down source. This PR is not ready yet.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -43,115 +43,115 @@ class AddressBookService {
     /// Get the full list of entries in the address book.
     /// This is cached after the first use unless requested using refetch.
     /// If an error occurs producing the entries the resulting list will be empty.
-    func listEntries(refetch: Bool = false) -> AnyPublisher<[AddressBookEntry], Never> {
-        if let entries = addressBook {
+    func listEntries(
```
gordonbrander, line 46: Refactor this logic using async/await semantics. It reads more clearly.

gordonbrander:
Swift async uses a small thread pool and expects IO to be async so that fn can yield to other work while io is happening.

Until we get Noosphere's non-blocked reads and callback-based non blocking api, we might want to shove Noosphere onto the global queue even though it's an actor already.

https://developer.apple.com/forums/thread/708664

xcode/Subconscious/Shared/Library/CombineUtilities.swift
```diff
@@ -8,36 +8,87 @@
...
+            }
+        }
+    }
+}
+
+extension Publisher {
+    /// Recover from a failure.
+    /// Similar to `catch` but allows you to map an `Error` to an `Output`,
+    /// without having to wrap in a publisher.
+    func recover(
```
bfollington, line 87: 💯

xcode/Subconscious/Shared/Components/AddressBook/AddressBookView.swift
```diff
@@ -58,7 +58,6 @@ struct AddressBookView: View {
             .toolbar {
                 ToolbarItem(placement: .cancellationAction) {
                     Button("Close", role: .cancel) {
-                        AddressBookModel.logger.debug("Close Address Book")
```
gordonbrander, line 61: Not needed / if needed should be in update function.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -988,16 +1080,10 @@ struct AppModel: ModelProtocol {
...
-            logger.log("No gateway configured. Skipping sync.")
-            return Update(state: state)
-        }
-        
         var model = state
         model.lastGatewaySyncStatus = .pending
-        
-        logger.log("Syncing with gateway: \(gatewayURL.absoluteString)")
-        let fx: Fx<AppAction> = environment.data.syncSphereWithGateway()
+        logger.log("Syncing with gateway: \(model.gatewayURL)")
```
gordonbrander, line 1085: We no longer have synchronous access to this property on the actor, so we use the copy on the model instead (the update fn keeps it up-to-date)

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1221,30 +1310,43 @@ struct AppEnvironment {
             gatewayURL: defaultGateway,
             sphereIdentity: defaultSphereIdentity
         )
+        self.noosphere = noosphere
```
gordonbrander, line 1313: Keep Noosphere around. Now that services are "flattened" we need to make sure we have one property for each service.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1221,30 +1310,43 @@ struct AppEnvironment {
...
 
-        let databaseService = DatabaseService(
+        let database = DatabaseService(
             database: SQLite3Database(
                 path: databaseURL.absoluteString,
                 mode: .readwrite
             ),
             migrations: Config.migrations
         )
+        self.database = database
```
gordonbrander, line 1325: Keep database around on service.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1221,30 +1310,43 @@ struct AppEnvironment {
...
         
-        let addresBook = AddressBookService(noosphere: noosphere, database: databaseService)
+        let addressBook = AddressBookService(
+            noosphere: noosphere,
+            database: database
+        )
+        self.addressBook = addressBook
```
gordonbrander, line 1331: Keep address book around on service.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1221,30 +1310,43 @@ struct AppEnvironment {
...
+            addressBook: addressBook
         )
         
-        self.addressBook = AddressBookEnvironment(data: data)
-
         self.feed = FeedService()
     }
 }
 
+extension AddressBookEnvironment {
```
gordonbrander, line 1344: Extension to easily construct AddressBookEnvironment from AppEnvironment. Note relationship of extension = AppEnvironment knows about AddressBookEnvironment, but AddressBookEnvironment doesn't know about AppEnvironment. This mirrors component relationship where parents may know about children, but children may not know about parents.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -939,18 +939,18 @@ struct NotebookModel: ModelProtocol {
...
-        let address = environment.data
-            .findAddressInOurs(slug: slug) ?? fallbackAddress
-        return update(
-            state: state,
-            action: .pushDetail(
-                MemoEditorDetailDescription(
-                    address: address,
-                    fallback: fallback
+        let fx: Fx<NotebookAction> = environment.data
+            .findAddressInOursPublisher(slug: slug)
```
gordonbrander, line 943: Fix sync call.

xcode/Subconscious/Shared/Library/CombineUtilities.swift
```diff
@@ -8,36 +8,87 @@
...
-        Future({ promise in
-            DispatchQueue.global(qos: qos).async(execute: {
-                let result: T = execute()
-                promise(.success(result))
-            })
-        })
-        .eraseToAnyPublisher()
+/// Create a Combine Future from an async closure that never fails.
+/// Async actions are run in a task and fulfil the future's promise.
+public extension Future where Failure == Never {
```
gordonbrander, line 13: New convenience extensions to `Future` allow us to easily create Combine Futures from async functions.

This gives us much more flexibility in constructing publishers from async functions where and when we need them. It also takes advantage of Swift's new concurrency primitives (Tasks, Actors, async/await) for more ergonomic and efficient concurrency.

This replaces the older `CombineUtilities.async` helper, which scheduled things on the global queue.

xcode/Subconscious/Shared/Components/AddressBook/AddressBookModel.swift
```diff
@@ -18,10 +18,17 @@ struct AddressBookEntry: Equatable {
 }
 
 struct AddressBookEnvironment {
+    // Default logger for environment
+    static let logger = Logger(
+        subsystem: Config.default.rdns,
+        category: "AddressBook"
+    )
+    var logger: Logger = logger
```
bfollington, line 26: Nice!

bfollington (approved):
LGTM!

PR #505, opened 2023-04-17
Read traversal identity using `ns_sphere_identity` FFI method

bfollington:
- Add DispatchQueue.future extension
- Add Tests_CombineUtilities
- Introduce SpherePublisherProtocol
- SpherePublisherProtocol returns AnyPublisher
- Progress toward SpherePublisherProtocol conformance
- More progress toward conformance
- More progress
- Complete conformance
- Wire up non-blocking version refresh
- Comments and whitespace
- Update name to Future.resolve
- Save NoosphereService on AppEnvironment
- Introduce identity() and identityPublisher()
- Go back to concrete types
- Move everything to DispatchQueue.future
- Remove Future.resolve
- Convert services to actors
- Fix a number of tests
- Remove DispatcheQueue.future
- Update more tests
- Fix yet more tests
- Fix remaining tests
- Place logger on environment instance
- Add test for Func.run async
- Make logger private
- Make static logger private
- Read identity from underlying sphere after traverse


xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -178,6 +180,20 @@ public actor Sphere: SphereProtocol, SpherePublisherProtocol {
         logger.debug("init with identity: \(identity)")
     }
     
+    private static nonisolated func fetchIdentityFromSphere(noosphere: OpaquePointer, sphere: OpaquePointer) throws -> String {
```
gordonbrander, line 183: Nit: wrap arguments https://google.github.io/swift/#function-declarations

gordonbrander, line 183: Initializing with a blocking call (`nonisolated` makes the function synchronous and blocking in the calling context) worries me after fixing other similar synchronous calls in #501. These seemingly innocuous calls caused UI freezes due to unexpected latencies caused by sync blocking.

Async initializers are not allowed in swift AFAIK. However, what if we instead created an async static factory function?

```swift
actor Sphere {
    // ...
    static func from(noosphere: Noosphere, sphere: OpaquePointer) async throws -> Self {
        // ...etc
        return self.init(noosphere: Noosphere, identity: identity)
    }
}
```

and at call site:

```swift
let sphere = await Sphere.from(noosphere: noosphere, sphere: pointer)
```

This prevents main thread from calling the code in a synchronous manner.

gordonbrander (changes requested):

gordonbrander (approved):
LGTM

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -178,6 +179,39 @@ public actor Sphere: SphereProtocol, SpherePublisherProtocol {
         logger.debug("init with identity: \(identity)")
     }
     
+    private static func from(
+        noosphere: Noosphere,
+        sphere: OpaquePointer
+    ) async throws -> Self {
+        let identity =
+            try await Self.fetchIdentityFromSphere(
```
gordonbrander, line 187: TIL you can await a future.

PR #506, opened 2023-04-18
Fix typo `detatched` -> `detached`

bfollington:
Tiny change, I kept getting compile errors using `Future.detached` because we had a typo!


gordonbrander (approved):
Whoops! Thanks!

PR #508, opened 2023-04-18
Improvements to 3P content view

### LINKED ISSUE #507: Improved 3P sphere content view
The current 3P sphere content view implemented in #451 is quite basic, and looks the same as an editor view.

We should update this view to include design details from Figma mock and look distinct from the editor.

gordonbrander:
Fixes #507 

This PR is pretty minimal. It replaces the read-only editor view with a new `SubtextView`. This SubtextView is based on ordinary `Text` and `AttributedString` instances. In future it will allow us to more easily implement transclusion and other rich features in read-only mode. It will also be useful for the feed and other story views.

I'd like to keep the PR small, so I am requesting to land this early version that minimally swaps out the editor for SubtextView, and does not do much else. It at least provides some visual differentiation vis editor by making the text `.body` style.


xcode/Subconscious/Shared/Parsers/SubtextAttributedStringRenderer.swift
```diff
@@ -113,17 +141,127 @@ struct SubtextAttributedStringRenderer {
     /// You can use this to produce strings which can be rendered in
     /// `Text` blocks.
     func render(_ string: String) -> AttributedString {
-        let nsAttributedString = NSMutableAttributedString(
-            string: string
-        )
-        _ = renderAttributesOf(nsAttributedString)
-        return AttributedString(nsAttributedString)
+        let dom = Subtext(markup: string)
```
gordonbrander, line 144: Unfortunately, converting from `NSAttributedString` to `AttributedString` proved to be lossy. We lost the links in particular.

I ended up-re-implementing the renderer in terms of the new API.

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
```diff
@@ -0,0 +1,58 @@
...
+import SwiftUI
+
+struct SubtextView: View {
+    private static var renderer = SubtextAttributedStringRenderer()
+    var subtext: Subtext
+
+    var body: some View {
+        VStack(alignment: .leading) {
+            ForEach(subtext.blocks, id: \.self) { block in
+                Text(Self.renderer.render(block.description))
```
bfollington, line 17: We should add `.textSelection(.enabled)` to the `Text` or the `VStack` so that users can copy/share the text in blocks. While this lets you share a block, it seems there is no way to get actual region-based text selection without writing a custom view, see: https://stackoverflow.com/questions/58005434/how-do-i-allow-text-selection-on-a-text-label-in-swiftui

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -145,53 +145,31 @@ struct MemoViewerDetailLoadedView: View {
...
-        // because the editor is (currently) always pointed to
-        // our data.
-        let slashlink = Func.run {
-            guard let basePetname = address.petname else {
-                return sub.slashlink
-            }
-            
-            return sub.slashlink.relativeTo(petname: basePetname)
-        }
-
```
bfollington, line 164: Why was this removed? 

Without this petname links do not work correctly. If I am viewing `@gordon/hello` and tap a link to `@ben` I want `@ben.gordon` but without this code I will just get `@ben`. Only the viewer has this case to worry about, the editor is always starting from your address book as the root but the viewer is relative to the current memo's owner.

xcode/Subconscious/Shared/Parsers/SubtextAttributedStringRenderer.swift
```diff
@@ -10,6 +10,11 @@ import SwiftUI
 /// A type that can be encoded and decoded to `sub://slashlink` URLs.
 /// Used by `SubtextAttributedStringRenderer`.
 struct SubSlashlinkLink {
+    static let schemeKey = "sub"
+    static let hostKey = "slashlink"
+    static let slashlinkKey = "slashlink"
+    static let textKey = "text"
```
bfollington, line 16: 👌

bfollington (changes requested):
Looking good, couple of small issues.

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
gordonbrander: Good catch.

Annoying that there is no easy way to support full-document select across text elements. I can accept that as a tradeoff for now, if it gets us to view-only transclusion faster.

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
gordonbrander: Oops! Mistaken deletion. Fixed.

bfollington (approved):
LGTM!

PR #514, opened 2023-04-20
Implement callback-based Noosphere FFI

### LINKED ISSUE #510: Update Noosphere to 0.9.x non-blocking callback FFI
- Subconscious: #514 
- Noosphere: https://github.com/subconsciousnetwork/noosphere/pull/322

gordonbrander:
Fixes #510 

This PR implements the Noosphere 0.9.x non-blocking callback-based FFI. It wraps these APIs in async functions for ergonomic use.

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
@@ -240,7 +240,7 @@ public actor Sphere: SphereProtocol, SpherePublisherProtocol {
                 self.noosphere.noosphere,
                 sphere,
                 slashlink.description
-            ) { error, file in
+            ) { error, pointer in
                 if let message = Noosphere.readErrorMessage(error) {
                     continuation.resume(
                         throwing: NoosphereError.foreignError(message)
@@ -249,7 +249,7 @@ public actor Sphere: SphereProtocol, SpherePublisherProtocol {
                 }
                 guard let file = SphereFile(
                     noosphere: self.noosphere,
-                    pointer: file
+                    file: pointer
                 ) else {
                     continuation.resume(throwing: NoosphereError.nullPointer)
                     return
--- a/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
@@ -269,7 +269,7 @@ public actor Sphere: SphereProtocol, SpherePublisherProtocol {
         guard let file = try? await open(slashlink: slashlink) else {
             return nil
         }
-        return await file.readHeaderValueFirst(name: name)
+        return try? await file.readHeaderValueFirst(name: name)
     }
     
     /// Read first header value for memo at slashlink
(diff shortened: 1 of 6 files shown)


xcode/Subconscious/Shared/Services/Noosphere/SphereFile.swift
```diff
@@ -0,0 +1,106 @@
...
+                }
+                let data = Data(bytes: contents.ptr, count: contents.len)
+                continuation.resume(returning: data)
+            }
+            return
+        }
+    }
+    
+    deinit {
+        ns_sphere_file_free(file)
```
gordonbrander, line 103: Currently crashing here.

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -28,7 +26,7 @@ public protocol SphereProtocol {
     
     func readHeaderNames(slashlink: Slashlink) async -> [String]
     
-    func read(slashlink: Slashlink) async throws -> Memo
+    func read(slashlink: Slashlink) async throws -> MemoData
```
gordonbrander, line 29: Make type concrete. We were not using this customizable associatedtype anywhere.

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -158,13 +155,35 @@ public actor Sphere: SphereProtocol, SpherePublisherProtocol {
     public let sphere: OpaquePointer
     private let _identity: String
     
-    private init(noosphere: Noosphere, identity: String, sphere: OpaquePointer) throws {
+    /// Initialize a sphere from an already-created sphere pointer, retrieving
+    /// identity from Noosphere.
+    ///
+    /// This is a private initializer used in implementing traversal.
+    private init(
```
gordonbrander, line 162: Create private intializer that calls `ns_sphere_identity` to read identity from already-initialized sphere pointer.

Note: initializers don't support async, so initialization must always be synchronous in Swift. Implementing this private init this way is based on the assumption that as of Noosphere 0.9.x, reading sphere identity from pointer will have minimal latency (will not be blocked on writes, and will not reach out to the network).

Note however that we do only call this initializer from the async method `traverse` (https://github.com/subconsciousnetwork/subconscious/pull/514/files#diff-6deb9e69a5a4545d3d22efbc1696a7f88ab6dc55137b68618eca8063aad0f1ccR605). Calling even a blocking initializer in that context will never block the main thread (if Swift's threadpool is doing its job).

cdata: This is reasonable. This call should never reach out to the network or block for a long period of time.

xcode/Subconscious/Subconscious.xcodeproj/project.pbxproj
```diff
@@ -2327,8 +2335,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/gordonbrander/ObservableStore";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 0.3.0;
+				kind = exactVersion;
```
gordonbrander, line 2338: Locking in this dependency too, while I'm at it.

ObservableStore has some breaking changes in the upcoming 0.4.0. Better to upgrade when we choose to.

cdata (approved):
Very exciting!

PR #537, opened 2023-04-24
Fix redundant unwrap

gordonbrander:
`MemoAddress.toSlashlink()` is no longer an optional return value since https://github.com/subconsciousnetwork/subconscious/commit/db855e99657fbfbcaca3e04dd8b413d6694e52d2#diff-00211e56380835626ac7bd8af84a29b28e53c12b1d0298acef9588b2a942b83fR74.


jsantell (approved):
LGTM! Thank you!

PR #545, opened 2023-04-25
Require valid nickname

### LINKED ISSUE #542: FRE does not require specifying a nickname
> As a user, I want to have my nickname validated in real time, so I can have a valid nickname that works in the protocol.

Discovered while investigating #540.

What happens: the "next" button is enabled in the FRE when the nickname field is in its initial state.

What should happen: the "next" button should only be enabled when a valid nickname has been entered.

gordonbrander:
Fixes #542

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -93,6 +93,8 @@ struct AppView: View {
     }
 }
 
+typealias NicknameFormField = FormField<String, Petname>
+
 //  MARK: Action
 enum AppAction: CustomLogStringConvertible {
     /// Sent immediately upon store creation
@@ -101,6 +103,7 @@ enum AppAction: CustomLogStringConvertible {
     case recoveryPhrase(RecoveryPhraseAction)
     case addressBook(AddressBookAction)
     case appUpgrade(AppUpgradeAction)
+    case nicknameFormField(NicknameFormField.Action)
 
     /// Scene phase events
     /// See https://developer.apple.com/documentation/swiftui/scenephase
@@ -112,7 +115,7 @@ enum AppAction: CustomLogStringConvertible {
     case setAppUpgraded(_ isUpgraded: Bool)
 
     /// Set sphere/user nickname
-    case setNicknameTextField(_ nickname: String)
+    case persistNicknameFormField(_ nickname: String)
 
     /// Set gateway URL
     case setGatewayURLTextField(_ gateway: String)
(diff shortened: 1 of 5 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -298,18 +326,25 @@ enum GatewaySyncStatus: Equatable {
...
-    /// We assign UserDefault property to model property at startup.
+    ///
+    /// This property is updated at `.start` with the value stored in
+    /// `AppDefaults.standard.isNoosphereEnabled`.
+    ///
     /// This property is changed via `persistNoosphereEnabled`, which will
     /// update both the model property (triggering a view re-render)
     /// and persist the new value to UserDefaults.
-    var isNoosphereEnabled = AppDefaults.standard.isNoosphereEnabled
+    var isNoosphereEnabled = false
```
gordonbrander, line 336: Previously we were loading several initializer defaults directly from `AppDefaults`.

This works, but it means that initializing the model becomes stateful, and this does not play well with tests, since the model's values will be the values of whatever got set previously.

I've moved all of these `AppDefaults` values to `.start`. This brings us one step closer to eventually moving `AppDefaults` into the `AppEnvironment`, allowing us to fully mock the `AppDefaults`.

When we upgrade `ObservableStore`, the new factory initializer will help us do even better here. It lets us generate the first state and fx from a special update function, allowing for this kind of "loading from IO" without compromising on mockability and testability https://github.com/subconsciousnetwork/ObservableStore/pull/30

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -684,31 +792,39 @@ struct AppModel: ModelProtocol {
...
+    }
+
+    static func persistNickname(
+        state: AppModel,
+        environment: AppEnvironment,
+        text: String
+    ) -> Update<AppModel> {
+        // Persist any valid value
+        if let validated = state.nicknameFormField.validated {
+            AppDefaults.standard.nickname = validated.description
```
bfollington, line 819: This gives us a nice spot to handle #559 when we get around to it.

xcode/Subconscious/Shared/Components/FirstRun/FirstRunProfileView.swift
```diff
@@ -19,12 +19,12 @@ struct FirstRunProfileView: View {
...
                         text: Binding(
-                            get: { app.state.nicknameTextField },
+                            get: { app.state.nicknameFormFieldValue },
                             send: app.send,
-                            tag: AppAction.setNicknameTextField
+                            tag: AppAction.setNickname
                         ),
                         caption: "Lowercase letters, numbers and dashes only.",
-                        hasError: !app.state.isNicknameTextFieldValid
+                        hasError: !app.state.isNicknameFormFieldValid
```
bfollington, line 27: `!app.state.nicknameFormField.isValid`?

bfollington, line 27: Nit: consistency, see below. 

Do we prefer `app.state.isNicnameFormField.isValid` or `app.state.isNicknameTextFieldValid`?

xcode/Subconscious/Shared/Components/FirstRun/FirstRunProfileView.swift
```diff
@@ -43,7 +43,7 @@ struct FirstRunProfileView: View {
                 .simultaneousGesture(TapGesture().onEnded {
                     app.send(.createSphere)
                 })
-                .disabled(!app.state.isNicknameTextFieldValid)
+                .disabled(!app.state.nicknameFormField.isValid)
```
bfollington, line 46: Nit: consistency, see above. 

Do we prefer `app.state.isNicnameFormField.isValid` or `app.state.isNicknameTextFieldValid`?

bfollington (approved):
LGTM, just one nitpick

xcode/Subconscious/Shared/Components/FirstRun/FirstRunProfileView.swift
gordonbrander: Fixed

Issue #553, opened 2023-04-25
Auto-titling sometimes grabs content from subsequent line


gordonbrander:
Title should always be derived from first line.

PR #557, opened 2023-04-26
Use only the first line to generate entry title

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/553

This does not address the capitalisation of the slug, I'm not sure where we landed on that?

If the first line starts with a capital this will generate a slug with a capital letter but it seems like leaving and reloading the note causes it to become lowercase?

### DIFF
--- a/xcode/Subconscious/Shared/Services/DataService.swift
+++ b/xcode/Subconscious/Shared/Services/DataService.swift
@@ -481,8 +481,10 @@ actor DataService {
         _ text: String,
         audience: Audience
     ) async -> MemoAddress? {
+        let excerpt = Subtext.excerpt(markup: text, fallback: text)
+        
         // If we can't derive slug from text, exit early.
-        guard let slug = Slug(formatting: text) else {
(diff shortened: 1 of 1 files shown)


xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -481,6 +481,25 @@ actor DataService {
...
+            case _:
+                return false
+            }
+        }
+        
+        guard let block = block else {
+            return nil
+        }
+        
+        let text = block.body().toString()
```
bfollington, line 501: This does not address the capitalisation of the string, I'm not sure where we landed on that? If the first line starts with a capital this will generate a slug with a capital letter but it seems like leaving and reloading the note causes it to become lowercase?

gordonbrander (approved):

gordonbrander:
Capitalization of the slug in the Omnibox is a view problem (It's probably using `verbatim` instead of `description`). We can address in a separate issue.

Issue #551, opened 2023-04-25
As a user, I want the editor to autocomplete follows


gordonbrander:
> As a user, I want the editor to autocomplete follows, so I can easily link to my follows.

Issue #518, opened 2023-04-21
As a user, when I am linking to a note, I want to see suggested notes from spheres I follow


gordonbrander:
As a user, when I am linking to a note, I want to see suggested notes from spheres I follow, so I can discover related notes to connect.

PR #561, opened 2023-04-27
Support multiple peer types (did, petname) in slashlink type

### LINKED ISSUE #323: Multiplayer backlinks
## Story

As a user, when I am visiting a note, I want to see related notes from spheres I follow, so that I have ambient awareness of related thoughts my follows have had on this topic.

gordonbrander:
DIDs are the only stable way to address 3p content, since the petname for that content could change. To index 3p content, we need to introduce the concept of an "absolute" slashlink (a slashlink that is rooted to a did, rather than a petname).

To put it another way, Slashlinks are locators. We also need a stable identifier.

This PR introduces several new types to resolve the issue:

- `Peer` is an enum that can be either a Did or a Petname. It makes Slashlink a type that holds a peer and a slug. This concept mirrors the [Subtext Rust parser](https://github.com/subconsciousnetwork/subtext/blob/9d61c0461568d1d6168537955a91221ff1ebc34f/parsers/rust/subtext/src/slashlink.rs#L4) and Noosphere's treatment of Slashlink types.
- `Slashlink` now contains a `Peer` and a `Slug`.
- `Link`, a type containing a `Did` and `Slug`, which provides a stable identifier for indexing.

We use `Slashlink` wherever we can, and in the few cases where we need an identifier (database), we resolve the `Slashlink` to a `Link` at point of use.

This PR also updates the way the database indexes content to use DIDs.

Note this PR deliberately does not address the markup form of did-slashlinks, only the datatype. For markup form, we are tracking here https://github.com/subconsciousnetwork/subtext/issues/51

Lays groundwork toward #323, #551, #518

## Limitations

This PR will not correctly re-relativize the slashlinks of 3P content saved to the database. So if you put something in the database and then query it again, you'll get a slashlink with a did, instead of a petname.

To get nice petname-slashlinks, we'll need to integrate a new FFI function to get a petname from a did https://github.com/subconsciousnetwork/noosphere/issues/358.

However, this PR does not yet index 3P content. It just updates the system so we can. We should be safe to land it, and then start indexing 3P content once we get the new `DID -> Petname` FFI.

## TODO

- [x] Record sync status in database for multiple spheres
- [x] Figure out how to index posts who's petname may have changed
    - [x] Figure out whether `MemoAddress` should hold dids not petnames
- [x] Figure out LosslessStringConvertible regex for Slashlinks with dids
- [x] Update database schema
- [x] Save to database by did + slug
- [x] Filter list queries to constrain to local + own sphere dids
- [x] Update Omnibox to handle did peers.

## QA Checklist

- [x] List notes
- [x] Open note
- [ ] Traverse spheres
- [x] Follow link
- [x] Backlinks
    - [x] List backlinks
    - [x] Follow backlink
    - [x] Relative address for backlinks
- [x] Search notes
- [x] Autocomplete link
    - [x] Autocomplete wikilink
    - [x] Autocomplete slashlink
- [x] Create public note with fallback text
- [x] Create private note with fallback text
- [x] Create public note without fallback text
    - [x] Create public note without fallback text, conflicting slug
- [x] Create private note without fallback text
    - [x] Create private note without fallback text, conflicting slug
- [x] Autosave without text (no-op)
- [x] Autosave does nothing when no change
- [x] Autosave with text
    - [x] Derives slug from text
    - [x] Sets slug on omnibox
    - [x] Sets slug on bottom sheet
    - [x] Sets slug on rename sheet
    - [x] Relative slashlink for bottom sheet
    - [x] Relative slashlink for rename sheet
- [x] Save on back
- [x] Change note link
    - [x] Change note audience
    - [x] Merge note
    - [x] Move note
- [x] Delete note


xcode/Subconscious/Shared/Models/Did.swift
```diff
@@ -7,14 +7,14 @@
 
 import Foundation
 
-struct Did : Hashable, Identifiable, Codable {
-    let did: String
-    var id: String { did }
+public struct Did : Hashable, Identifiable, Codable {
```
gordonbrander, line 10: Make Did public so we can use it in `SphereProtocol`

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -10,14 +10,16 @@ import Combine
 import SwiftNoosphere
 import os
 
+public typealias Cid = String
```
gordonbrander, line 13: Introduce typealias for Cid strings. This will make it easier if we want to add a wrapping type in future (ala Did or Slug).

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -10,14 +10,16 @@ import Combine
...
+
 /// Describes a Sphere.
 /// See `Sphere` for a concrete implementation.
 public protocol SphereProtocol {
-    func identity() async throws -> String
+    func identity() async throws -> Did
```
gordonbrander, line 18: Think in Dids and Cids.

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -133,7 +135,38 @@ actor DataService {
         }
         .eraseToAnyPublisher()
     }
-    
+
+    /// Sync the contents of a specific sphere to the database
+    func syncSphereWithDatabase(petname: Petname) async throws -> String {
```
gordonbrander, line 140: Introduce method to sync database with any Petname.

xcode/Subconscious/Shared/Models/Did.swift
```diff
@@ -7,26 +7,28 @@
...
+    public let did: String
+    public var id: String { did }
     
-    // Approximate, based on https://www.w3.org/TR/did-core/#did-syntax
-    static let regex = /^(did:[a-z0-9]{3}:[a-zA-Z0-9-_\.%:]+)$/
+    // Approximate, based on
+    // https://www.w3.org/TR/did-core/#did-syntax
+    // https://w3c-ccg.github.io/did-method-key/
+    private static let regex = /did:key:[a-zA-Z0-9]+/
```
gordonbrander, line 17: I would prefer to be stricter about this for now and only support `did:key`.

xcode/Subconscious/Shared/Models/Petname.swift
```diff
@@ -71,9 +71,12 @@ public struct Petname:
     
     /// Join a list of petnames into a dotted string, i.e. [foo, bar, baz] -> foo.bar.baz
     /// Names are joined in order of their appearance in `petnames`
-    public init?(petnames: [Petname]) {
-        let petnamePath = petnames.map({ s in s.verbatim }).joined(separator: ".")
-        self.init(petnamePath)
+    public init(petnames: [Petname]) {
```
gordonbrander, line 74: Update so initializer no longer needs to be failable. The petnames are already parsed, so no need to fail.

xcode/Subconscious/Shared/Models/Slashlink.swift
```diff
@@ -16,7 +16,7 @@ public struct Slashlink:
     Codable,
     LosslessStringConvertible
 {
-    private static let slashlinkRegex = /(\@(?<petname>(?:[\w\d\-]+)(?:\.[\w\d\-]+)*))?(\/(?<slug>(?:[\w\d\-]+)(?:\/[\w\d\-]+)*))?/
+    private static let slashlinkRegex = /(?:(?<did>(?:did:key:[a-zA-Z0-9]+))|(?<petname>@[\w\-]+(?:\.[\w\-]+)*))?(?<slug>\/[\w\-]+(?:\/[\w\-]+)*)?/
```
gordonbrander, line 19: This regex matches the following forms:

- `[did][slug]`
- `[petname][slug]`
- `[slug]`
- (nothing)

For simplicity, I am being strict right now and only matching `did:key`.

xcode/Subconscious/Shared/Components/Common/Suggestion/SuggestionLabelView.swift
```diff
@@ -35,6 +35,23 @@ struct SuggestionLabelView: View, Equatable {
...
+        case let .memo(address, title) where address.isLocal:
+            // Do not show full did address for local content
+            Label(
+                title: {
+                    TitleGroupView(
+                        title: Text(
+                            verbatim: !title.isEmpty ? title : empty
+                        ),
+                        subtitle: Text(
+                            verbatim: address.slug.markup
```
gordonbrander, line 47: We relativize own-sphere content. However, we don't really want to relativize local content, as this would be lossy (we would lose the did part that tells us the content is local).

Instead we make an exception case in the view for displaying local content by slug only.

xcode/Subconscious/Shared/Components/Common/SlashlinkDisplayView.swift
```diff
@@ -0,0 +1,120 @@
...
+struct SlashlinkDisplayView: View {
+    var slashlink: Slashlink
+    var baseColor = Color.accentColor
+    var slugColor = Color.accentColor
+    var labelColor = Color.secondary
+    
+    var body: some View {
+        HStack(spacing: 0) {
+            switch slashlink.peer {
+            case let .petname(petname) where slashlink.isProfile:
```
bfollington, line 20: very clean

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -114,69 +108,67 @@ final class DatabaseService {
...
+            INSERT OR REPLACE INTO sphere_sync_info (
+                did,
+                version
             )
             VALUES (?, ?)
             """,
-            parameters: [.text(key), .text(value)]
+            parameters: [.text(sphereIdentity.description), .text(version)]
         )
     }
```
bfollington, line 152: Clever change, very flexible for handling multiple spheres without much disruption.

bfollington (approved):
Some questions and comments but great work. Thanks for adding so many tests!

xcode/Subconscious/Shared/Components/Common/SlashlinkDisplayView.swift
```diff
@@ -0,0 +1,120 @@
...
+
+/// A view for slashlinks, optimized for compact display
+struct SlashlinkDisplayView: View {
+    var slashlink: Slashlink
+    var baseColor = Color.accentColor
+    var slugColor = Color.accentColor
+    var labelColor = Color.secondary
+    
+    var body: some View {
+        HStack(spacing: 0) {
```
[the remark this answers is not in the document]
gordonbrander: It's setting a max line limit on all children.

In general, though, I have a tendency to make sure all views return a single view at the top level, since I had problems with switch statements that return heterogeneous views in earlier versions of SwiftUI (I think this has since been resolved, so this is probably superstition on my part).

xcode/Subconscious/Shared/Components/OmniboxView.swift
```diff
@@ -64,15 +64,30 @@ struct OmniboxSlashlinkView: View {
     
     var body: some View {
         HStack(spacing: 0) {
```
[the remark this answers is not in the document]
gordonbrander: No change from previous. Tried to minimize what I touched. In general, though, I have a tendency to return a single top-level view, since in earlier versions of SwiftUI I had trouble with switches that returned heterogeneous view types.

I think this has since been resolved, so probably just superstition on my part.

xcode/Subconscious/Shared/Models/Did+SubconsciousLocal.swift
```diff
@@ -0,0 +1,128 @@
+//
+//  Did+SubconsciousLocalFile.swift
+//  Subconscious (iOS)
+//
+//  Created by Gordon Brander on 5/1/23.
+//
+//  Extends Did to include a the non-standard `did:subconscious:local`.
+//  We keep the extension here so that Did is not complected with this
```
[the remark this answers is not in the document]
gordonbrander: This is the way. https://youtu.be/Mw7zSQ7ja7Y?t=12

xcode/Subconscious/Shared/Models/Slashlink.swift
```diff
@@ -26,38 +26,36 @@ public struct Slashlink:
...
     let slug: Slug
     
     public var description: String {
-        guard let petname = petname else {
-            return "/\(slug.description)"
+        guard let peer = peer else {
+            return slug.markup
         }
-        return "@\(petname.description)/\(slug.description)"
+        return "\(peer.markup)\(slug.markup)"
```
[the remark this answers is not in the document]
gordonbrander: Yeah intentional. We probably should have been doing it this way to begin with. Each grammatical part should handle its own markup rendering.

This is especially important now that we have the peer type enum. Each enum case may have different grammar, so it must handle its own markup.

xcode/Subconscious/Shared/Models/Slashlink.swift
```diff
@@ -26,38 +26,36 @@ public struct Slashlink:
...
     }
 
     public var verbatim: String {
-        guard let petname = petname else {
-            return "/\(slug.verbatim)"
+        guard let peer = peer else {
+            return "\(slug.verbatimMarkup)"
         }
-        return "@\(petname.verbatim)/\(slug.verbatim)"
+        return "\(peer.verbatimMarkup)\(slug.verbatimMarkup)"
```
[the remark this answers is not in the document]
gordonbrander: Same as above.

xcode/Subconscious/Shared/Models/Link.swift
```diff
@@ -0,0 +1,80 @@
+//
+//  Link.swift
+//  Subconscious (iOS)
+//
+//  Created by Gordon Brander on 5/3/23.
+//
+
+import Foundation
+
+/// An absolute link that can be used as an identifier, as well as a locator.
```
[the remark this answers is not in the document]
gordonbrander: Done!

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -63,18 +65,43 @@ public protocol SphereProtocol {
...
+extension SphereProtocol {
+    /// Resolve a relative slashlink, making it an absolute slashlink.
+    /// - Returns Slashlink with did peer
+    func resolve(slashlink: Slashlink) async throws -> Slashlink {
+        switch slashlink.peer {
+        case .did:
+            return slashlink
+        case .petname(let petname):
+            // Get did for petname
+            let did = try await self.getPetname(petname: petname)
```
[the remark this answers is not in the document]
gordonbrander: The use-cases I have in mind are all related to indexing spheres you immediately follow. However, I think this enhancement would be worthwhile.

Filing a followup: https://github.com/subconsciousnetwork/subconscious/issues/578

xcode/Subconscious/Shared/Models/HeaderSubtext.swift
```diff
@@ -10,14 +10,19 @@ import Foundation
...
         "\(headers.toHeaderString())\(body)"
     }
-
+    
     func toMarkup() -> String {
         self.description
     }
+
+    /// Calculate size of text, including inlined headers
+    func size() -> Int? {
```
gordonbrander, line 23: Had to add this for test. Our actual code-path in the app re-uses the size returned by reading the file metadata. However, we can also calculate it via the string.

This will also introduce the possibility of simplifying the `writeMemo` signature for local content.

PR #562, opened 2023-04-28
Follow subconscious geist automatically after sphere creation

bfollington:
- Implement follow user form sheet on profile view
- Follow subconscious geist on sphere creation


gordonbrander (approved):
Very exciting how little code this takes.

Issue #565, opened 2023-04-29
"Recent" tab lists notes out of order


bfollington:
This list should be sorted reverse chronologically, currently it is sorted by whatever order it comes out of the underlying call to noosphere.

PR #567, opened 2023-04-30
Sort recent entries on profile by recency

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/565

The entries were not sorted at all before. We still do not have an implementation for top entries, but we can probably build it after #323 lands.

### DIFF
--- a/xcode/Subconscious/Shared/Services/UserProfileService.swift
+++ b/xcode/Subconscious/Shared/Services/UserProfileService.swift
@@ -66,8 +66,9 @@ extension UserProfileServiceError: LocalizedError {
 struct UserProfileContentResponse: Equatable, Hashable {
     var profile: UserProfile
     var statistics: UserProfileStatistics
+    var recentEntries: [EntryStub]
+    var topEntries: [EntryStub]
     var following: [StoryUser]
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -292,6 +293,8 @@ actor UserProfileService {
             slugs: notes,
             sphere: self.noosphere
         )
+        let topEntries = entries
```
gordonbrander, line 296: Note that top entries ought to be determined by # incoming backlinks, while recent by modified.  These two queries are both a filter and a sort, so should be separate sql queries, rather than the same query sorted different ways.

gordonbrander (approved):
Approved as a progressive improvement, but with a question.

xcode/Subconscious/Shared/Services/UserProfileService.swift
bfollington: As discussed on our call, I'll put up a PR to hide this tab or just say "coming soon" when you view it for now instead of presenting the same data twice.

Issue #564, opened 2023-04-29
Incorrect second-order sphere traversal from slashlink in content


cdata:
This issue is a follow-up after investigating https://github.com/subconsciousnetwork/noosphere/issues/320

TL;DR I think that whatever is responding to slashlinks being tapped in a note is trying to traverse from the wrong origin sphere (in the case of a second-order traversal).

When performing a second-order traversal (to a peer of my peer), I noticed that I can perform the traversal via the address book of my peer, but if I try to do the same traversal by clicking a slashlink in my peer’s note it fails. See below attached video for an example.

In the example, I first traverse to `@gordon`, and then to `@gordon`'s `try-again-fail-better` slug, and from there I attempt a second-order traversal by clicking the link `@cdata/all-your-base` (which fails). I then demonstrate that I can navigate to `@cdata.gordon` via `@gordon`'s address book.

In diagnosing this issue, I enabled full logging of Noosphere code paths and watched the output in XCode. The logs can tell us which spheres we are traversing from and to whenever a traversal happens. The logs clearly show that when the failed second-order traversal happens, the sphere we are traversing from is not `@gordon`'s, but `@cdata`'s. That is to say: we are effectively trying to go to `@cdata.cdata` (a petname sequence which does not exist) when we actually mean to go to `@cdata.gordon`.

This can be confusing because if I have `@ben` in my address book, and there is also a `@ben.gordon`, such a second-order traversal will appear to work even though it is erroneously traversing from my sphere, not `@gordon`'s.

An example of the relevant logs follows. Note that in both traversals the "origin" is my sphere (`did:key:z6MkjiAUCg157324ekAE2W33YBuWxvVo7fW2pSuUMZS6XqcZ`), even though in the traversal to `cdata` the origin __should__ be `@gordon.cdata` (`did:key:z6MkseshcgDTj2X6vWmqurQ9rJxqhr7jL3HDz2CWzgZ7hGPp`):

```
2023-04-29 00:49:35.649509-0700 Subconscious[3014:1593844] [SubtextTextViewRepresentable] updateUIView
2023-04-29 00:49:35.650095-0700 Subconscious[3014:1593844] [detail] [action] autosave
2023-04-29 00:49:36.170057-0700 Subconscious[3014:1593844] [SubtextTextViewRepresentable] updateUIView
[...]

PR #568, opened 2023-04-30
Fix incorrect sphere traversal

bfollington:
Resolves https://github.com/subconsciousnetwork/subconscious/issues/564

We were not actually computing addresses correctly in either of the possible traversal cases, but for different reasons.


xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -214,7 +214,7 @@ actor UserProfileService {
     
     /// List all the users followed by the passed sphere.
     /// Each user will be decorated with whether the current app user is following them.
-    private func getFollowingList<Sphere: SphereProtocol>(
+    func getFollowingList<Sphere: SphereProtocol>(
         sphere: Sphere,
```
gordonbrander, line 218: I believe the sphere _must_ belong to the same Noosphere instance. It may be better to pass in a sphere identifier, and use `self.noosphere` to get the sphere instance. That way it is impossible to pass the wrong sphere instance.

gordonbrander (changes requested):

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -143,11 +143,12 @@ struct MemoViewerDetailLoadedView: View {
         )
     }
     
-    private func onLink(
+    public static func requestFindDetail(
```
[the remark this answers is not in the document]
bfollington: I moved this logic into the `NotebookModel` and added tests for both the editor and viewer case 👍

xcode/Subconscious/Shared/Services/UserProfileService.swift
bfollington: I had to introduce `NoosphereService.sphere(did: Did) -> Sphere` to actually implement this. Even if you have `NoosphereService` and a `Did` you can't call `Sphere.init()` because the inner `Noosphere` field is private on the service.

I think this solution is fine but just flagging because any change to `NoosphereService` should be considered.

gordonbrander (approved):
LGTM

PR #572, opened 2023-05-04
Introduce Swiftlint

bfollington:
Trying out a basic set of rules for catching mistakes like line length etc. everything is configured as a warning rather than an error right now so the build will never fail for lint reasons. I started with all the rules on and removed anything that was obviously in conflict with our existing style / rules.

We have ~500 lint warnings in the codebase atm and it would be good to fix them because github actions can only display a maximum of 10 linting errors per run (yeah, I know). So lint errors introduced in a PR won't actually show up in the UI until we clear the backlog of underlying lint issues. 

The github action I used (made by the author of Swiftlint) purports to be able to scope the linting to only files changed in the PR but I get errors whenever I try to configure it as per their example.

---

Swiftlint can fix basic issues by running `swiftlint --fix` but we need to pick our moment carefully to avoid conflicts with in-flight work.

You can also choose to integrate swiftlint into Xcode as a build action (see this draft PR https://github.com/subconsciousnetwork/subconscious/pull/573). If we gain enough confidence in rules I would be tempted to add a pre-commit hook that autofixes.

### DIFF
--- a/.github/workflows/swiftlint.yaml
+++ b/.github/workflows/swiftlint.yaml
@@ -14,6 +14,8 @@ jobs:
       - uses: actions/checkout@v3
         with:
             set-safe-directory: '/github/workspace'
+      - name: Fetch PR's target branch
+        run: git fetch --no-tags --prune --depth=1 origin ${{ github.base_ref }}
       - name: GitHub Action for SwiftLint
         uses: norio-nomura/action-swiftlint@master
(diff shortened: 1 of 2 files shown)


gordonbrander (approved):
LGTM.

Let's pick a time to merge this when it won't cause a bunch of disruption.

PR #576, opened 2023-05-08
Simplify `UserProfileService`

bfollington:
Fixes a bug that blocks us from loading profiles, I collapsed the two paths into one to ensure this particular edge-case cannot reoccur.

The actual issue was introduced by `NoosphereService.sphere(did: Did)`, opening a sphere this way DID seems to succeed but leads to null pointer errors down the line. I replaced it with `NoosphereService.sphere(address: MemoAddress)` which will call traverse when appropriate. 

This will need to be adjusted slightly in the wake of https://github.com/subconsciousnetwork/subconscious/pull/561

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -417,11 +417,11 @@ actor NoosphereService:
         switch (address.peer) {
         case .none:
             return try self.sphere()
-        case .some(.did(let did)) where did == identity:
+        case .did(let did) where did == identity:
             return try self.sphere()
-        case .some(.petname(let petname)):
+        case .petname(let petname):
             return try await self.traverse(petname: petname)
-        case _:
+        default:
             throw NoosphereServiceError.cannotFindSphereForUnknownIdentity
         }
     }
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -407,8 +407,14 @@ actor NoosphereService:
         try await self.sphere().traverse(petname: petname)
     }
     
-    func sphere(did: Did) async throws -> Sphere {
-        try Sphere(noosphere: self.noosphere(), identity: did.did)
+    /// Intelligently open a sphere by traversing or, if this is our address, returning the default sphere.
+    func sphere(address: Slashlink) async throws -> Sphere {
+        switch (address.petname) {
+        case .none:
+            return try self.sphere()
+        case .some(let petname):
+            return try await self.traverse(petname: petname)
+        }
     }
     
     nonisolated func traversePublisher(
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -19,11 +19,14 @@ import os
 
 enum NoosphereServiceError: Error, LocalizedError {
     case defaultSphereNotFound
+    case cannotFindSphereForUnknownIdentity
     
     var errorDescription: String? {
         switch self {
         case .defaultSphereNotFound:
             return "Default sphere not found"
+        case .cannotFindSphereForUnknownIdentity:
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -407,8 +407,14 @@ actor NoosphereService:
         try await self.sphere().traverse(petname: petname)
     }
     
-    func sphere(did: Did) async throws -> Sphere {
-        try Sphere(noosphere: self.noosphere(), identity: did.did)
+    /// Intelligently open a sphere by traversing or, if this is our address, returning the default sphere.
+    func sphere(address: MemoAddress) async throws -> Sphere {
```
gordonbrander, line 411: Will need to update in light of #561

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -202,7 +202,10 @@ actor UserProfileService {
 
             entries.append(
                 EntryStub(
-                    address: Slashlink(petname: petname, slug: slug).toPublicMemoAddress(),
+                    address: Slashlink(
+                        petname: address.petname,
+                        slug: slug
+                    ).toPublicMemoAddress(),
```
gordonbrander, line 208: Will need to update in light of #561

gordonbrander (changes requested):

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
bfollington: Updated & tested 👍

xcode/Subconscious/Shared/Services/UserProfileService.swift
bfollington: Updated & tested 👍

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -407,8 +407,14 @@ actor NoosphereService:
         try await self.sphere().traverse(petname: petname)
     }
     
-    func sphere(did: Did) async throws -> Sphere {
-        try Sphere(noosphere: self.noosphere(), identity: did.did)
+    /// Intelligently open a sphere by traversing or, if this is our address, returning the default sphere.
+    func sphere(address: Slashlink) async throws -> Sphere {
```
gordonbrander, line 411: Should we switch on `peer` and also handle the did case?

gordonbrander (approved):

gordonbrander (changes requested):
Must fix traversal logic to include did case.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
bfollington: Handled.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -407,8 +410,20 @@ actor NoosphereService:
...
+        let identity = try await self.identity()
+        
+        switch (address.peer) {
+        case .none:
+            return try self.sphere()
+        case .some(.did(let did)) where did == identity:
+            return try self.sphere()
```
bfollington, line 421: Annoyingly I can't merge these two cases because `did` is only bound in one of them?

Unless I missed some syntax for where clauses + multiple cases.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -407,8 +410,20 @@ actor NoosphereService:
...
+        case .some(.petname(let petname)):
+            return try await self.traverse(petname: petname)
+        case _:
```
gordonbrander, line 424: Nit: prefer `default`?

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
gordonbrander: Nit: you can omit `.some(...)` and just say `.did(let did) where did == identity`

gordonbrander (approved):

Issue #519, opened 2023-04-21
As a user, I want to draft private notes, so I can develop my thoughts before sharing.


gordonbrander:
## Story

As a user, I want to draft private notes, so I can develop my thoughts before sharing.

gordonbrander:
Fixed by #415 introducing the concept of "local draft notes".

PR #616, opened 2023-05-10
Change user-facing instances of "local" to "draft"

### LINKED ISSUE #605: Change "local" to "draft"
Better communicates the value prop.

gordonbrander:
Fixes #605

While the domain model still thinks in "sphere" and "local" scopes, we use "public" and "draft" to refer to these concepts when speaking to the user. "Local" is an implementation detail. "Draft" more closely matches user intent (see #519).


bfollington (approved):
LGTM

PR #623, opened 2023-05-11
Pull to refresh on profile

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/598
Fixes https://github.com/subconsciousnetwork/subconscious/issues/600

Reworked the structure to add empty states and make everything support `.refreshable{}`

https://github.com/subconsciousnetwork/subconscious/assets/5009316/b9f1ff2f-039e-4eb9-bfb9-cb592246aa13

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
@@ -35,14 +35,23 @@ struct UserProfileView: View {
             label: "Recent",
             view: Group {
                 if let user = state.user {
-                    ForEach(state.recentEntries) { entry in
-                        StoryEntryView(
-                            story: StoryEntry(
-                                author: user,
-                                entry: entry
-                            ),
-                            action: { address, _ in onNavigateToNote(address) }
-                        )
+                    List(state.recentEntries) { entry in
+                        VStack {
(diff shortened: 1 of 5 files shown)


xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -33,7 +33,7 @@ struct UserProfileView: View {
     var body: some View {
         let columnRecent = TabbedColumnItem(
```
gordonbrander, line 34: Nit: I tend to move any variables into private instance variables. It used to be the case that the view builder would get tripped up on the combination of statements and expressions.

gordonbrander (approved):
LGTM

gordonbrander:
For future: `.refreshable` now takes an async function. The function will keep the spinner going until the async call completes.

We could consider a special exception to the way we typically handle effects:

```swift
.refreshable {
    // Any setup state
    store.send(.refreshing)
    let value = await store.environment.refreshThing()
    // Pass back into model update logic
    store.send(.refreshed(value))
}
```

bfollington:
@gordonbrander if we had an API to wait for an action on the store we could keep it all in the model?

```swift
.refreshable {
    store.send(.refresh)
    await store.listen(action: .refreshed)
}
```

Issue #632, opened 2023-05-11
Surface user's preferred nickname on profile header and preview cards


bfollington:
We should show the user's nickname from their `_profile_` if they have one. Additionally, in list contexts, we should show the address of the user (i.e. `ben.gordon.chris`).

This is a stepping stone to #615 and is only a small clarification to help us explore further.

Issue #628, opened 2023-05-11
second-order following lists sometimes display nickname rather than petname-path


jsantell:
Unclear specifically what triggers this. When viewing first-order sphere `@gordon`'s following, some spheres are correctly displayed as `@foo.gordon`, though others are simply `@foo` or specifically the sphere's nickname.  This may occur when both not following a second-order sphere and there's a profile description defined.

Issue #619, opened 2023-05-11
Do not use dot-name when suggesting new petname for follow


jsantell:
When following from a sphere's "Following" tab, the sphere's petname is concatenated with my sphere's petname for the sphere, which should be disallowed by noosphere (#379) due to ambiguity.

STR:
* Go to a sphere with petname `foo`
* Find sphere follower with petname `bar`, rendered as `@bar.foo`
* Press to follow and the suggested petname is `bar.foo`

bfollington:
We should use the nickname from the user's `_profile_` here as the default.

Issue #615, opened 2023-05-10
Petname UX


gordonbrander:
This umbrella issue documents user stories, design goals and principles for petname UX.

WIP. We may eventually move it to the wiki.

## Petname design

User stories:
- As a user, I want friendly recognizable names for people I know so I can reference them.
- As a user, I want to see that this is the same person I know, so I can recognize them.
- As a user, I want acquaintances to have familiar names, so I can recognize them at a distance, and get to know them.
- As a user, I want to know if multiple people in my neighborhood follow this person, so I have a sense of whether they are part of my trusted social circle.

Principles:
- Petnames are human-friendly handles
- ["If we show a user a did, we have failed"](https://spritelyproject.org/news/petname-systems.html)
- Petnames should work in a plain text editor
- Petname in markup and petname in UI should not diverge in meaningful ways

Observations:
- The closer someone is, the more trustworthy they are.
    - This has the effect of encouraging small densely-connected networks.
- Trust falls off exponentially with address depth, because real social networks are exponentially distributed (scale-free).
    - Therefore, most useful relationships will be one step out `@bob.alice`
    - This has the effect of encouraging hubs
- The first and last name are the most important. The first name is the account you are referencing. The last name is the root of trust.

Tensions:
- Deep petnames (more than n levels deep) are unfriendly. They don't satisfy the friendliness goals (memorable name, familiarity), since they are long, difficult to read, and must frequently be truncated. They don't satisfy the trust goals (I know this person), because trust falls off exponentially. They are only nominally better than DIDs.
    - Traditional petname systems often include the concept of "[edge names](https://spritely.institute/static/papers/petnames.html)", or names that are one step out (`@bob.alice` in our scheme).
    - Few seem to go deeper, although GNU Name system does.

### Petnames should truncate in the middle

When truncating `@eve.dave.carol.bob.alice`.

- **GOOD**: `@eve...alice`
- **BAD**: `@eve.dave...`

The first name is the nickname for the resource being referred to, the last name is the root of trust.

 We could consider truncating everything > 2 steps out.

```
@bob.alice
@carol...alice
```

Tracking here #680 

### Show either petname or proposed name (nickname)

Keep it simple. If petname, just show petname. Otherwise, just show proposed name with some kind of "scare quote" treatment to indicate it's not a real name until you follow. Proposed names could come from user's profile, or from the petname given by the user you followed to get here.

Scare quotes:

- iOS says `Maybe: Carol`
- Italics
- Quote marks

[...]

PR #635, opened 2023-05-12
Surface nickname to aid identification

bfollington:
From https://github.com/subconsciousnetwork/subconscious/issues/632

> We should show the user's nickname from their _profile_ if they have one. Additionally, in list contexts, we should show the address of the user (i.e. ben.gordon.chris).

> This is a stepping stone to https://github.com/subconsciousnetwork/subconscious/issues/615 and is only a small clarification to help us explore further.

- [x] https://github.com/subconsciousnetwork/subconscious/issues/628
- [x] https://github.com/subconsciousnetwork/subconscious/issues/619

Fixes https://github.com/subconsciousnetwork/subconscious/issues/632
Fixes https://github.com/subconsciousnetwork/subconscious/issues/628
Fixes https://github.com/subconsciousnetwork/subconscious/issues/619

image


xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -10,27 +10,20 @@ import SwiftUI
...
-                .lineLimit(1)
-            
-            let rest = parts[1...]
-                .map { p in p.description }
-                .joined(separator: ".")
-            
-            if rest.count > 0 {
-                // Particular structure to ensure truncation trims the path and never the name
-                Text(".\(rest)")
-                    .lineLimit(1)
```
bfollington, line 32: We should come back and re-think this logic as part of https://github.com/subconsciousnetwork/subconscious/issues/615

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -10,27 +10,20 @@ import SwiftUI
...
+            VStack(alignment: .leading) {
+                Text(petname.description)
+                Text(address.markup)
+                    .foregroundColor(.secondary)
+                    .fontWeight(.regular)
+                    .font(.caption)
             }
+        } else {
+            Text(petname.markup)
+                .lineLimit(1)
```
bfollington, line 26: IMG_3634

xcode/Subconscious/Shared/Components/Common/Profile/ProfilePicSm.swift
```diff
@@ -15,7 +15,7 @@ struct ProfilePicSmImage: View {
         image
             .resizable()
             .aspectRatio(contentMode: .fill)
-            .frame(width: 24, height: 24)
+            .frame(width: 32, height: 32)
```
bfollington, line 18: Looks better with the reworked Story views

xcode/Subconscious/Shared/Models/Petname.swift
```diff
@@ -61,6 +61,14 @@ public struct Petname:
         "@\(self.verbatim)"
     }
     
+    public var leaf: Petname {
+        self.parts().first ?? self
+    }
+    
+    public var root: Petname {
+        self.parts().last ?? self
+    }
```
bfollington, line 70: Unsure on names but I sense this is an important concept

xcode/Subconscious/Shared/Components/Common/Profile/FollowUserSheet.swift
```diff
@@ -99,8 +99,8 @@ struct FollowUserSheetModel: ModelProtocol {
                         .petnameField(.setValue(input: petname.verbatim))
                     )
                 }
-                .catch { error in
-                    Just(FollowUserSheetAction.failToFindUniquePetname)
+                .recover { _ in
+                    FollowUserSheetAction.failToFindUniquePetname
```
gordonbrander, line 103: Nit: I often like to include the error's message in the fail action, so I can log it in the fail handler.

xcode/Subconscious/Shared/Components/Common/Profile/ProfilePicSm.swift
gordonbrander: WFM.

Nit: One thought is that we should restrict ourselves to 2 (Sm, Lg), or max 3 (Sm, Md, Lg) pfp sizes throughout the app. This probably deserves to be an app constant.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -143,11 +143,13 @@ struct UserProfileView: View {
                             onProfileAction(user, action)
                         },
                         onTapStatistics: {
-                            send(
-                                .tabIndexSelected(
-                                    UserProfileDetailModel.followingTabIndex
+                            withAnimation {
+                                send(
```
gordonbrander, line 147: Use `.animation()` modifier on `Update` in `.tabSelected` handler instead.

Why:

- More consistent. Keeps logic in update function
- More control over animations.
    - You can trigger an effect which triggers an animation later on.
    - Can drive animation as a result of model state
- Store joins actions on main thread, to ensure that results of effects never change the UI off main thread (disallowed). This can mean some state changes can happen "next tick". `withAnimation` sets a global state during the same tick that `send` is called, but by the time the state changes, it might be next tick, so you can lose the animation in some cases.

`Update(state: state).animation(.default)` solves all of these issues.

xcode/Subconscious/Shared/Components/Common/Story/StoryEntryView.swift
```diff
@@ -19,37 +19,42 @@ struct StoryEntryView: View {
...
+                    .frame(height: AppTheme.unit * 12)
+                    
+                    Divider()
+                    
+                    Text(story.entry.excerpt)
+                        .padding(AppTheme.tightPadding)
+                }
+                .background(Color.background)
+                .contentShape(Rectangle())
+                }
```
gordonbrander, line 56: Indentation is off

xcode/Subconscious/Shared/Models/Petname.swift
gordonbrander: Yes.

Per some of the bugs @jsantell found, I think in future, we should create a `Petname.Part` type that performs the same sanitization as Petname, but disallows `.`.

Petname should store,

```
let parts: [Part]
let leaf: Part
let root: Part
let description: String {
    parts.map({ part in part.description }).join(".")
}
```

...instead of storing a string.

Non-blocking. We can address in a follow-up.

gordonbrander (changes requested):
Minor notes

xcode/Subconscious/Shared/Models/Petname.swift
bfollington: Yeah this feels like the right shape of solution to me, packing into and unpacking the string representation is wasteful and confusing.

xcode/Subconscious/Shared/Components/Common/Profile/ProfilePicSm.swift
bfollington: I realised I totally missed the existence of the Md size option!

I reabstracted a little bit and merged down to one component `ProfilePic` that accepts a `ProfilePicSize` enum with three options:

1. sm - 24px - backlinks, stacked avatar lists
2. md - 32px - story byline
3. lg - 48px - profile header

xcode/Subconscious/Shared/AppTheme.swift
```diff
@@ -35,6 +35,10 @@ extension AppTheme {
     static let cornerRadius: Double = 8
     static let cornerRadiusLg: Double = 16
     
+    static let smPfpSize: CGFloat = 24
+    static let mdPfpSize: CGFloat = 32
+    static let lgPfpSize: CGFloat = 48
```
gordonbrander, line 40: I like these sizes. They feel right. 👍

xcode/Subconscious/Shared/Components/Common/Profile/ProfilePic.swift
```diff
@@ -13,39 +13,75 @@ enum ProfilePicVariant: Equatable, Codable, Hashable {
...
-            .frame(width: 48, height: 48)
+            .frame(width: size, height: size)
             .clipShape(Circle())
-            .overlay(Circle().stroke(Color.separator, lineWidth: 0.5))
+            .overlay(Circle().stroke(Color.separator, lineWidth: lineWidth))
+    }
+}
+
+extension View {
+    fileprivate func profilePicFrame(size: CGFloat, lineWidth: CGFloat) -> some View {
```
gordonbrander, line 36: Interesting! I've mostly avoided fileprivate, because it is a "fancy" scoping mechanism that might not be familiar to people who aren't used to Swift's odd project-level scoping rules.

It might not be a bad idea to make this a generally available method, and give it is own file `Common/ProfilePickframeViewModifier.swift`

Non-blocking.

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -10,27 +10,20 @@ import SwiftUI
 /// Byline style for displaying a petname
 struct PetnameView: View {
     var petname: Petname
+    var address: Slashlink?
```
gordonbrander, line 13: Should this rather be another `Petname`?

```swift
/// User's preferred name
var nickname: Petname
/// Name given by others
var petname: Petname?
```

gordonbrander (approved):
LGTM, but with one question.

Issue #631, opened 2023-05-11
Nickname is not stored in profile after app setup


bfollington:
We should populate the users's `_profile_.json` as part of the first run process (after capturing the nickname).

We could even remove the concept of the nickname from the AppDefaults entirely and always read from the user's `_profile_` memo.

bfollington:
Duplicate / related to https://github.com/subconsciousnetwork/subconscious/issues/559

Issue #559, opened 2023-04-26
Remove `AppDefaults.nickname` and use `_profile_` instead


bfollington:
We no longer need to store a nickname in `AppDefaults` since a user can save this in their profile.

During onboarding we should still collect the nickname but write it to the underlying `_profile_` memo instead, after sphere creation.

We should still keep `AppDefaults.ownerKeyName` going forward.

gordonbrander:
Note AppDefaults.ownerKeyName got replaced by Config.noosphere.ownerKeyName

PR #639, opened 2023-05-15
Persist nickname in profile memo during onboarding

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/631
Fixes https://github.com/subconsciousnetwork/subconscious/issues/559

# Changes

- Write to `_profile_` during onboarding so every user has a populated nickname
- Remove `AppDefaults.nickname`

# Questions

Should we also remove the Nickname field in the settings? It seems superfluous. We _could_ allow access to the same "Edit Profile" form via the settings view (in addition to editing from the "your profile" view) but two ways to edit something sets off my design spidey sense.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -111,6 +111,10 @@ enum AppAction: CustomLogStringConvertible {
     case succeedCreateInitialProfile
     case failCreateInitialProfile(_ message: String)
     
+    case fetchNicknameFromProfile
+    case succeedFetchNicknameFromProfile(_ nickname: Petname)
+    case failFetchNicknameFromProfile(_ message: String)
+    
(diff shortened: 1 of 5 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -763,9 +795,10 @@ struct AppModel: ModelProtocol {
                 .notifyFirstRunComplete(
                     AppDefaults.standard.firstRunComplete
                 ),
-                .setNickname(
-                    AppDefaults.standard.nickname ?? state.nickname
-                )
+                .fetchNicknameFromProfile
+//                .setNickname(
+//                    AppDefaults.standard.nickname ?? state.nickname
+//                )
```
gordonbrander, line 801: Remove if not needed

gordonbrander (approved):
Remove commented-out code, then good to merge

Issue #621, opened 2023-05-11
Use `ShareLink` for Recovery Phrase instead of copy/paste


bfollington:
Came up in testing (thanks @jsantell!) that it would be nice to send the recovery phrase to another app without leaving Subconscious. 

<img width="237" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/20c4fe6f-c858-452b-b68d-1584dcbdbf16">

We can changeout the `Button` for a `ShareLink`, there are examples of this pattern in the profile meta sheet for sharing the DID.

bfollington:
@gordonbrander it supports it! 
![IMG_127E521C70EC-1](https://github.com/subconsciousnetwork/subconscious/assets/5009316/a173428f-a58a-497e-b008-652b20002f3b)

PR #640, opened 2023-05-15
Replace Copy button with ShareLink

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/621

image

Copy is accessible via the share menu:

image


gordonbrander (approved):
LGTM!

PR #642, opened 2023-05-15
Implement keyboard submit navigation in First Run

### LINKED ISSUE #620: Advance "First Run" wizard using keyboard submit action
When a form consists of a single field, it makes sense to allow the user to submit using the iOS keyboard submit action:

image

https://github.com/subconsciousnetwork/subconscious/pull/614 introduces support on `ValidatedTextField` for providing `submitLabel` and `onSubmit` callbacks but doesn't add this feature since we need to support programmatic navigation.

So, this is actually blocked by refactoring the NavigationStack used in the First Run experience to be driven from the store. After that, we can simply dispatch the same action from tapping buttons and submitting text fields to advance.

bfollington:
Moved the FirstRun navigation path into the store and used that to implement keyboard navigation. The diff is very noisy but most views are unchanged, just removed the wrapping NavigationStack from every step past the welcome view.

# Changes
- Introduce `submitLabel` and `onSubmit` fields to `ValidatedTextField`
- Implement navigation using `onSubmit`
- Move `.createSphere` logic into the model

Fixes #620

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -508,6 +508,14 @@ struct AppModel: ModelProtocol {
             var model = state
             model.firstRunPath = path
             
+            if path.last == .nickname  {
+                return update(
+                    state: model,
+                    action: .createSphere,
+                    environment: environment
+                )
+            }
+            
             return Update(state: model)
         case let .pushFirstRunStep(step):
             var model = state
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -202,6 +202,9 @@ enum AppAction: CustomLogStringConvertible {
     case beginProvisionGateway(String)
     case completeProvisionGateway(URL)
     case failProvisionGateway(String)
+    
+    case setFirstRunPath([FirstRunStep])
+    case pushFirstRunStep(FirstRunStep)
 
     /// Set settings sheet presented?
     case presentSettingsSheet(_ isPresented: Bool)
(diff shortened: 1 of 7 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -484,6 +495,24 @@ struct AppModel: ModelProtocol {
                 state: state,
                 environment: environment
             )
+        case let .setFirstRunPath(path):
+            var model = state
+            model.firstRunPath = path
+            
+            if path.last == .nickname  {
```
gordonbrander, line 502: Generating this action here seems too implicit/indirect to me.

It might be easier to understand the logical flow if the effect were generated by an action somehow. For example, we could send an action when the nickname "next" navigation link is tapped by decorating the navigation link with `.simultaneousGesture`:

```swift
.simultaneousGesture(TapGesture().onEnded({
    store.send(.createSphere)
}))
```

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -20,6 +20,9 @@ struct ValidatedTextField: View {
     var autoFocus: Bool = false
     @FocusState var focused: Bool
     
+    var submitLabel: SubmitLabel = .return
+    var onSubmit: () -> Void = { }
```
gordonbrander, line 24: Nice addition to this control 👍

gordonbrander (changes requested):
Looks great!

One minor change requested.

xcode/Subconscious/Shared/Components/AppView.swift
bfollington: As discussed on call, I moved this into `.onAppear {}` on `FirstStepView` which seems much easier all around.

gordonbrander (approved):
LGTM!

PR #644, opened 2023-05-15
Wait for profile refresh to complete in `.refreshable {}`

bfollington:
# Changes

- Created `.refresh()` helper that is used by the model and the UI

### DIFF
--- a/xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
+++ b/xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
@@ -170,9 +170,9 @@ extension Store<UserProfileDetailModel> {
         }
         
         // Ensure updates are sent on the main thread
-        let send = { action in
-            DispatchQueue.main.async {
-                self.send(action)
+        let send = { (action) -> Void in
+            Task {
+                @MainActor in self.send(action)
             }
         }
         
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -162,6 +163,47 @@ struct EditProfileSheetCursor: CursorProtocol {
     }
 }
 
+extension Store<UserProfileDetailModel> {
```
gordonbrander, line 166: Clever

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -162,6 +163,47 @@ struct EditProfileSheetCursor: CursorProtocol {
...
+    func refresh() async {
+        guard let address = state.user?.address else {
+            return
+        }
+        
+        // Ensure updates are sent on the main thread
+        let send = { action in
+            DispatchQueue.main.async {
+                self.send(action)
+            }
```
gordonbrander, line 176: The new way:

```
Task { @MainActor in self.send(action) }
```

gordonbrander (approved):
One non-blocking suggestion.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
bfollington: Urgh I was looking for something like this.

Issue #588, opened 2023-05-08
Editing profile image can cause crash


jsantell:
We aggressively query for the image URL set in a profile on every change; typing in a link e.g. "http://foo.com/profile.png" makes `NSUrlRequest` for each change:

* `h`
* `ht`
* `htt`
* `http`
* `http:`

The first few are invalid URLs, but the `http:` request goes through and causes an unhandled error, causing the application to hang. We should also be cautious about sending intermediary HTTP requests without confirmation.

bfollington:
Try tightening up the parsing/validation. We are just trusting the swift URL type atm.

PR #646, opened 2023-05-16
Introduce strict validation constructor for URL

bfollington:
Introduce a new URL extension that is stricter than the default `URL(string:)` init. 

Fixes https://github.com/subconsciousnetwork/subconscious/issues/588

(Trivia: I generated this with copilot chat, including the tests! The tests failed at first but after telling it that it produced working code)

### DIFF
--- a/xcode/Subconscious/Shared/Library/URLUtilities.swift
+++ b/xcode/Subconscious/Shared/Library/URLUtilities.swift
@@ -8,6 +8,37 @@
 import Foundation
 
 extension URL {
+    /// Attempt to construct a URL from the passed string and validate its components
+    /// This will only allow http & https URLs with a full hostname
+    init?(validatedString string: String) {
+        guard let url = URL(string: string) else {
+            return nil
+        }
+        
+        var components = URLComponents()
+        components.scheme = url.scheme
+        components.host = url.host
+        components.path = url.path
+        
+        guard let scheme = components.scheme,
+              ["http", "https"].contains(scheme.lowercased()) else {
+            return nil
+        }
+        guard components.host != nil else {
+            return nil
+        }
+        
+        if components.path.isEmpty {
+            components.path = "/"
+        }
+        
+        guard let validatedUrl = components.url else {
+            return nil
+        }
+        
+        self = validatedUrl
+    }
+    
     /// Check if URL is HTTP or HTTPs
     func isHTTP() -> Bool {
         return self.scheme == "http" || self.scheme == "https"
--- a/xcode/Subconscious/Shared/Library/URLUtilities.swift
+++ b/xcode/Subconscious/Shared/Library/URLUtilities.swift
@@ -10,33 +10,20 @@ import Foundation
 extension URL {
     /// Attempt to construct a URL from the passed string and validate its components
     /// This will only allow http & https URLs with a full hostname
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Library/URLUtilities.swift
```diff
@@ -8,6 +8,37 @@
 import Foundation
 
 extension URL {
+    /// Attempt to construct a URL from the passed string and validate its components
+    /// This will only allow http & https URLs with a full hostname
+    init?(validatedString string: String) {
```
bfollington, line 13: Struggling with this name.

`validatedString` seems ambiguous to me, validated _for_ what or _as_ what etc.

Perhaps a better name is `httpUrlString:` to more specifically indicate usage?

gordonbrander: I like `URL(httpUrlString:)`.

More ideas:

```swift
URL(validating: String)
URL(strictlyValidating: String)
URL(validatingHttp: String)
URL(httpString: String)
```

gordonbrander: I think the tense should be present tense `validating`. `validated` suggests it has already been validated, to my eye.

xcode/Subconscious/Shared/Library/URLUtilities.swift
```diff
@@ -8,6 +8,37 @@
...
+        }
+        guard components.host != nil else {
+            return nil
+        }
+        
+        if components.path.isEmpty {
+            components.path = "/"
+        }
+        
+        guard let validatedUrl = components.url else {
```
gordonbrander, line 35: Because `components` is only assigned scheme, host, and path, we will lose other parts of the URL, including fragment, query, etc.

xcode/Subconscious/Shared/Library/URLUtilities.swift
```diff
@@ -8,6 +8,37 @@
...
+    /// This will only allow http & https URLs with a full hostname
+    init?(validatedString string: String) {
+        guard let url = URL(string: string) else {
+            return nil
+        }
+        
+        var components = URLComponents()
+        components.scheme = url.scheme
+        components.host = url.host
+        components.path = url.path
```
gordonbrander, line 21: Question: is it not possible to check the URL's properties, rather than constructing a URLComponents?

Non-blocking.

xcode/Subconscious/SubconsciousTests/Tests_URLUtilities.swift
```diff
@@ -22,4 +22,44 @@ final class Tests_URLUtilities: XCTestCase {
...
+        let url = URL(validatedString: urlString)
+        XCTAssertNil(url)
+    }
+    
+    func testMissingPath() {
+        let urlString = "https://example.com"
+        let url = URL(validatedString: urlString)
+        XCTAssertNotNil(url)
+        XCTAssertEqual(url?.absoluteString, urlString + "/")
+    }
```
gordonbrander, line 64: Would add tests for query, fragment, port, etc.

gordonbrander (changes requested):
A couple questions

gordonbrander (approved):
LGTM!

Issue #584, opened 2023-05-08
Rendering profile as JSON when linked from notes


jsantell:
Adding a link in a note to `_profile_` resolves to a note titled "Your profile" and in the editable textbox is `{"bio":"hello", "nickname":"j", "version":"0.0"}`

jsantell:
This also has the side-effect of showing up when searching or creating a new note; a note entry with "Empty note" and the subtitle "Your profile".

jsantell:
If your profile has not been modified and does not yet exist, opening `/_profile_` creates a new note with the content `_profile_`; modifying profile afterwards goes back into JSON viewer mode, but the phantom `_profile_` (non JSON) still exists in backlinks, possibly indexed

bfollington:
We should detect navigation attempts to any `_profile_` slug in a single codepath to make sure this can't happen.

Additionally, we should protect against creating a note called `_profile_` via tapping a link _or_ manually titling a note.

Issue #637, opened 2023-05-12
Components with separate stores / notify pattern


gordonbrander:
This issue tracks an open question: should we treat the `app` store as a shared message bus?

## Background on current architecture

In general, we aspire toward "one store to rule them all". However, there are a few components we have manage their own store for performance reasons:

- Each major app tab:
  - Notebook
  - Feed (future)
- Detail views in the NavigationStack
  - Editor detail (important for perf, since state is invalidated with each keystroke)
  - Content detail
  - Profile Detail

We also have a single `app` store that is shared between views that need access to certain global resources/state, such as sync status.

This architecture is a pragmatic easement on the "one store to rule them all" policy that leans into SwiftUI's perf characteristics. It also leans into recommendations given by Elm 

We strongly discourage creating store-ful components except in these very specific cases:

- App tabs: appropriate, because according to the iOS HIG, they should never really interact with each other, or navigate from one tab to the other. App tabs are almost like separate apps.
- The main navigation stack: appropriate because it contains a potentially large number of self-contained views that interact minimally with each other.

Any view managing its own store should be considered almost a "mini app" or iframe. There should be minimal communication between store-ful components, and that communication should take the form of "notifications" that a thing has happened elsewhere, rather than "commands" to do something.

Views can use the following pattern to replay select actions from one store to another, in either direction:

```
.onReceive(store.actions.compactMap(Cursor.tagStoreToApp)) { action
    app.send(action)
}
.onReceive(app.actions.compactMap(Cursor.tagAppToStore)) { action
    store.send(action)
}
```

## Treating App as a shared message bus

There are cases where we may wish to relay "notifications" from a store-ful component to a parent component. Currently we do this through a callback parameter from child to parent. This is similar to the way the Elm component pattern works.

However, we also have this shared `app`. It may just be easier to send all notifications through `app`, and allow all store-ful components to respond to whatever `app.actions` they care about. This would treat `app` as a message bus shared between store-ful components.

It's a flat, rather than hierarchical relationship.

bfollington:
I am a big fan of this pattern but it (obviously) comes with trade offs. I find a shared message bus makes it easy extremely easy to evolve the application structure and promotes loose coupling (and avoids a lot of glue code in the process). In game development I've found this open-by-default model useful for building layers of systems that hook into existing systems etc. Perhaps most importantly: _it allows for quick experiments and throwing them away._

[...]

PR #647, opened 2023-05-16
Prevent access to JSON profile editor

bfollington:
- Prevent access to JSON view of own profile
- Ensure nickname is populated

Fixes https://github.com/subconsciousnetwork/subconscious/issues/584

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -1132,13 +1132,6 @@ struct AppModel: ModelProtocol {
     ) -> Update<AppModel> {
         var model = state
         model.isFirstRunComplete = isComplete
-        if isComplete {
-            return update(
-                state: model,
-                action: .fetchNicknameFromProfile,
-                environment: environment
-            )
-        }
         return Update(state: model)
     }
     
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -794,8 +794,7 @@ struct AppModel: ModelProtocol {
                 ),
                 .notifyFirstRunComplete(
                     AppDefaults.standard.firstRunComplete
-                ),
-                .fetchNicknameFromProfile
+                )
             ],
             environment: environment
         ).mergeFx(fx)
@@ -837,7 +836,8 @@ struct AppModel: ModelProtocol {
             state: state,
             actions: [
                 .migrateDatabase,
-                .refreshSphereVersion
+                .refreshSphereVersion,
+                .fetchNicknameFromProfile
             ],
             environment: environment
         )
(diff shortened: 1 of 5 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1132,6 +1132,13 @@ struct AppModel: ModelProtocol {
     ) -> Update<AppModel> {
         var model = state
         model.isFirstRunComplete = isComplete
+        if isComplete {
+            return update(
+                state: model,
+                action: .fetchNicknameFromProfile,
+                environment: environment
+            )
+        }
```
bfollington, line 1141: I've realised that we'd benefit from something like #637 here. 

We want to notify whenever the nickname is changed and re-fetch it from the profile memo to keep `app.state.nickname` properly in-sync.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -381,6 +394,14 @@ private struct EditProfileSheetModifier: ViewModifier {
...
                 }
             }
+            .onReceive(store.actions) { action in
+                switch(action) {
+                case .succeedEditProfile:
+                    app.send(.fetchNicknameFromProfile)
+                case _:
+                    break
+                }
+            }
```
bfollington, line 404: It's not _strictly_ related to the issue this PR is tackling, but I realised that the nickname could end up out-of-sync with `AppModel.nickname` while testing this fix.

I've adopted the proposed approach in #637 instead of implementing several layers of `notify` here.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -938,7 +925,22 @@ struct NotebookModel: ModelProtocol {
...
+        guard slashlink.peer != nil else {
+            return update(
+                state: state,
+                action: .findAndPushMemoEditorDetail(
+                    slug: slashlink.toSlug(),
+                    fallback: fallback
+                ),
+                environment: environment
+            )
+        }
```
bfollington, line 941: This needed to move down _after_ checking for a profile link, we were short circuiting when finding local content but profile's should not be caught by that.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -230,10 +237,10 @@ private extension View {
     }
     
     func editProfile(
-        state: UserProfileDetailModel,
-        send: @escaping (UserProfileDetailAction) -> Void
+        app: Store<AppModel>,
+        store: Store<UserProfileDetailModel>
     ) -> some View {
-      self.modifier(EditProfileSheetModifier(state: state, send: send))
+      self.modifier(EditProfileSheetModifier(app: app, store: store))
```
gordonbrander, line 243: Nit: indentation should be 4 spaces

gordonbrander (approved):
LGTM (one formatting nit)

Issue #492, opened 2023-04-07
Kick off gateway sync after adding new follower


gordonbrander:
We should kick off a gateway sync in the background after adding a new follower. This will make our gateway aware of them so we can immediately traverse to their profile.

Followup to #470 and #490.

## User stories

As a user, when I follow another user, I want to be able to traverse to their profile and content.

Issue #604, opened 2023-05-09
Pull-to-refresh on "loading" profile state


gordonbrander:
Profiles you have just followed should have a loading state. You should also be allowed to pull to refresh.

It would be ok if this was a no-op if there is already a sync in-flight. But the action should be there.

PR #648, opened 2023-05-16
Introduce pending state for unresolved follows

### LINKED ISSUE #555: Sync after following user
You won't get the did immediately, since it will take some time for the name to resolve. Use exponential backoff.

> @cdata: Doing it a couple times over 2-3 seconds will mostly get you the new sphere right away.

bfollington:
Builds on https://github.com/subconsciousnetwork/subconscious/pull/647 because I already threaded `app` into the profile view there.

- [x] Pull to refresh should dispatch `.syncAll` on `app`
- [x] Follow user form is not being reset between usages
- [x] Track which petnames are in-flight
    - [x] Render this as an icon on the card
- [x] Stop waiting if petname is removed from addressbook (check getPetname before resolvePetname)
- [x] Move pending state into AddressBookService, provide a method to query the status and do that check when refreshing the profile

image

# Demo

https://github.com/subconsciousnetwork/subconscious/assets/5009316/1f6c7339-e714-4f76-aa24-9418c2f4104e

Fixes #555 
Fixes https://github.com/subconsciousnetwork/subconscious/issues/492

Related to https://github.com/subconsciousnetwork/subconscious/issues/604

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
@@ -405,7 +405,9 @@ private struct EditProfileSheetModifier: ViewModifier {
                 }
             }
             .onReceive(store.actions) { action in
-                switch(action) {
+                switch action {
+                case .refresh:
+                    app.send(.syncAll)
                 case .succeedEditProfile:
                     app.send(.fetchNicknameFromProfile)
                 case _:
--- a/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
@@ -404,16 +404,10 @@ private struct EditProfileSheetModifier: ViewModifier {
                     )
                 }
             }
-            .onReceive(store.actions) { action in
-                switch action {
-                case .refresh:
-                    app.send(.syncAll)
-                case .succeedEditProfile:
-                    app.send(.fetchNicknameFromProfile)
-                case _:
-                    break
-                }
-            }
+            .onReceive(
+                store.actions.compactMap(AppAction.from),
+                perform: app.send
+            )
     }
 }
 
--- a/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
@@ -58,7 +58,7 @@ struct StoryUserView: View {
                         
                     Spacer()
                     
-                    switch story.resolutionStatus {
+                    switch story.user.resolutionStatus {
                     case .unresolved:
                         Image(systemName: "exclamationmark.triangle.fill")
                             .foregroundColor(.secondary)
@@ -79,7 +79,7 @@ struct StoryUserView: View {
                     }
                     
                 }
-                .disabled(story.resolutionStatus != .resolved)
+                .disabled(story.user.resolutionStatus != .resolved)
                 
                 Menu(
                     content: {
@@ -129,10 +129,6 @@ struct StoryUserView: View {
         }
         .contentShape(.interaction, RectangleCroppedTopRightCorner())
         .onTapGesture {
-            guard story.resolutionStatus == .resolved else {
-                return
-            }
-            
             action(
                 story.user.address,
                 story.user.bio.text
@@ -154,10 +150,10 @@ struct StoryUserView_Previews: PreviewProvider {
                         address: Slashlink(petname: Petname("ben.gordon.chris.bob")!),
                         pfp: .image("pfp-dog"),
                         bio: UserProfileBio("Ploofy snooflewhumps burbled, outflonking the zibber-zabber."),
-                        category: .human
+                        category: .human,
+                        resolutionStatus: .unresolved
                     ),
-                    isFollowingUser: false,
-                    resolutionStatus: .unresolved
+                    isFollowingUser: false
                 ),
                 action: { _, _ in }
             )
@@ -169,10 +165,10 @@ struct StoryUserView_Previews: PreviewProvider {
                         address: Slashlink(petname: Petname("ben.gordon.chris.bob")!),
                         pfp: .image("pfp-dog"),
                         bio: UserProfileBio("Ploofy snooflewhumps burbled, outflonking the zibber-zabber."),
-                        category: .human
+                        category: .human,
+                        resolutionStatus: .pending
                     ),
-                    isFollowingUser: true,
-                    resolutionStatus: .pending
+                    isFollowingUser: true
                 ),
                 action: { _, _ in }
             )
@@ -184,10 +180,10 @@ struct StoryUserView_Previews: PreviewProvider {
                         address: Slashlink(petname: Petname("ben.gordon.chris.bob")!),
                         pfp: .image("pfp-dog"),
                         bio: UserProfileBio("Ploofy snooflewhumps burbled, outflonking the zibber-zabber."),
-                        category: .you
+                        category: .you,
+                        resolutionStatus: .resolved
                     ),
-                    isFollowingUser: false,
-                    resolutionStatus: .resolved
+                    isFollowingUser: false
                 ),
                 action: { _, _ in }
             )
@@ -199,10 +195,10 @@ struct StoryUserView_Previews: PreviewProvider {
                         address: Slashlink(petname: Petname("ben.gordon.chris.bob")!),
                         pfp: .image("pfp-dog"),
                         bio: UserProfileBio.empty,
-                        category: .you
+                        category: .you,
+                        resolutionStatus: .pending
                     ),
-                    isFollowingUser: false,
-                    resolutionStatus: .pending
+                    isFollowingUser: false
                 ),
                 action: { _, _ in }
             )
--- a/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
@@ -79,7 +79,7 @@ struct StoryUserView: View {
                     }
                     
                 }
-                .disabled(story.user.resolutionStatus != .resolved)
+                .disabled(!story.user.resolutionStatus.isReady)
                 
                 Menu(
                     content: {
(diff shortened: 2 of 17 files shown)


xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
```diff
@@ -47,26 +47,34 @@ struct StoryUserView: View {
...
+                                .foregroundColor(.secondary)
+                        case (_, .you):
+                            Image.from(appIcon: .you(colorScheme))
+                                .foregroundColor(.secondary)
+                        case (_, _):
+                            EmptyView()
+                        }
+                    }
                 }
+                .disabled(!story.isResolved)
```
bfollington, line 77: I can't disable the entire view because it would block the `Menu` from opening (which contains the shortcut to unfollow)

xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
```diff
@@ -77,7 +85,7 @@ struct StoryUserView: View {
                                 },
                                 label: {
                                     Label(
-                                        title: { Text("Unfollow \(story.user.nickname.markup)") },
+                                        title: { Text("Unfollow") },
```
bfollington, line 88: It's repetitious to see the name again and ugly when it wraps the line.

xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
```diff
@@ -47,26 +47,39 @@ struct StoryUserView: View {
...
+                        case (_, .you):
+                            Image.from(appIcon: .you(colorScheme))
+                                .foregroundColor(.secondary)
+                        case (_, _):
+                            EmptyView()
+                        }
+                    }
+                    
                 }
+                .disabled(story.resolutionStatus != .resolved)
```
bfollington, line 82: I can't disable the entire view because it would block the Menu from opening (which contains the shortcut to unfollow), so just disabling this `Group` to get the visual effect.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -288,7 +294,17 @@ struct UserProfileDetailModel: ModelProtocol {
...
             model.recentEntries = content.recentEntries
-            model.following = content.following
+            model.following = content.following.map { f in
+                if let petname = f.user.address.petname,
+                   model.pendingFollows.contains(petname) {
+                    
+                    var user = f
+                    user.resolutionStatus = .pending
+                    return user
+                }
```
bfollington, line 304: I am not in love with this, but the model is the right place to track pending follows imo.

I experimented with tracking them in the `AddressBookService` but that introduces some nuanced sequenced update problems, since the model ultimately needs access to the pending list to render the UI.

I think we might want to move this up to an app-level concern ultimately? All these background processes are owned the by profile view which is discarded if we pop the navigation stack.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -288,7 +294,17 @@ struct UserProfileDetailModel: ModelProtocol {
...
+                if let petname = follow.user.address.petname,
+                   model.pendingFollows.contains(petname) {
+                    
+                    var user = follow
+                    user.resolutionStatus = .pending
+                    return user
+                }
+                
+                return follow
+            }
```
bfollington, line 307: I am not in love with this, but the model is the right place to track pending follows imo.

I experimented with tracking them in the AddressBookService but that introduces some nuanced sequenced update problems, since the model ultimately needs access to the pending list to render the UI.

I think we might want to move this up to an app-level concern ultimately? All these background processes are owned the by profile view which is discarded if we pop the navigation stack.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -289,7 +293,18 @@ struct UserProfileDetailModel: ModelProtocol {
...
+                   model.pendingFollows.contains(petname),
+                   follow.resolutionStatus != .resolved {
+                    
+                    var user = follow
+                    user.resolutionStatus = .pending
+                    return user
+                }
+                
+                return follow
+            }
```
bfollington, line 307: I am not in love with this, but the model is the right place to track pending follows imo.

I experimented with tracking them in the AddressBookService but that introduces some nuanced sequenced update problems, since the model ultimately drives everything and needs access to the pending list to render the UI.

I think we might want to move this up to an app-level concern ultimately? All these background processes are owned the by profile view which is discarded if we pop the navigation stack. Meaning that if I follow a user, they're pending resolution and I go back then forward again the user will not longer be "pending".

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -405,7 +405,9 @@ private struct EditProfileSheetModifier: ViewModifier {
                 }
             }
             .onReceive(store.actions) { action in
-                switch(action) {
+                switch action {
+                case .refresh:
+                    app.send(.syncAll)
```
bfollington, line 410: Ensure everything is refreshed if a user is trying to manually refresh their following list.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -357,6 +372,15 @@ struct UserProfileDetailModel: ModelProtocol {
...
             var model = state
             model.isFollowNewUserFormSheetPresented = presented
+            
+            if presented {
+                return update(
+                    state: model,
+                    action: .followNewUserFormSheet(.form(.reset)),
+                    environment: environment
+                )
+            }
```
bfollington, line 382: I noticed this form wasn't resetting between uses

bfollington:
@gordonbrander not related to this work but I think I'll disable the swiftlint github check, it just adds too much noise to the PR review. I'm using it in Xcode locally so I might update #573 to remove the github action as well.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
gordonbrander: Another way we do this is through the slightly more compact/fancy approach of:

```swift
.onReceive(
    store.actions.compactMap(AppAction.from),
    perform: app.send
)
```

where `AppAction.from` is an extension that creates an app action from a local action, or else nil. The advantage being that the `AppAction.from` extension can be tested.

Non-blocking.

xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
```diff
@@ -116,6 +129,10 @@ struct StoryUserView: View {
         }
         .contentShape(.interaction, RectangleCroppedTopRightCorner())
         .onTapGesture {
+            guard story.resolutionStatus == .resolved else {
+                return
+            }
```
gordonbrander, line 134: Preference for this no-op to belong to the update function, rather than the view. When actions handle their own no-op based on model state, it's more easily testable. It also makes sending the action foolproof.

Non-blocking.

xcode/Subconscious/Shared/Library/Func.swift
```diff
@@ -58,6 +62,8 @@ struct Func {
     ) async throws -> T? {
         do {
             return try await perform(attempts)
+        } catch RetryError.cancelled {
+            return nil
```
gordonbrander, line 66: Open question, but would it be better to have this retry with backoff logic be part of an update function instead? (If fx fails, send another after a growing delay).

The advantage would be that retry would be forced to fold with app state at each step. The actions would be sent through the logs, and current backoff value would be written to the model where we can inspect it.

Non-blocking.

xcode/Subconscious/Shared/Models/Petname.swift
```diff
@@ -7,6 +7,12 @@
 
 import Foundation
 
+enum PetnameResolutionStatus: Equatable, Hashable, Codable {
```
gordonbrander, line 10: Qs:

- Should this be a general enum, similar to `LoadingState`?
  - Is this the same semantics as `LoadingState`? It's ok if subtly different.
  - If different, still worth generalizing this I bet. We have a lot of places that have the same network resolution lifecycle.
- Is there a resolved success/failure? (`resolved(Result)`)

Non-blocking.

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -9,9 +9,10 @@ import os
 import Foundation
 import Combine
 
-struct AddressBookEntry: Equatable {
+struct AddressBookEntry: Equatable, Hashable, Codable {
```
gordonbrander, line 12: 👍

gordonbrander (approved):
LGTM with a couple questions

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
bfollington: Clean! I like that this moves the mapping nearer to the model and away from the view.

xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
bfollington: I introduced a new action `requestNavigateToProfile` which can result in `pushDetail` if resolved or `failPushDetail` if not. This logs a debug message and does nothing else atm.

xcode/Subconscious/Shared/Models/Petname.swift
bfollington: I dropped the `Petname` prefix and extracted the type and added a `Cid` payload to `resolved` since I currently view the definition of resolution as "take this address and give me the CID on the other end".

xcode/Subconscious/Shared/Library/Func.swift
bfollington: I can see the value of this but it could be a little over-complicated right now. 

The main issue I see is that multiple of these processes can be running at once (rapidly following multiple users before the last one resolves), so tracking the state will require attaching an ID to each background job etc.

I'd like to think a bit more about the right abstraction to make it trivial for different components to sign up for this functionality and share the code, put some initial thoughts down here: https://github.com/subconsciousnetwork/subconscious/issues/658

Issue #587, opened 2023-05-08
Handling verbose profile descriptions


jsantell:
When using a very large profile description (lots of long strings and new lines < 280 characters), a few problems occurred:

* Editing large amounts of text was a challenge within the small space, but the carot moving around, jumping offscreen, deleting unintended characters would make it frustrating if my content wasn't nonsense.
* The full description is truncated on profile view (probably a good idea), but with no way to view the description in its entirety.
* The 280/280 indicator prevents descriptions of exactly 280.

jsantell:
Via https://github.com/subconsciousnetwork/subconscious/issues/584, and with all sphere data with user input, the 280 limit can be disregarded through other means. When attempting to load a >=280 character description, confirming out of the prompt deleted the whole profile description.

PR #650, opened 2023-05-17
Process user profile bio before reading or writing JSON

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/587

# Changes

- Bio is processed before reading or writing to change all whitespace to spaces & truncate at the max length
  - Even with #647 users could still modify their profile to inject massive bio strings etc. 
- Remove line limit on bio editor, it makes life harder than it has to be

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Profile/EditProfileSheet.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/EditProfileSheet.swift
@@ -89,7 +89,7 @@ struct EditProfileSheetModel: ModelProtocol {
     var bioField: FormField<String, String> = FormField(
         value: "",
         validate: { value in
-            value.count >= 280 ? nil : value
+            value.count > 280 ? nil : value
         }
     )
     var pfpUrlField: FormField<String, URL> = FormField(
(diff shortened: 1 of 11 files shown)


xcode/Subconscious/Shared/Components/Common/Profile/EditProfileSheet.swift
```diff
@@ -89,7 +89,7 @@ struct EditProfileSheetModel: ModelProtocol {
     var bioField: FormField<String, String> = FormField(
         value: "",
         validate: { value in
-            value.count >= 280 ? nil : value
+            value.count > 280 ? nil : value
```
bfollington, line 92: Off by one 🥲

gordonbrander: Should this be factored out into a function, and add a regression test?

`Prose.swift` is a good spot for stuff like this. We may want to add a constant or int-valued enum (something like `bioCount` or `shortTextCount`), and a utility fn (e.g. `isShortText(_ text: String)`).

Non-blocking.

gordonbrander (approved):
LGTM. Two non-blocking suggestions.

PR #656, opened 2023-05-19
Wait to provision gateway until after we have the invite code

bfollington:
Gateway provisioning was broken because we tried to start it before we have the invite code. We should probably decouple the sphere creation and gateway provisioning further, but this fixes the immediate issue.


gordonbrander (approved):
LGTM

PR #662, opened 2023-05-24
Ensure `requestCreateInitialProfile` is idempotent

bfollington:
Builds on https://github.com/subconsciousnetwork/subconscious/pull/661

This was causing the profile to be trashed and recreated **whenever** the nickname changed regardless of whether the user already had a profile, erasing the bio. This explains a decent chunk of the weirdness we saw in playtesting.

### DIFF
--- a/xcode/Subconscious/Shared/Services/UserProfileService.swift
+++ b/xcode/Subconscious/Shared/Services/UserProfileService.swift
@@ -51,7 +51,7 @@ extension UserProfileServiceError: LocalizedError {
             }
         case .profileAlreadyExists:
             return String(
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -137,7 +142,7 @@ actor UserProfileService {
                 guard let string = String(data: data.body, encoding: .utf8) else {
                     throw UserProfileServiceError.failedToDeserializeProfile(error, nil)
                 }
-               
+
```
bfollington, line 145: 😡

gordonbrander (approved):
LGTM

PR #663, opened 2023-05-24
Use `go` label for keyboard submit button

bfollington:
This ensures it appears in blue:

image

Of the available options:
- done
- go
- send
- join
- route
- search
- return
- next
- continue

Only three are blue (that I've found anyway):
- done
- go
- search


gordonbrander (approved):
LGTM

Issue #657, opened 2023-05-19
Following the same DID under two petnames and then unfollowing one unfollows both


bfollington:
I don't think this is desirable behaviour, but there is an edgecase where it sort of makes sense:

You're browsing another user's profile. You look at their following list and see that you also follow someone that they follow. This user's card present a content menu button.

In the context menu you can request to unfollow that overlapping user, but it's unclear which petname in your addressbook should be removed.

PR #666, opened 2023-05-24
Distinguish between full petnames and petname parts

### LINKED ISSUE #629: Disallow "." in petnames
Related to #619, `"."` is allowed when adding a new petname, leading to ambiguities when referencing later

bfollington:
Fixes #629 
Fixes https://github.com/subconsciousnetwork/subconscious/issues/657

# Changes

This PR introduces the `Petname.Name` type. `Petname` is now composed of 1 or more `Name` fragments, which are typeset with `.` in between. For example `Petname(parts: [Petname.Name('hello'), Petname.Name('world')])` is rendered as `@hello.world`.

Throughout the codebase we were freely mixing these concepts before, so I've changed the all relevant places (🤞) to use `Name` instead. I have tried to clarify the naming scheme for parameters:

- `nickname`: a name that a user has given themselves, comes from `_profile_`
- `petname`: the disambiguated, relative name assigned to a user
- `name` / `preferredName`: a single `Petname.Part` to show in the UI, derived from address & profile

It's possible I have not been 100% consistent here, especially around `SphereProtocol` which has unfortunately ambiguous method names (resolvePetname takes a `Petname`, getPetname takes a `Petname.Part`). We _could_ change those method names, but the underlying FFI names will still be ambiguous.

# Tasks

- [x] Add tests for `Petname.Name`
- [x] Test that `Pename.append` behaves correctly for two higher-order petnames
- [x] Explain mental model

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -107,7 +107,7 @@ enum AppAction: CustomLogStringConvertible {
     case persistNickname(_ nickname: String)
     
     /// Write to `Slashlink.ourProfile` during onboarding
-    case requestCreateInitialProfile(_ nickname: String)
+    case requestCreateInitialProfile(_ nickname: PetnamePart)
     case succeedCreateInitialProfile
     case failCreateInitialProfile(_ message: String)
     
@@ -399,12 +399,12 @@ struct AppModel: ModelProtocol {
     /// Validated nickname
     ///
     /// This property is updated at `.start` with the corresponding value
-    /// stored in `AppDefaults`.
+    /// stored in the user's `_profile_` memo.
     var nickname = ""
     /// Nickname form
     ///
     /// This property is updated at `.start` with the corresponding value
-    /// stored in `AppDefaults`.
+    /// stored in the user's `_profile_` memo.
     var nicknameFormField = NicknameFormField(
         value: "",
         validate: { value in PetnamePart(value) }
@@ -909,14 +909,14 @@ struct AppModel: ModelProtocol {
         text: String
     ) -> Update<AppModel> {
         // Persist any valid value
-        if let validated = state.nicknameFormField.validated {
+        if let validated = PetnamePart(text) {
             var model = state
             model.nickname = validated.description
             logger.log("Nickname saved: \(validated)")
             
             return update(
                 state: model,
-                action: .requestCreateInitialProfile(text),
+                action: .requestCreateInitialProfile(validated),
                 environment: environment
             )
         }
@@ -928,7 +928,7 @@ struct AppModel: ModelProtocol {
     static func requestCreateInitialProfile(
         state: AppModel,
         environment: AppEnvironment,
-        nickname: String
+        nickname: PetnamePart
     ) -> Update<AppModel> {
         let fx: Fx<AppAction> = Future.detached {
             try await environment.userProfile.requestSetOurInitialNickname(nickname: nickname)
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -80,7 +80,7 @@ struct AppView: View {
 }
 
 typealias InviteCodeFormField = FormField<String, InviteCode>
-typealias NicknameFormField = FormField<String, Petname.Part>
+typealias NicknameFormField = FormField<String, Petname.Name>
 
 // MARK: Action
 enum AppAction: CustomLogStringConvertible {
@@ -107,12 +107,12 @@ enum AppAction: CustomLogStringConvertible {
     case persistNickname(_ nickname: String)
     
     /// Write to `Slashlink.ourProfile` during onboarding
-    case requestCreateInitialProfile(_ nickname: Petname.Part)
+    case requestCreateInitialProfile(_ nickname: Petname.Name)
     case succeedCreateInitialProfile
     case failCreateInitialProfile(_ message: String)
     
     case fetchNicknameFromProfile
-    case succeedFetchNicknameFromProfile(_ nickname: Petname.Part)
+    case succeedFetchNicknameFromProfile(_ nickname: Petname.Name)
     case failFetchNicknameFromProfile(_ message: String)
     
     case setInviteCode(_ inviteCode: String)
@@ -407,7 +407,7 @@ struct AppModel: ModelProtocol {
     /// stored in the user's `_profile_` memo.
     var nicknameFormField = NicknameFormField(
         value: "",
-        validate: { value in Petname.Part(value) }
+        validate: { value in Petname.Name(value) }
     )
     /// Expose read-only value for view
     var nicknameFormFieldValue: String {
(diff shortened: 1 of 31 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -909,14 +909,14 @@ struct AppModel: ModelProtocol {
         text: String
     ) -> Update<AppModel> {
         // Persist any valid value
-        if let validated = state.nicknameFormField.validated {
+        if let validated = Petname.Part(text) {
```
bfollington, line 912: We had a possible desync here between the passed `text` and reading from the form field.

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -9,37 +9,56 @@ import SwiftUI
...
+    }
     
     var body: some View {
-        if let petname = petname {
+        if address != nil {
             VStack(alignment: .leading) {
-                Text(nickname.description)
-                Text(petname.markup)
+                Text(petname.description)
+                Text(identifier)
```
bfollington: This is probably going to change again in #664

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -405,21 +405,30 @@ actor AddressBookService {
         await self.addressBook.invalidateCache()
     }
     
+    private static func shouldBeUnfollowed(
+        _ entry: AddressBookEntry,
+        _ did: Did,
+        _ petname: Petname.Part?
+    ) -> Bool {
+        entry.did == did && (petname == nil || entry.identifier == petname)
+    }
```
bfollington, line 414: This addresses https://github.com/subconsciousnetwork/subconscious/issues/657 by trying to optionally refine the unfollow request using a passed petname.

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -9,37 +9,52 @@ import SwiftUI
 
 /// Byline style for displaying a petname
 struct PetnameView: View {
-    var nickname: Petname
-    var petname: Petname?
+    var address: Slashlink?
+    var petname: Petname
```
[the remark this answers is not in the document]
bfollington: This `petname` param should _maybe_ be called `identifier`...

xcode/Subconscious/Shared/Models/Petname.swift
```diff
@@ -63,66 +45,172 @@ public struct Petname:
...
-           let number = Int(numberString) {
-            return Petname(formatting: "\(match.output.petname)\(separator)\(String(number + 1))")
-        } else {
-            return Petname(formatting: "\(match.output.petname)\(separator)1")
-        }
+        self.init(parts: mappedParts)
+    }
+     
+    /// Combines two petnames to build up a traversal path
+    /// i.e. `Petname("foo")!.append(petname: Petname.Part("bar")!)` -> `bar.foo`
```
gordonbrander, line 104: Update comment to `Petname.Name`

xcode/Subconscious/Shared/Models/Petname.swift
```diff
@@ -63,66 +45,172 @@ public struct Petname:
...
+    public func append(name: Petname.Name) -> Petname? {
```
gordonbrander, line 105: Because of the confusing ordering (append prepends), I might suggest another verb. Ideas:

- join `alice.join(bob)`
- subjoin (join under)

gordonbrander (changes requested):

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -185,62 +185,62 @@ actor AddressBook<Sphere: SphereProtocol> {
...
+        petname: Petname.Name
     ) -> AnyPublisher<Void, Error> {
         Future.detached(priority: .utility) {
           try await self.unsetPetname(petname: petname)
         }
         .eraseToAnyPublisher()
     }
 
-    func listPetnames() async throws -> [Petname] {
+    func listPetnames() async throws -> [Petname.Name] {
```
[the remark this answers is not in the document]
bfollington: I think this depends on the semantics of what `Petname` as a type actually represents. 

One major question to determine this is whether it makes sense to say that a user's _nickname_ is a `Petname` or whether it's just a `Name`. If it is, then a `Petname` is really just a certain string format. 

BUT it seems like in Noosphere we often refer to a `Petname` as a _resolvable_ name, not just a well-formatted string. I actually don't feel strongly either way, but because of the inertia I would maybe lean towards the Noosphere model?

Worth saying: we could still return a first-order `Petname` (as opposed to a `Name`) from some or all of these Noosphere methods. Perhaps the Noosphere layer only accepts `Petname`s but the app sometimes has to model other, more general `Name`s?

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -141,12 +141,18 @@ enum UserCategory: Equatable, Codable, Hashable, CaseIterable {
...
     let address: Slashlink
     let pfp: ProfilePicVariant
     let bio: UserProfileBio
     let category: UserCategory
     let resolutionStatus: ResolutionStatus
+    
+    private static let unknownName = Petname.Name("unknown")!
+    
+    var preferredName: Petname.Name {
+        nickname ?? address.petname?.leaf ?? Self.unknownName
```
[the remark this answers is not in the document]
bfollington: I thought about this some more and reabstracted. There are now two fields: `preferredName: Petname.Name?` and `displayName: String`

`preferredName` is the name we would use when suggesting to follow a user, first trying their nickname then falling back to the address if possible. There may not be a preferred name at all.

`displayNamer` is a string we can print on the UI that always has a value. I realised I could fall back to the last 4 chars of the DID instead, so there is no "unknown" string anymore.

xcode/Subconscious/Shared/Models/Petname.swift
```diff
@@ -63,66 +45,172 @@ public struct Petname:
...
-        self.parts().first ?? self
+    public var isFirstOrder: Bool {
+        self.parts.count == 1
     }
     
-    public var root: Petname {
-        self.parts().last ?? self
+    public var leaf: Petname.Name {
+        // Invariant: parts.count > 0
+        self.parts.first!
```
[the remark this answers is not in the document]
bfollington: So, interestingly, if I try and switch to member fields (`var` or `let`) to store `leaf` and `root` on initialization of Petname, I start getting bad access errors (reading 0x0) while running the tests in seemingly unrelated code. (`Slug.toSlashlink` specifically)

I have no idea what is going on there, I suspect it's related to `Codable` but I'm stumped. Leaving as-is for now.

gordonbrander (changes requested):
Only one small change requested (fix stray `do` block). Otherwise LGTM.

gordonbrander (approved):
LGTM

PR #669, opened 2023-05-29
Fix seeded RNG implementation

bfollington:
Fixes a bug in my RNG implementation that resulted in an incomplete random distribution.

Before:

After:


gordonbrander (approved):
LGTM

A small code comment about the what and why of this might be nice. Non-blocking.

PR #670, opened 2023-05-30
Prevent nickname truncation during onboarding

bfollington:
Due to an oversight in https://github.com/subconsciousnetwork/subconscious/pull/662 we are truncating the nickname to only the first letter entered right now.

I hadn't realised that the nickname is saved on every keystroke, so my idempotent profile creation would create the profile immediately then never update it again. Instead, I've renamed the methods and updated the logic to always update the nickname but preserve the rest of the profile as-is. Y'know, actual idempotency.


gordonbrander (approved):
LGTM

Issue #683, opened 2023-06-05
Possible race condition when saving content


cdata:
I think I may have stumbled upon a race condition related to saving notes this weekend. In this example, I write a new note, add a link, tap through creating another new (draft) note, and add a link back. I’ve seen various permutations of this problem, where the first note I added can get into an “old” state and then be saved that way leading to data loss. Please find example video attached.

Steps to reproduce (although reproduction isn't always exactly what you see in the video):

1. Press FAB to create a new note
2. Start typing, then select "New note (public note)"
3. In the editor, append a link to the note to some other note that does not yet exist
4. Tap the link to the other note that does not yet exist, creating it
5. In the editor for the new note, append a link back to the note you originally created
6. Tap the link back to the original note
7. Hit the back button until you arrive back at the home screen

You may see the bug as you traverse back trough history, but I've also seen it happen _after_ I get back to the home screen and open the first note I created.

https://github.com/subconsciousnetwork/subconscious/assets/240083/4d2e725d-37e1-4010-9b66-a20cd22abf76

PR #686, opened 2023-06-06
Ensure draft editor loads latest version of its content

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/683

When initially presented, the draft editor has no address associated with it (because the note lacks a title). Once the note has been saved (e.g. when leaving the view by tapping a slashlink) it now _has_ an address. When the draft editor's `.appear` action fired it incorrectly assumed it had no address and attempts to recreate the from scratch. We need to check if the `state.address` exists and use _that_ instead of the stale `description.address`.

h/t @gordonbrander for telling me to check the description logic 🙏 

Simple fix, subtle issue.


gordonbrander (approved):
LGTM!

Issue #585, opened 2023-05-08
Back-button history view uses First Word at time of writing


jsantell:
Taking two notes that are linked, say "a" with first line "1" and note "b" with first line "2", and linked such that they can be navigated between either via backlink or link.

* From "a", link to "b"
* from "b", link to "a",
* From "a", change the first line from "1" to "3"
* link to "b"

The history view will display:
* `3`
* `2`
* `1`

This is a bit confusing. Could be more clear if they were just slug names/slashlinks e.g. `/foo`, `@\ben/foo`

PR #687, opened 2023-06-06
Fix stale navigation titles

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/585

Use the address for navigation titles, this is more scannable anyway and avoids the problem of loading content in the background just to update the title.

image


gordonbrander (approved):
LGTM!

Issue #649, opened 2023-05-16
Removing a sphere from followers still shows up


jsantell:
I followed `@bf`, and later, unfollowed (sorry Ben), but on refreshing/sync/later, `@bf` sometimes appears in my profile's "Following" tab, albeit without the person-check-icon, and the ellipses menu offering `"Follow @\bf"`, correctly identifying this sphere as one I'm not following, but as if this screen is populated from an outdated source

Issue #613, opened 2023-05-10
If the DID of a user appears in our address book, rebase their address


bfollington:
I follow `@alice` who follows `@bob`. `@bob` also follows `@alice`.
I browse to `@alice`'s profile, look at their following list and see `@bob.alice`
I browse to `@bob.alice`'s profile, look at their following list and see `@alice.bob.alice`.
If we tap on `@alice.bob.alice` this process loops infinitely.

Instead, we should detect that this is _our_ `@alice` and replace both their display name and their the navigation address with `@alice`.

This would benefit from https://github.com/subconsciousnetwork/noosphere/issues/358 but we can iterate over the address book cache for now.

Issue #611, opened 2023-05-10
Colorize @names based on type of identifier


bfollington:
From https://github.com/subconsciousnetwork/subconscious/issues/609:

> We have up to three names for a user:
> 1. our petname for them
> 2. their profile nickname
> 3. their relative address to us

These could be colorized by case to help with visual parsing:

1. petname: purple (accent)
2. nickname: pink/blue
3. address: grey

gordonbrander:
Capturing petname system design requirements in #615

Issue #610, opened 2023-05-09
Show "Also Known As" on profile header


bfollington:
Similar to https://github.com/subconsciousnetwork/subconscious/issues/609, we should investigate showing multiple names on the profile header. These could also be revealed on long press if it's too distracting.

gordonbrander:
Capturing petname design system requirements in #615

PR #688, opened 2023-06-06
Refine Petname UX

### LINKED ISSUE #615: Petname UX

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/649
Fixes https://github.com/subconsciousnetwork/subconscious/issues/613
Fixes https://github.com/subconsciousnetwork/subconscious/issues/610
Fixes https://github.com/subconsciousnetwork/subconscious/issues/611

Potentially resolves #615 

- [x] If we are following a user, we need access to the name we follow them under
  - use a enum with payload for this
- [x] if we follow a user, show our @petname on the first line and full path on the second
  - except in our address book where there's only one name
- [x] Show "Maybe: " in profile header context
- [x] Use leaf name on first line in .unknown case, italics, no @
- [x] Own profile name should count as `.known`

# Demo

https://github.com/subconsciousnetwork/subconscious/assets/5009316/f05efb6c-c017-4d48-a5f7-f0b14452efbb

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
@@ -86,7 +86,7 @@ struct PetnameView: View {
                  .proposedName(let address, let name):
                 Text(showMaybePrefix
                      ? "Maybe: \(name.description)"
-                     : "\(name.description)"
+                     : name.description
                 )
                 .italic()
                 .fontWeight(.medium)
--- a/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
@@ -53,10 +53,8 @@ struct PetnameView: View {
                 Text(name.toPetname().markup)
                     .fontWeight(.medium)
                     .foregroundColor(.accentColor)
-                // Ensure we do not show the exact same name twice
-                if let peer = address.peer,
-                   let petname = address.petname,
-                   petname.leaf != name {
+                
+                if let peer = address.peer {
                     PeerView(peer: peer)
                 }
             case .named(let address, let name):
--- a/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
@@ -7,13 +7,13 @@
 
 import SwiftUI
 
-public enum Name {
+public enum NameVariant {
     case known(Slashlink, Petname.Name)
     case named(Slashlink, Petname.Name)
     case unknown(Slashlink)
 }
 
-struct AddressView: View {
+struct PeerView: View {
     var peer: Peer
 
     var body: some View {
@@ -37,12 +37,12 @@ struct AddressView: View {
 
 /// Byline style for displaying a petname
 struct PetnameView: View {
-    var name: Name
+    var name: NameVariant
     var annotation: String? = nil
     
-    public func prefix(msg: String) -> Self {
+    public func annotated(with annotation: String) -> Self {
         var this = self
-        this.annotation = msg
+        this.annotation = annotation
         return this
     }
     
@@ -57,14 +57,14 @@ struct PetnameView: View {
                 if let peer = address.peer,
                    let petname = address.petname,
                    petname.leaf != name {
-                    AddressView(peer: peer)
+                    PeerView(peer: peer)
                 }
             case .named(let address, let name):
                 Text("\(annotation ?? "")\(name.description)")
                     .italic()
                     .fontWeight(.medium)
                 if let peer = address.peer {
-                    AddressView(peer: peer)
+                    PeerView(peer: peer)
                 }
             case .unknown(let address):
                 Text(address.peer?.markup ?? address.markup)
@@ -78,9 +78,9 @@ struct PetnameView: View {
 extension PetnameView {
     init(address: Slashlink, name: Petname.Name?) {
         if let name = name {
-            self.name = Name.named(address, name)
+            self.name = NameVariant.named(address, name)
         } else {
-            self.name = Name.unknown(address)
+            self.name = NameVariant.unknown(address)
         }
     }
 }
--- a/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
@@ -19,7 +19,7 @@ struct PeerView: View {
     var body: some View {
         HStack(spacing: AppTheme.unit) {
             ZStack {
-                RoundedRectangle(cornerRadius: AppTheme.cornerRadius/2)
+                RoundedRectangle(cornerRadius: AppTheme.cornerRadiusSm)
                     .foregroundColor(.secondaryBackground)
                 Text("AKA")
                     .foregroundColor(.secondary)
@@ -55,13 +55,7 @@ extension UserProfile {
 /// Byline style for displaying a petname
 struct PetnameView: View {
     var name: NameVariant
-    var annotation: String? = nil
-    
-    public func annotated(with annotation: String) -> Self {
-        var this = self
-        this.annotation = annotation
-        return this
-    }
+    var showMaybePrefix = false
     
     var body: some View {
         VStack(alignment: .leading, spacing: AppTheme.unit) {
@@ -83,17 +77,15 @@ struct PetnameView: View {
                    petname != name.toPetname() {
                     PeerView(peer: peer)
                 }
-            case .selfNickname(let address, let name):
-                Text("\(annotation ?? "")\(name.description)")
-                    .italic()
-                    .fontWeight(.medium)
-                if let peer = address.peer {
-                    PeerView(peer: peer)
-                }
-            case .proposedName(let address, let name):
-                Text("\(annotation ?? "")\(name.description)")
-                    .italic()
-                    .fontWeight(.medium)
+            case .selfNickname(let address, let name),
+                 .proposedName(let address, let name):
+                Text(showMaybePrefix
+                     ? "Maybe: \(name.description)"
+                     : "\(name.description)"
+                )
+                .italic()
+                .fontWeight(.medium)
+                
                 if let peer = address.peer {
                     PeerView(peer: peer)
                 }
@@ -116,7 +108,8 @@ struct PetnameView: View {
                     name: .proposedName(
                         Slashlink(profile: Petname("melville.bobby.tables")!),
                         Petname.Name("melville")!
-                    )
+                    ),
+                    showMaybePrefix: true
                 )
                 PetnameView(
                     name: .petname(
--- a/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
@@ -47,31 +47,30 @@ struct PetnameView: View {
     }
     
     var body: some View {
-        switch name {
-        case .known(let name, let address):
-            VStack(alignment: .leading, spacing: AppTheme.unit) {
+        VStack(alignment: .leading, spacing: AppTheme.unit) {
+            switch name {
+            case .known(let name, let address):
                 Text(name.toPetname().markup)
                     .fontWeight(.medium)
                     .foregroundColor(.accentColor)
+                // Ensure we do not show the exact same name twice
                 if let peer = address?.peer,
                    let petname = address?.petname,
                    petname.leaf != name {
                     AddressView(peer: peer)
                 }
-            }
-        case .named(let address, let name):
-                VStack(alignment: .leading, spacing: AppTheme.unit) {
-                    Text("\(annotation ?? "")\(name.description)")
-                        .italic()
-                        .fontWeight(.medium)
-                    if let peer = address.peer {
-                        AddressView(peer: peer)
-                    }
+            case .named(let address, let name):
+                Text("\(annotation ?? "")\(name.description)")
+                    .italic()
+                    .fontWeight(.medium)
+                if let peer = address.peer {
+                    AddressView(peer: peer)
                 }
-        case .unknown(let address):
+            case .unknown(let address):
                 Text(address.peer?.markup ?? address.markup)
                     .lineLimit(1)
                     .fontWeight(.medium)
+            }
         }
     }
 }
--- a/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
@@ -8,9 +8,9 @@
 import SwiftUI
 
 public enum NameVariant {
-    case known(Slashlink, Petname.Name)
-    case named(Slashlink, Petname.Name)
-    case unknown(Slashlink)
+    case petname(Slashlink, Petname.Name)
+    case proposedName(Slashlink, Petname.Name)
+    case selfNickname(Slashlink, Petname.Name)
 }
 
 struct PeerView: View {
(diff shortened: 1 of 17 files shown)


xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,104 @@
 
 import SwiftUI
 
+public enum Name {
+    case known(Slashlink, Petname.Name)
+    case named(Slashlink, Petname.Name)
+    case unknown(Slashlink)
+}
+
+struct AddressView: View {
```
bfollington, line 16: `PeerView`?

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,104 @@
 
 import SwiftUI
 
+public enum Name {
```
bfollington, line 10: `NameVariant`? `TrustLevel`?

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,104 @@
...
+}
+
 /// Byline style for displaying a petname
 struct PetnameView: View {
-    var address: Slashlink
-    var name: Petname?
+    var name: Name
+    var annotation: String? = nil
+    
+    public func prefix(msg: String) -> Self {
```
bfollington, line 43: `annotated(_ with: String)`

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -104,9 +104,6 @@ enum UserProfileDetailAction {
     case editProfileSheet(EditProfileSheetAction)
     case followNewUserFormSheet(FollowNewUserFormSheetAction)
     
-    case fetchFollowingStatus(Did)
-    case populateFollowingStatus(Bool)
```
bfollington, line 108: Redundant, we can just refresh the profile

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -144,20 +141,30 @@ enum UserCategory: Equatable, Codable, Hashable, CaseIterable {
 struct UserProfile: Equatable, Codable, Hashable {
     let did: Did
     let nickname: Petname.Name?
-    let address: Slashlink
+    var address: Slashlink
```
bfollington, line 144: We need to be able to change this field to achieve https://github.com/subconsciousnetwork/subconscious/issues/613

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -341,14 +346,11 @@ struct UserProfileDetailModel: ModelProtocol {
             model.statistics = content.statistics
             model.recentEntries = content.recentEntries
             model.following = content.following
-            model.isFollowingUser = content.isFollowingUser
             model.loadingState = .loaded
             
             return update(
```
bfollington, line 351: Replace with regular update

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -501,16 +482,26 @@ struct UserProfileDetailModel: ModelProtocol {
             )
             
         case .attemptUnfollow:
-            guard let did = state.unfollowCandidate?.did else {
+            guard let candidate = state.unfollowCandidate else {
                 return Update(state: state)
             }
             
+            let x = Func.run {
```
bfollington, line 489: name this

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -501,16 +482,26 @@ struct UserProfileDetailModel: ModelProtocol {
...
+                switch (candidate.ourFollowStatus) {
+                case .following(let name):
+                    return environment.addressBook
+                        .unfollowUserPublisher(
+                            did: candidate.did,
+                            petname: name
```
bfollington, line 495: Not sure this makes sense. We want to unfollow the user by checking if the leaf name of their address is the same as one of our petnames for them and unfollow _that_ name, preserving any others.

If the leaf name and our petname do not match, then unfollow by DID instead pursing all names.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -9,6 +9,22 @@ import os
 import Foundation
 import Combine
 
+enum UserProfileFollowStatus: Codable, Hashable, Equatable {
```
bfollington, line 12: name?

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,111 @@
...
+    case unknown(Slashlink)
+}
+
+struct PeerView: View {
+    var peer: Peer
+
+    var body: some View {
+        HStack(spacing: AppTheme.unit) {
+            ZStack {
+                RoundedRectangle(cornerRadius: AppTheme.cornerRadius/2)
```
gordonbrander, line 22: Nit: perhaps this should be a constant `AppTheme.cornerRadiusSm`.

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,111 @@
 
 import SwiftUI
 
+public enum NameVariant {
+    case known(Slashlink, Petname.Name)
+    case named(Slashlink, Petname.Name)
+    case unknown(Slashlink)
```
gordonbrander, line 13: Might consider type names that map directly to concepts from one of the petname papers? https://spritely.institute/static/papers/petnames.html

- `petname` (following directly)
- `selfNickname` or `autonym` or `selfProposedName` (the nickname a sphere proposes for itself in its header)
- `nickname` or `proposedName` (The nickname a 3p proposes for another sphere. This is the petname that sphere gives another sphere.)

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,111 @@
...
+}
+
 /// Byline style for displaying a petname
 struct PetnameView: View {
-    var address: Slashlink
-    var name: Petname?
+    var name: NameVariant
+    var annotation: String? = nil
+    
+    public func annotated(with annotation: String) -> Self {
```
gordonbrander, line 43: For localization reasons, we want to express this as a string interpolation via `LocalizedStringKey`.

Given it's an interpolation, I don't think we can to express this as a generalized `annotated` method. Rather, we probably want to do the interpolation close to, or at the call site, probably in `PetnameView` body.

```swift
Text("Maybe: \(name)")
```

https://developer.apple.com/documentation/swiftui/localizedstringkey/init(stringinterpolation:)

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,111 @@
...
+                Text("\(annotation ?? "")\(name.description)")
+                    .italic()
+                    .fontWeight(.medium)
+                if let peer = address.peer {
+                    PeerView(peer: peer)
+                }
+            case .unknown(let address):
+                Text(address.peer?.markup ?? address.markup)
+                    .lineLimit(1)
+                    .fontWeight(.medium)
```
gordonbrander, line 79: I think we want to treat this case (other-proposed name) the same as the self-proposed name case above.

```
Maybe: bob
AKA @bob.alice
```

We don't have a self-proposed name from bob, so we use the name that alice proposes instead. This is a valid fallback.

In the case the address does not have a `Peer.petname`, we could fall back on just showing the first N characters of the did.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileHeaderView.swift
```diff
@@ -27,32 +26,46 @@ struct UserProfileHeaderView: View {
...
+                    
+                case (.following(let name), _):
+                    PetnameView(name: .known(user.address, name))
+                    
+                case _:
+                    PetnameView(
+                        address: user.address,
+                        name: user.nickname
+                    )
+                    .annotated(with: "Maybe: ")
```
gordonbrander, line 44: Noted above. I think we need to use interpolation for localization reasons.

xcode/Subconscious/Shared/Models/Slashlink.swift
```diff
@@ -118,7 +118,7 @@ public struct Slashlink:
     }
 
     /// Convenience initializer that creates a link to `@user/_profile_`
-    init(petname: Petname) {
```
gordonbrander, line 121: Nit: I think I would personally prefer to keep this arg name unchanged. Reason: `petname:` mirrors the `Peer.petname` naming.

gordonbrander (changes requested):
A few small questions + comments

gordonbrander:
I think we can say this fixes https://github.com/subconsciousnetwork/subconscious/issues/615. If we decide we need further refinement, we can always re-open.

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
bfollington: Yes, this is better. I dropped the `unknown` case reworked the logic, introducing `UserProfile.toNameVariant()` which can hold all the logic about this mapping.

xcode/Subconscious/Shared/Models/Slashlink.swift
bfollington: I see this argument. I felt it was slightly confusing for the `Slashlink(petname:)` constructor to implicitly return a profile link despite saying nothing about profiles in the name, because of the more common `Slashlink(petname:slug:)`.

Happy to roll it back regardless.

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -7,54 +7,122 @@
...
+}
+
+extension UserProfile {
+    func toNameVariant() -> NameVariant? {
+        switch (
+            self.category,
+            self.ourFollowStatus,
+            self.nickname,
+            self.address.petname?.leaf
+        ) {
```
bfollington, line 23: Couldn't find a specific recommendation in the google swift styleguide for formatting long switch tuples like this, so this is a best guess.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -279,6 +279,24 @@ extension NotebookAction {
...
+                    return ourUser
+                case (_, .following(let name)):
+                    var ourUser = user
+                    // Rewrite address using our name
+                    ourUser.address = Slashlink(petname: name.toPetname())
+                    return ourUser
+                case _:
+                    return user
+                }
+            }
```
bfollington, line 297: https://github.com/subconsciousnetwork/subconscious/issues/613

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
gordonbrander: Works for me

gordonbrander (approved):
LGTM!

Issue #690, opened 2023-06-06
Tapping `@name` in a memo viewer / editor can lead to `Not Found` state


bfollington:
Currently, if you type an `@name` in a note that is _not_ in your address book you can still attempt to tap it. The app will try to navigate to the address but fail to load the content.

We could:
1. Prevent navigation to an unresolvable name
    - May require a difference in syntax highlighting to make it clear why the tap isn't working?
2. Improve the not-found state

bfollington:
After chatting to @gordonbrander we're leaning towards 2. in the short term.

We can design an nicer empty state with a quote, similar to the empty notes screen. Some possible directions:

> “Wanderer, your footsteps are the road, and nothing more; wanderer, there is no road, the road is made by walking. By walking one makes the road, and upon glancing behind one sees the path that never will be trod again. Wanderer, there is no road-- Only wakes upon the sea.

> "The true nature of phenomena is emptiness"

PR #694, opened 2023-06-09
Improve "Not Found" state

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/690

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/NotFoundView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/NotFoundView.swift
@@ -1,0 +1,44 @@
+//
+//  NotFoundView.swift
+//  Subconscious
+//
+//  Created by Ben Follington on 9/6/2023.
+//
+
+import SwiftUI
+
+public struct NotFoundView: View {
+    public var body: some View {
+        VStack(spacing: AppTheme.unit * 6) {
+            Spacer()
+            Text("Not Found")
+            Image(systemName: "moon.stars.fill")
+                .font(.system(size: 64))
+            VStack(spacing: AppTheme.unit) {
+                Text("""
+                     When emptiness is possible,
+                     Everything is possible;
+                     Were emptiness impossible,
+                     Nothing would be possible.
+                     """)
+                .italic()
+                Text(
+                    "Nāgārjuna"
+                )
+            }
+            .frame(maxWidth: 240)
+            // Some extra padding to visually center the group.
+            // The icon is large and rather heavy. This offset
+            // helps prevent the illusion of being off-center.
+            .padding(.bottom, AppTheme.unit * 24)
+            .font(.caption)
+            Spacer()
+        }
+        .multilineTextAlignment(.center)
+        .padding()
+        .frame(maxWidth: .infinity)
+        .foregroundColor(Color.secondary)
+        .background(Color.background)
+        .transition(.opacity)
+    }
+}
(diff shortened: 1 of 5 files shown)


xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -320,6 +321,7 @@ struct UserProfileDetailModel: ModelProtocol {
             model.initialTabIndex = initialTabIndex
             // We might be passed some basic profile data
             // we can use this in the loading state for a preview
+            model.address = address
```
bfollington, line 324: Ensures that even if we fail to load the user's profile we can still render the omnibus with the intended navigation address.

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -41,10 +41,7 @@ struct MemoViewerDetailView: View {
                     notify: notify
                 )
             case .notFound:
-                MemoViewerDetailNotFoundView(
-                    backlinks: store.state.backlinks,
```
gordonbrander, line 45: I think we want to include backlinks with this view (no dead ends). That was the original intent of passing them in, but I don't know if it was actually ever implemented.

Non-blocking, but we should file a bug to track.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -261,6 +261,7 @@ struct UserProfileDetailModel: ModelProtocol {
     var isUnfollowConfirmationPresented = false
     var isEditProfileSheetPresented = false
     
+    var address: Slashlink? = nil
```
gordonbrander, line 264: Address is already part of `UserProfile`. Should we look for it there instead?

Keeping properties in sync with update pattern is no problem, but I try to avoid it unless needed.

gordonbrander (approved):
Approved! One question.

xcode/Subconscious/Shared/Components/Common/NotFoundView.swift
```diff
@@ -0,0 +1,44 @@
...
+            Spacer()
+            Text("Not Found")
+            Image(systemName: "moon.stars.fill")
+                .font(.system(size: 64))
+            VStack(spacing: AppTheme.unit) {
+                Text("""
+                     When emptiness is possible,
+                     Everything is possible;
+                     Were emptiness impossible,
+                     Nothing would be possible.
```
gordonbrander, line 22: ✨

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
bfollington: This is to deal with the fact that (in a "Not Found" state) we do not have access to any user's profile so we need to track the address we attempted to find separately to let us render it.

I am thinking of this in terms of: `address` is where you are trying to navigate, `user` is only present if `address` resolves. Further, the `address` could be `@alice.bob.charlie` _but_ when resolving we might realise we know `@alice` already, so the logical value for `user.address` is `@alice` _not_ `@alice.bob.charlie`.

gordonbrander: That totally makes sense, and the current solution WFM.

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
bfollington: Wasn't sure on this one but I've restored the backlinks view. 

image

Since the only way to reach the memo viewer empty state is via 3P content we don't see any backlinks yet, pending #323.

Issue #608, opened 2023-05-09
Re-assigning a petname still returns old DID


bfollington:
It seems as if a petname returns the _first_ DID it was ever assigned, if it is current assigned one? This may be a result of syncing with the gateway, it reminds me of #569.

Reproduction Steps:

1. Follow a user with petname `@test` (`did:key:123`)
2. Unfollow them
3. Follow a new user with petname `@test` (`did:key:456`)
4. Find that `@test` resolves to `did:key:123` NOT `did:key:456`

bfollington:
Found during playtesting.

Noosphere logs show that immediately after assigning a new identity to the name `jordan` it returns the old identity, it's possible this is a noosphere issue.

```DEBUG noosphere_sphere::petname::write: Adopting 'jordan' (did:key:z6MkpDvpHue5XMimBKjd8xZ2GkaKCrSPwihncUaBnPs62SgU)```

cdata:
Repro'd, this is a Noosphere bug

PR #695, opened 2023-06-09
Update to noosphere@0.11.1

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/608


cdata:
LGTM, just bear in mind that we need to re-deploy gateways as well.

cdata (approved):

Issue #677, opened 2023-06-01
Default gateway settings trigger an error state


bfollington:
If a user does not have a gateway configured then the settings view then the default state displays an error:

<img width="366" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/a4928842-7826-4c15-86ca-b50afa8b1a20">

This is confusing for a new user who has no idea what a gateway even is. Instead, we should replace the gateway section with the provisioning status when there is no URL configured:

<img width="359" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/6147d0c6-d4ee-4aa4-aecb-cc0fed4a395a">

These two sections can be mutually exclusive for the first beta. Eventually users will want to set their own gateway URLs without provisioning first but... let's worry about that when it's an issue.

PR #696, opened 2023-06-09
Refactor Gateway Settings View(s)

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/677

This PR simplifies the experience of dealing with the gateway. If you do not have a gateway configured we show the provision section (with invite code input). If you _do_ have a gateway configured we show the sync section.

image

image


gordonbrander (changes requested):
We should preserve the ability to plug in a URL to a self-hosted gateway.

We could merge this and file a follow-up to re-implement. However, I thought I would mention here, since the PR removes the code for saving a custom gateway URL.

bfollington:
@gordonbrander hmmm good point. I've been playing around for a bit and landed on this:

image

image

Which doesn't remove the initial error message from failing to sync with the default gateway, but it puts it one level deeper in the menu so you'll only encounter it if you go looking. You can't edit it while gateway provisioning is in progress, so on the expected onboarding journey you'd avoid seeing the error state.

gordonbrander (approved):
LGTM

Issue #658, opened 2023-05-22
Track state of background jobs in the model


bfollington:
Currently anything using `Func.retryWithBackoff` has some hidden state within the background job (how many attempts have passed etc.) which cannot be accessed from the model that dispatched the job. These processes are fire-and-forget right now but that approach is unlikely to scale.

A better solution might be to create a generic `BackgroundProcessModel` similar to `FormFieldModel` that can be used like so:

```swift
struct UserProfileDetailModel: ModelProtocol {
    // ...
    let pendingFollowProcess = BackgroundProcessModel(maxAttempts: 10)
}
```

```swift
enum BackgroundProcessAction<T> {
    case .start(fx: Fx<T>)
    case .retry(attempt: int)
    case .cancel
    case .succeed
    case .fail
}
```

```swift
// Consuming model
case .startMyJob:
    let fx: Fx<UserProfileDetailAction> = 
        environment.addressBook
              .waitForPetnameResolutionPublisher(petname: petname)
              .map { _ in
                  .succeedResolveFollowedUser
              }
              .recover { error in
                  .failResolveFollowedUser(error.localizedDescription)
              }
              .eraseToAnyPublisher()

    return update(
        state: state,
        action: .pendingFollowProcess(.start(fx)),
        environment: environment
    )
```

This is just a sketch and the best solution may not take this exact form. This issue can be closed when we have control and visibility of all background jobs from the model that initiates them.

Issue #618, opened 2023-05-10
Toast notifications for confirmation of background actions


bfollington:
There are a number of cul de sacs in the app right now where after a key operation there is no feedback to the user. This is particularly noticeable during:

1. Note Rename
2. Follow/Unfolow
3. Manual Sync

We need some kind of notification / toast / snackbar system to present information to the user:

![image](https://github.com/subconsciousnetwork/subconscious/assets/5009316/570bb776-abb2-4483-83be-87460bd174b5)

This gives us a useful hook for undo and will be important to surface sync conflicts when multiple clients are involved.

bfollington:
From discussion with @gordonbrander:

- Each major view owns responsibility for rendering its snackbar overlay
    - App view has a default renderer, but will be obscured by sheets etc.
- We should have a root "queue" of notifications which can be filtered and conditionally rendered by other views (when and if needed)

bfollington:
Spent some time exploring possible designs for this. I'm unsure whether we will need or want support for displaying an action button in our notifications.

# Concept A

As simple as possible. Straight out of the Google playbook, where it would cover the FAB the snackbar is moved up according to: https://m2.material.io/components/snackbars#usage

<img src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/5d81b86f-cd96-4d0a-97d6-20d21d7f9b5f" alt="Frame 1" width="320"> <img src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/af223c79-78e3-4d96-978f-3e9b8a8ce06e" alt="Frame 2" width="320">

# Concept B

Echoing the pattern of the omnibar. This restricts the amount of space we have and it's hard to fit an action (like "Undo" into the space. 

The omnibar _could_ become the "home" of the UI showing location, loading status (by animating the border etc.) and notifications. You can imagine notifications popping in with a nice animation, it has elements of the [Dynamic Island](https://support.apple.com/en-au/guide/iphone/iph28f50d10d/ios).

<img src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/5f1aa224-2660-4266-8b56-4a8f05579f03" alt="Frame 3" width="320"> <img src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/f7bd68b9-91f8-4fed-a614-20fd6f57a9ea" alt="Frame 4" width="320">

# Concept C

[...]

PR #704, opened 2023-06-12
Notification UX

bfollington:
#618 
#658 

- Basic views


bfollington:
Note to self, for background processes (sync, gateway provisioning, migration of dat) #658 I think Live Activities and ActivityKit are a good fit. In-app notifications are still needed to supplement live activities though, and can be used to communicate status updates _for_ those activities when the app has focus.

Issue #708, opened 2023-06-13
nickname field should not capitalize


gordonbrander:
Autocomplete suggestion I got was capitalized.

- We should use the decorator for that field that does not autocapitalize.
- We should also try to lowercase-ify the text.

Issue #674, opened 2023-06-01
Profile memo is updated every time nickname is loaded


bfollington:
We hooked into the existing `persistNickname` action to write the profile memo, but this is actually called on every keystroke while the nickname is updated and causes a new version to be created on every sync operation.

We should clarify the meaning of `nicknamePersisted` vs `nicknameUpdated` (for example) and only save the memo when the field is _submitted_. For onboarding this is leaving the step, for the settings this is defocusing the field?

Additional note: perhaps it's actually fine to write to the profile memo on every keystroke, but a sync should not cause an update to the profile memo.

---

This seems like it _should_ be fine, if a little wasteful, but actually causes a data corruption issue when syncing with the gateway. I suspect this is a race condition between the new version created via the profile edit and any new versions on the gateway (from NS resolution, for example).

PR #719, opened 2023-06-14
Fix formfield cursor behaviour

### LINKED ISSUE #718: Cursor jumps to end of textfield
Has superficial similarity to https://github.com/subconsciousnetwork/subconscious/issues/667.

In some text fields adding or removing a character in the middle of the string will jump the cursor to the end. This is most obvious when editing a long textfield such as a DID input, gateway URL or the profile bio field. 

It seems like this is due to mutating the value of the text field (perhaps via validation) and blowing away the selection context as a result? We need a simplified example to reproduce it.

Specifically, we should compare @State / ObservableStore against TextField / ValidatedTextField. I doubt that all TextFields are broken on iOS so something must be wrong in one of the 2x2 quadrants here.

bfollington:
Fixes #718 
Fixes https://github.com/subconsciousnetwork/subconscious/issues/708
Fixes https://github.com/subconsciousnetwork/subconscious/issues/674

# Changes

- `ValidatedTextField` has an internal `@State var` that holds the current field value
    - Edits are replayed to the store to keep the two in sync
- New `ValidatedFormField` component which wraps `ValidatedTextField` wires up everything to talk to a backing `FormField<>`
- Replaced all fields by `FormField`s
- Nickname can be entered with capitals but will always persist and present in lowercase
- Bio will automatically truncate >280 chars
- Clarified profile memo lifecycle

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -457,14 +457,6 @@ struct AppModel: ModelProtocol {
         value: "",
         validate: { value in Petname.Name(value) }
     )
-    /// Expose read-only value for view
-    var nicknameFormFieldValue: String {
-        nicknameFormField.value
-    }
-    /// Expose read-only valid value for view
-    var isNicknameFormFieldValid: Bool {
-        nicknameFormField.isValid
-    }
     
     var inviteCodeFormField = InviteCodeFormField(
         value: "",
@@ -498,8 +490,6 @@ struct AppModel: ModelProtocol {
     /// stored in `AppDefaults`.
     var gatewayURL = ""
     var gatewayId: String? = nil
-//    var gatewayURLTextField = ""
-//    var isGatewayURLTextFieldValid = true
     var gatewayURLField = GatewayUrlFormField(
         value: "",
         validate: { value in
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -104,7 +104,7 @@ enum AppAction: CustomLogStringConvertible {
     /// Set sphere/user nickname
     /// Sets form field, and persists if needed.
     case setNickname(_ nickname: String)
-    case persistNickname(_ nickname: String)
+    case persistNickname
     
     /// Write to `Slashlink.ourProfile` during onboarding
     case updateOurProfileWithNickname(_ nickname: Petname.Name)
@@ -607,11 +607,10 @@ struct AppModel: ModelProtocol {
                 environment: environment,
                 text: nickname
             )
-        case let .persistNickname(nickname):
+        case .persistNickname:
             return persistNickname(
                 state: state,
-                environment: environment,
-                text: nickname
+                environment: environment
             )
         case let .updateOurProfileWithNickname(nickname):
             return updateOurProfileWithNickname(
@@ -1015,13 +1014,13 @@ struct AppModel: ModelProtocol {
         environment: AppEnvironment,
         text: String
     ) -> Update<AppModel> {
+//        let text = text.lowercased().trimmingCharacters(in: .whitespaces)
         /// First pass down setValue to form field,
         /// then persist the nickname by reading the updated model.
         return update(
             state: state,
             actions: [
                 .nicknameFormField(.setValue(input: text)),
-                .persistNickname(text)
             ],
             environment: environment
         )
@@ -1029,11 +1028,10 @@ struct AppModel: ModelProtocol {
     
     static func persistNickname(
         state: AppModel,
-        environment: AppEnvironment,
-        text: String
+        environment: AppEnvironment
     ) -> Update<AppModel> {
         // Persist any valid value
-        if let validated = Petname.Name(text) {
+        if let validated = Petname.Name(state.nickname) {
             var model = state
             model.nickname = validated.description
             logger.log("Nickname saved: \(validated)")
@@ -1308,7 +1306,11 @@ struct AppModel: ModelProtocol {
             model.firstRunPath = []
         }
         
-        return Update(state: model).animation(.default)
+        return update(
+            state: model,
+            action: .persistNickname,
+            environment: environment
+        ).animation(.default)
     }
     
     /// Reset NoosphereService managed instances of `Noosphere` and `Sphere`.
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -106,7 +106,7 @@ enum AppAction: CustomLogStringConvertible {
     /// Set sphere/user nickname
     /// Sets form field, and persists if needed.
     case setNickname(_ nickname: String)
-    case persistNickname
+    case persistNickname(_ nickname: Petname.Name)
     
     /// Write to `Slashlink.ourProfile` during onboarding
     case updateOurProfileWithNickname(_ nickname: Petname.Name)
@@ -637,10 +637,11 @@ struct AppModel: ModelProtocol {
                 environment: environment,
                 text: nickname
             )
-        case .persistNickname:
+        case .persistNickname(let nickname):
             return persistNickname(
                 state: state,
-                environment: environment
+                environment: environment,
+                nickname: nickname
             )
         case let .updateOurProfileWithNickname(nickname):
             return updateOurProfileWithNickname(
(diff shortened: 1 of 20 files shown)


xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -7,17 +7,57 @@
...
 
 /// A text field that comes with help text and a validation flag
 struct ValidatedTextField: View {
+    @State var innerText: String = ""
+    
     var alignment: HorizontalAlignment = .leading
     var placeholder: String
     @Binding var text: String
     var onFocusChanged: (Bool) -> Void = { _ in}
+    var onTextChanged: () -> Void = {}
```
bfollington, line 56: Drop this

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -104,7 +106,7 @@ enum AppAction: CustomLogStringConvertible {
     /// Set sphere/user nickname
     /// Sets form field, and persists if needed.
     case setNickname(_ nickname: String)
-    case persistNickname(_ nickname: String)
+    case submitNickname(_ nickname: Petname.Name)
```
bfollington, line 109: Updated to more clearly reflect how it's being used, mirrors the way GatewayURL is managed

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -929,32 +955,29 @@ struct AppModel: ModelProtocol {
...
-            logger.log("Nickname saved: \(validated)")
-            
-            return update(
-                state: model,
-                action: .updateOurProfileWithNickname(validated),
-                environment: environment
-            )
-        }
+        var model = state
+        model.nickname = nickname.description
```
bfollington, line 970: We store the lowercase version of the nickname

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -7,17 +7,56 @@
...
+    
     var alignment: HorizontalAlignment = .leading
     var placeholder: String
     @Binding var text: String
     var onFocusChanged: (Bool) -> Void = { _ in}
     var caption: String
-    var hasError: Bool = false
     var axis: Axis = .horizontal
     var autoFocus: Bool = false
+    var isValid: Bool = true
```
bfollington, line 59: I changed this back to `isValid` and renamed the concept of `hasError` to `shouldPresentAsInvalid` in `FormField`.

xcode/Subconscious/Shared/Models/FormField.swift
```diff
@@ -27,8 +27,8 @@ struct FormField<Input: Equatable, Output>: ModelProtocol {
     static func == (lhs: FormField<Input, Output>, rhs: FormField<Input, Output>) -> Bool {
         return (
             lhs.value == rhs.value &&
+            lhs.defaultValue == rhs.defaultValue &&
             lhs.touched == rhs.touched &&
-            lhs.isValid == rhs.isValid &&
```
bfollington, line 31: No point checking a derived property for equality.

xcode/Subconscious/SubconsciousTests/Tests_App.swift
```diff
@@ -66,8 +62,8 @@ final class Tests_App: XCTestCase {
         )
         XCTAssertEqual(b.state.nicknameFormField.value, "valid", "Nickname form field is set")
         XCTAssertEqual(b.state.nicknameFormField.isValid, true, "Nickname form field is valid")
-        XCTAssertEqual(b.state.nicknameFormFieldValue, "valid", "Nickname form field value set")
-        XCTAssertEqual(b.state.isNicknameFormFieldValid, true, "Nickname form field is valid")
-        XCTAssertEqual(b.state.nickname, "valid", "Nickname is saved, because form field value is valid")
+        
+        let c = AppModel.update(state: model, action: .submitNickname(Petname.Name("valid")!), environment: AppEnvironment())
+        XCTAssertEqual(c.state.nickname, "valid", "Nickname is saved after submitted valid nickname.")
```
bfollington, line 67: Updated to capture new behaviour: `state.nickname` won't be update until `submitNickname` is called with a valid name.

xcode/Subconscious/Shared/Components/Settings/ProfileSettingsView.swift
```diff
@@ -10,23 +10,27 @@ import ObservableStore
...
-                    ),
-                    caption: "Lowercase letters, numbers, and dashes only",
-                    hasError: !app.state.isNicknameFormFieldValid
+                    field: app.state.nicknameFormField,
+                    send: app.send,
+                    tag: AppAction.nicknameFormField,
+                    caption: "Lowercase letters, numbers and dashes only."
                 )
-                .formField()
+                .textInputAutocapitalization(.never)
```
gordonbrander, line 24: 🥳

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -7,17 +7,56 @@
 
 import SwiftUI
 import Combine
+import ObservableStore
+
+struct ValidatedFormField<T: Equatable, Action>: View {
```
gordonbrander, line 12: This wrapping view feels strange to me. Can we do it with one view?

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -7,17 +7,56 @@
...
+            submitLabel: submitLabel,
+            onSubmit: onSubmit
+        )
+        .formField()
+    }
+}
 
 /// A text field that comes with help text and a validation flag
 struct ValidatedTextField: View {
+    @State var innerText: String = ""
```
gordonbrander, line 50: State should be `private`

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -7,17 +7,56 @@
...
 import SwiftUI
 import Combine
+import ObservableStore
+
+struct ValidatedFormField<T: Equatable, Action>: View {
+    var alignment: HorizontalAlignment = .leading
+    var placeholder: String
+    var field: FormField<String, T>
+    var send: (Action) -> Void
+    var tag: (FormFieldAction<String>) -> Action
```
gordonbrander, line 17: I would like to avoid passing down tagging functions, and instead pass down a tagged `send` function. The tagging function is an implementation detail the child shouldn't know about, and we don't want the child to be able to pass up anything except its own actions.

Eventually this tagging pattern will be replaced by `StoreProtocol.viewStore` helper https://github.com/subconsciousnetwork/ObservableStore/blob/main/Sources/ObservableStore/ObservableStore.swift#LL347C17-L347C26

gordonbrander:
It would also be worth investigating why this happens for this component, but not other input fields that are driven by ObservableStore. 

One noted difference is the closure in this model. I wonder if it might be confusing SwiftUI's change detection under the hood.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
bfollington: Yep, I was preserving the path to using the prior API but you're right, I can just flatten this out since we always use this new form.

bfollington:
> It would also be worth investigating why this happens for this component, but not other input fields that are driven by ObservableStore.
> 
> One noted difference is the closure in this model. I wonder if it might be confusing SwiftUI's change detection under the hood.

I think there was a little bit of a red herring here, thought I'm still not 100% sure. Based on my testing here is my understanding:

I _think_ that the fundamental cause of the cursor issues is a race condition between updating the binding via the store and internal input events of the TextField. It seemed like `FormField`s were experience _more_ cursor issues but I saw (and have seen) the same issues crop up in other fields. I recall the first report I heard of this problem was the Gateway URL field, which did not use the `FormField` abstraction until this PR.

After decoupling the input binding from the store the cursor issues resolved, but I was seeing stale validation results instead. To fix that I introduced the `ValidatedFormField` API that takes the entire `FormField` struct as a property (as opposed to the current value and a computed `hasError` property). I suspected that the change detection would be more reliable checking struct equality vs. a computed field and it appears to behave correctly now, though I'll admit I am hand waving.

(Sidenote: sheets made no difference to anything)

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
bfollington: I'm not sure why I blanked on the fact we already have `Address.forward`. I switched everything to use that instead.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -7,21 +7,21 @@
 
 import SwiftUI
 import Combine
+import ObservableStore
 
-/// A text field that comes with help text and a validation flag
-struct ValidatedTextField: View {
+struct ValidatedFormField<T: Equatable>: View {
+    @State private var innerText: String = ""
+    @FocusState private var focused: Bool
```
gordonbrander, line 14: I like the way we've wrapped this up. It's unfortunate that we have to replay changes, but the way we're doing it makes it an invisible implementation detail of the view. 👍

gordonbrander (approved):
LGTM!

Thank you for this. Form input and focus bugs are often one of the trickiest aspects of UI frameworks.

PR #728, opened 2023-06-16
Improve handling of long petnames

### LINKED ISSUE #680: Truncate petname paths from the middle
> As a user, I want to see the proposed name of the user I'm viewing, and the petname of the root of the petname path, so I sense know *who* I know this account through.

We should truncate petnames from the middle, so that `@eve.dave.carol.bob.alice` becomes `@eve...alice` or `@eve.dave...alice`, etc.

The root petname is, in a sense, the root of trust. This makes the immediate name and the root name the most important pieces of information.

Screenshot from app with current truncation examples:

### LINKED ISSUE #713: Header layout can become cramped / wrap when names get too long
image

Consider moving the follow button down below so the name can reach across the screen. We will still need to truncate names in some cases, see #680

bfollington:
Fixes #680 
Fixes #713 

image

# Changes
- Truncate petname in the middle
- Reduce size of profile action button

# Questions
- Do we want to shrink the buttons? We can probably get away with truncation but I feel like the buttons were a little large

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Buttons/PillButtonStyle.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Buttons/PillButtonStyle.swift
@@ -39,10 +39,11 @@ struct PillButtonView: View {
                 case .regular:
                     return AppTheme.unit * 12
                 case .small:
-                    return AppTheme.unit * 6
+                    return AppTheme.unit * 8
                 }
             }
         )
+        
     }
 }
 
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Components/Common/Buttons/PillButtonStyle.swift
```diff
@@ -39,7 +39,7 @@ struct PillButtonView: View {
                 case .regular:
                     return AppTheme.unit * 12
                 case .small:
-                    return AppTheme.unit * 9
+                    return AppTheme.unit * 6
```
gordonbrander, line 42: Minimum hit target size is 44pt. The standard iOS border button style is visually 33pt high, but, testing, I can see it has generous invisible padding at the top and bottom to bring the hit target to something around 44pt.

Would suggest we stay around 32pt high for visual consistency with the rest of the OS (I often round to the nearest number divisible by 8 or 4), and add invisible overlay or padding to bring to 44pt.

gordonbrander (changes requested):
One note on button size. Otherwise, LGTM.

xcode/Subconscious/Shared/Components/Common/Buttons/PillButtonStyle.swift
bfollington: Good call, I've updated the visual height to 32pt (AppTheme.unit * 8) and given the button itself a minimum tap target height of 44pt.

image

Issue #699, opened 2023-06-09
Sync with gateway when pull-to-refresh on profiles


gordonbrander:
We sync with the gateway in the background when pulling-to-refresh on the notes list. We should also sync with the gateway in the background when pulling-to-refresh on our profile, or on someone else's profile.

"I want to refresh" always means "sync with gateway" given the way our protocol works, so I think we want to sync with gateway anywhere we can pull-to-refresh.

Issue #684, opened 2023-06-05
Button to retry resolving peer


gordonbrander:
> As a user, I want to be able to retry following after a failure, so I can recover from an error condition caused by bugs, service outages, or loss of connectivity.

Ran into this while trying to resolve some dids. The dids are/were valid, but once you get a failure, there's no obvious way to retry. We could add a button to retry for profiles that fail to resolve.

<img src="https://github.com/subconsciousnetwork/subconscious/assets/58421/d0daea07-8c4c-44f9-944b-3a5b013229bd" width="400">

bfollington:
The gateway will continue trying to resolve these in the background, so retrying resolution is effectively the same as syncing with the gateway. This means we can offer the affordance to retry as an alternative way to sync and refresh. 

Should the retry button trigger a background waiting process again, or just check once?

PR #732, opened 2023-06-19
Profile refresh & resolution

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/684
Fixes https://github.com/subconsciousnetwork/subconscious/issues/699

# Changes

- Changed "unresolved" icon to be less prominent (screenshot below)
- Introduced "Refresh" option in context menu
- Sync with gateway when doing pull-to-refresh on profile
- Slightly increase wait duration for resolution

image

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
@@ -43,6 +43,7 @@ struct StoryUserView: View {
     var action: (Slashlink) -> Void
     
     var profileAction: (UserProfile, UserProfileAction) -> Void = { _, _ in }
+    var onRefreshUser: () -> Void = {}
     
     var body: some View {
         VStack(alignment: .leading, spacing: 0) {
@@ -105,7 +106,23 @@ struct StoryUserView: View {
                                 }
                             )
                         }
-                       
+                        
+                        switch story.user.resolutionStatus {
+                        case .unresolved:
+                            Button(
+                                action: {
+                                    onRefreshUser()
+                                },
+                                label: {
+                                    Label(
+                                        title: { Text("Refresh") },
+                                        icon: { Image(systemName: "arrow.clockwise") }
+                                    )
+                                }
+                            )
+                        case _:
+                            EmptyView()
+                        }
                     },
                     label: {
                         Image(systemName: "ellipsis")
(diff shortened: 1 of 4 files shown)


xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -53,7 +53,10 @@ struct UserProfileDetailView: View {
             onNavigateToNote: self.onNavigateToNote,
             onNavigateToUser: self.onNavigateToUser,
             onProfileAction: self.onProfileAction,
-            onRefresh: store.refresh
+            onRefresh: {
+                app.send(.syncAll)
```
gordonbrander, line 57: 👍

xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -355,7 +355,7 @@ actor AddressBookService {
     func waitForPetnameResolution(
         petname: Petname
     ) async throws -> Cid? {
-        let maxAttempts = 10 // 1+2+4+8+16+32+32+32+32+32 = 191 seconds
+        let maxAttempts = 13 // 1+2+4+8+16+32+32+32+32+32+32+32+32 = 287 seconds
```
gordonbrander, line 358: 🙏

xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
```diff
@@ -105,7 +106,23 @@ struct StoryUserView: View {
...
+                        
+                        switch story.user.resolutionStatus {
+                        case .unresolved:
+                            Button(
+                                action: {
+                                    onRefreshUser()
+                                },
+                                label: {
+                                    Label(
+                                        title: { Text("Refresh") },
```
gordonbrander, line 118: My gut is that an explicit refresh control feels like exposing too much implementation detail to the user.

Could it be possible to invoke this automatically for any unresolved profile, or tie it into following list pull-to-refresh?

Non-blocking. We can file a follow-up to address.

bfollington: Tying it to pull-to-refresh is a good idea, I'll experiment with that before merging to investigate.

bfollington: As discussed, attempting resolution again on-tap feels way better.


https://github.com/subconsciousnetwork/subconscious/assets/5009316/f4e69a8d-d6f5-4e28-8a3e-b9dea34884cd

Issue #715, opened 2023-06-13
Onboarding will still proceed with an invalid invite code


bfollington:
We currently validate that the invite code is in the expected format (4 words) **but** if cloudctl rejects that invite code we are still letting the user proceed.

We need to wait until the code is actually accepted and exchanged for a gateway ID before letting users move, or else they have to choose the "Use Offline" flow.

Issue #697, opened 2023-06-09
Increase backoff factor, or number of polls for gateway provisioning


gordonbrander:
I frequently get to 5- or 6-of-10 polls before the gateway is ready. This feels too close for comfort to me.

We should also give user ability to retry.

PR #735, opened 2023-06-20
Rework Invite Code Experience

### LINKED ISSUE #737: Confusing CTA in settings screen: "Look for this in your welcome email"
When going through gateway provisioning after completing the first run experience, if I go to settings I will see something like this:

image

The invite code is displayed with a subtitle, "Look for this in your welcome email."

This sounds like a call to action for the user, but is probably intended to be a placeholder. It might be good to use something descriptive in this case such as, "Your invite code."

### LINKED ISSUE #763: QR code not populating the DID after scan
We have a regression. When scanning a QR code the DID is received and logged but the input field is not populated. 

We should add a test to assert this behaviour.

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/715
Fixes https://github.com/subconsciousnetwork/subconscious/issues/697
Fixes https://github.com/subconsciousnetwork/subconscious/issues/738
Fixes #763 
Fixes #737 

# Demo

I can't record a fresh demo right now since the cluster is down, I recommend playtesting the app. But here's a slightly stale video of the new settings layout. All the gateway settings are together now, it's maybe a bit confronting for users but they have full control to pick-up the process at any point it failed.

https://github.com/subconsciousnetwork/subconscious/assets/5009316/415004db-17dc-43fb-959e-08686e519bfb

# First Run

image

# Settings

image

Rework the invite code flow to ensure the code is actually valid. See screenshots in comments throughout.

- [x] Add latest graphic here + in docs
- [ ] Attach videos & screenshots to aid review
- [x] Introduce tests for first run step submission/validation
- [x] Test new FormField functionality
- [x] Tweak invite redemption failure message
- [x] Rework `isValid` abstraction
- [x] `.pushStep` for FirstRun


xcode/Subconscious/Shared/Components/FirstRun/FirstRunDoneView.swift
```diff
@@ -12,7 +12,11 @@ struct FirstRunDoneView: View {
     @Environment(\.colorScheme) var colorScheme
     
     private var status: ResourceStatus {
-        app.state.gatewayProvisioningStatus
+        if app.state.lastGatewaySyncStatus == .succeeded {
+            return app.state.lastGatewaySyncStatus
+        }
```
bfollington, line 17: Edge case: if you end up on this screen after relaunching the app and have sync'd with the gateway then we should say you're connected.

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
```diff
@@ -46,26 +46,62 @@ struct FirstRunView: View {
...
+                            Text(gatewayId)
+                        }
+                        .font(.caption)
                     }
-                )
-                .textFieldStyle(.roundedBorder)
-                .textInputAutocapitalization(.never)
-                .disableAutocorrection(true)
-                
+                    .foregroundColor(.secondary)
```
bfollington, line 68: On un-focusing the field we will attempt to redeem the invite code and, if successful, we replace the text field entirely by a message:

image

This is a quick pass visually, open to feedback / refinement.

bfollington: Here's the error state:

image

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
```diff
@@ -46,26 +46,62 @@ struct FirstRunView: View {
...
+                            send: app.send,
+                            tag: AppAction.inviteCodeFormField
+                        ),
+                        caption: Func.run {
+                            switch app.state.gatewayProvisioningStatus {
+                            case .failed(_):
+                                return "Failed to redeem invite code"
+                            case _:
+                                return "Look for this in your welcome email"
+                            }
```
bfollington, line 84: Display error state as a validation error. This meant expanding the API of `ValidatedFormField` but I also think it's the best way to handle this error?

image

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
```diff
@@ -77,24 +113,27 @@ struct FirstRunView: View {
                         }
                     )
                     .buttonStyle(PillButtonStyle())
-                    .disabled(!app.state.inviteCodeFormField.isValid)
+                    .disabled(app.state.gatewayId == nil)
```
bfollington, line 116: You cannot proceed without a gateway ID, unless going via the "Use Offline" path.

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
```diff
@@ -124,6 +163,13 @@ struct FirstRunView: View {
                     FirstRunDoneView(app: app)
                 }
             }
+            .onAppear {
+                guard app.state.sphereIdentity == nil else {
+                    return
+                }
+                
+                app.send(.createSphere)
+            }
```
bfollington, line 172: Create the sphere earlier so we can kick off provisioning as soon as the invite code is accepted.

xcode/Subconscious/Shared/Components/Settings/SettingsView.swift
```diff
@@ -51,21 +51,31 @@ struct GatewayProvisioningSection: View {
...
-                .autocorrectionDisabled(true)
-                .onDisappear {
-                    app.send(.setInviteCode(app.state.inviteCodeFormField.value))
+                if let gatewayId = app.state.gatewayId {
+                    LabeledContent(
+                        "Gateway ID",
+                        value: gatewayId
+                    )
+                    .lineLimit(1)
+                    .textSelection(.enabled)
```
bfollington, line 60: If we have already exchanged an invite code for a gateway ID, hide the invite code. This means that once you've redeemed one invite code you'll have to delete the app before you can redeem another, but it affords a cleaner UI for the more common case of having just one code.

gordonbrander:
The graphic in the issue is great.

We could consider creating a `docs` dir at the top level and dropping it in.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -22,6 +22,10 @@ struct ValidatedFormField<T: Equatable>: View {
     var autoFocus: Bool = false
     var submitLabel: SubmitLabel = .done
     var onSubmit: () -> Void = {}
+    var onFocusChanged: (_ focused: Bool) -> Void = { _ in }
+    
+    /// Indicate if this field should appear invalid due to a side effect failing etc.
+    var isValid: Bool = true
```
gordonbrander, line 28: Does this toggle on/off the validation feature?

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
gordonbrander: WFM

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
gordonbrander: Small tweak request: "Could not redeem invite code".

We use "succeed" / "fail" under the hood, mirroring the terminology used for the Result type. However, I think "fail" is a scary word to present to users.

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
gordonbrander: If proceeding without an accepted invite is an actual logical failure case, we may need to add a guard to the update function as well, ensuring an invite code has been accepted before pushing another panel onto the stack. The update function should be self-standing, in the sense that it should not be able to produce an invalid next state from an action.

The UI is blocking the production of this action, and the user should never actually encounter this case, we could just guard, no-op, and log.

xcode/Subconscious/Shared/Components/Settings/SettingsView.swift
```diff
@@ -51,21 +51,31 @@ struct GatewayProvisioningSection: View {
...
+                    )
+                    .autocapitalization(.none)
+                    .autocorrectionDisabled(true)
+                    .onDisappear {
+                        app.send(.setInviteCode(app.state.inviteCodeFormField.value))
+                    }
+                    .disabled(app.state.gatewayProvisioningStatus == .pending)
                 }
-                .disabled(app.state.gatewayProvisioningStatus == .pending)
+
```
gordonbrander, line 78: Nit: Stray 4 spaces.

xcode/Subconscious/Subconscious.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/twostraws/CodeScanner",
       "state" : {
-        "revision" : "4b2f9c57d03bfe94e6b45572e0c908c1433c32b2",
-        "version" : "2.3.2"
+        "revision" : "d8f10d992ef32a4187b78a6df320facf75d631d8",
+        "version" : "2.3.1"
```
gordonbrander, line 9: 🤔 as long as we're touching this, let's lock in this dependency at a specific revision, so XCode stops munging it.

gordonbrander (approved):
A few comments. Otherwise, LGTM!

gordonbrander (changes requested):
We should allow changing the invite code, at least in the settings, and probably in the FRE? If you put in the wrong invite code in this PR, there would be no recovery path.

I recently attempted to re-use invite code during a playtest, with a predictable failure. However, without a way to write in a new invite code, I would have been "bricked" https://github.com/subconsciousnetwork/subconscious/issues/738

bfollington:
Based on feedback here and the work we're doing in https://github.com/subconsciousnetwork/subconscious-cloud/issues/6 I think we should rethink the overall flow here a little.

We can address https://github.com/subconsciousnetwork/subconscious/issues/737 and https://github.com/subconsciousnetwork/subconscious/issues/738 easily once https://github.com/subconsciousnetwork/subconscious-cloud/issues/6 lands.

bfollington:
Doing some modelling to explore how things would work after https://github.com/subconsciousnetwork/subconscious-cloud/issues/6 lands.

@gordonbrander let me know what you think, especially about my callout here RE: trying to redeem a code multiple times.

image
https://www.figma.com/file/jLgkHEJgiXICUK4y3UmcU6/Subconscious-Whiteboard?type=whiteboard&node-id=27-1056&t=KZxLhKs7zUCCntdO-4

jsantell:
Clarifying the current planned changes in the redemption, mostly compatible with the flow above:

* Invite codes are bound to the provided sphere on redemption.
* Provisioning with the same invite code/sphere pair is idempotent.
* Provisioning *request* can fail via an internal error, or if invite code is invalid (not legit, or claimed by another sphere, we do not specify). The error `"type"` and its `"error"` message are available in JSON for non-200 responses.
* The provisioning request returns gateway ID regardless of state (whether provisioned or provisioning). Status must be queried after to handle whether pending/ready.
* If provisioning *task* fails, the provisioning request can be retried.

bfollington:
> Clarifying the current planned changes in the redemption, mostly compatible with the flow above:
> 
> * Invite codes are bound to the provided sphere on redemption.
> * Provisioning with the same invite code/sphere pair is idempotent.
> * Provisioning _request_ can fail via an internal error, or if invite code is invalid (not legit, or claimed by another sphere, we do not specify). The error `"type"` and its `"error"` message are available in JSON for non-200 responses.
> * The provisioning request returns gateway ID regardless of state (whether provisioned or provisioning). Status must be queried after to handle whether pending/ready.
> * If provisioning _task_ fails, the provisioning request can be retried.

Amazing! Thanks @jsantell, I had a quick read over https://github.com/subconsciousnetwork/subconscious-cloud/pull/7/files#diff-e1e65eeceb23dd47013779351844a41a2b781e987c92db2bb0d4fa60d51703b9 and updated the diagram with a few more details. 

image

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
bfollington: Yep, agreed, right now this is driven by NavigationLinks but we need a `.pushStep` action for the FirstRun experience.

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
bfollington: No, the name might be misleading... but I might just be approaching this the wrong way.

I wanted to mark a field as invalid due to an an error in a side effect... but as I write this I think I could just inline my logic into the regular validation function. Or display a seperate error panel.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -854,34 +866,39 @@ struct AppModel: ModelProtocol {
...
+                environment: environment,
+                gatewayId: gatewayId
+            )
+        case .failRedeemInviteCode(let error):
+            return failRedeemInviteCode(
+                state: state,
+                environment: environment,
+                error: error
+            )
+        case .submitProvisionGatewayForm:
```
bfollington, line 898: Drop this?

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
```diff
@@ -54,16 +51,22 @@ struct ValidatedFormField<T: Equatable>: View {
...
+                .onChange(of: innerText) { innerText in
                     send(.setValue(input: innerText))
-                })
-                .submitLabel(submitLabel)
+                }
+                .onChange(of: field) { field in
+                    // The has been reset, sync inner value
+                    if !field.touched && innerText != field.value {
+                        innerText = field.value
+                    }
```
bfollington, line 68: This is also the cause of the QR code scanning issue, attach it

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
```diff
@@ -124,6 +166,9 @@ struct FirstRunView: View {
                     FirstRunDoneView(app: app)
                 }
             }
+            .onAppear {
+                app.send(.createSphere)
+            }
```
bfollington, line 171: I don't like this, there must be a better place?

I made this action idempotent in the `update` function so maybe we just call it at app startup?

xcode/Subconscious/Shared/Components/FirstRun/FirstRunProfileView.swift
```diff
@@ -65,16 +62,6 @@ struct FirstRunProfileView: View {
             AppTheme.onboarding
                 .appBackgroundGradient(colorScheme)
         )
-        .onAppear {
-            app.send(.createSphere)
-        }
```
bfollington, line 70: Moved to the `.initial` step

xcode/Subconscious/Shared/Components/FirstRun/FirstRunProfileView.swift
```diff
@@ -65,16 +62,6 @@ struct FirstRunProfileView: View {
...
-        .onDisappear {
-            guard let nickname = app.state.nicknameFormField.validated else {
-                return
-            }
-            
-            app.send(.submitNickname(nickname))
-        }
```
bfollington, line 77: Moved into the model as part of the "next step" logic.

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
```diff
@@ -46,55 +72,69 @@ struct FirstRunView: View {
...
+                        send: Address.forward(
+                            send: app.send,
+                            tag: AppAction.inviteCodeFormField
+                        ),
+                        caption: Func.run {
+                            switch app.state.inviteCodeRedemptionStatus {
+                            case .failed(_):
+                                return "Could not redeem invite code"
+                            case _:
+                                return "Your ticket to the Noosphere"
```
bfollington, line 93: I tweaked this to address #737 but open to changing to anything else

xcode/Subconscious/Shared/Models/FormField.swift
```diff
@@ -38,6 +40,7 @@ struct FormField<Input: Equatable, Output>: ModelProtocol {
     var defaultValue: Input
     /// Should be a pure, static function
     var validate: FormFieldValidator<Input, Output>
+    var isValid: Bool = false
```
bfollington, line 43: By making this a field, instead of derived, we can mark a field as invalid via a store action.

This is used to present the "Could not redeem invite code" error on a field.

xcode/Subconscious/SubconsciousTests/Tests_FirstRun.swift
```diff
@@ -0,0 +1,209 @@
...
+            up6.state.firstRunPath,
+            [.nickname],
+            "clear invite code + gateway ID when requesting offline"
+        )
+        XCTAssertNil(up6.state.inviteCode)
+        XCTAssertNil(up6.state.gatewayId)
+        
+        // We should CANCEL the provisioniong process when offline mode is requested
+        // ...but we can't do that until we add support for store-driven cancellation
+        // XCTAssertTrue(up6.state.gatewayProvisioningStatus == .initial)
```
bfollington, line 137: We can test this properly after #658 lands

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
gordonbrander: I thought that a note for where to look was a good idea for the FRE in particular.

Perhaps "You can find your invite code in your welcome email"?

xcode/Subconscious/Shared/Models/FormField.swift
gordonbrander: Making it an actual data field also allows us to trigger a re-render on change of validation.

gordonbrander: One question this raises for future is if it is worth having the validation lambda be part of the struct.

As an alternative, the parent update function could do whatever validation is needed and manually toggle the `isValid` field of the form field. This would move the business logic of validation out of the form field component, and into the specific parent view component. May allow us to side-step the generic in the form field model.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1849,32 +1951,32 @@ struct AppModel: ModelProtocol {
...
     
-    static func requestProvisionGateway(
+    static func requestRedeemInviteCode(
         state: AppModel,
         environment: AppEnvironment,
         inviteCode: InviteCode
     ) -> Update<AppModel> {
         guard let did = state.sphereIdentity,
               let did = Did(did) else {
+            logger.error("Missing identity, cannot redeem invite code")
```
gordonbrander, line 1953: Is there a recovery path for this case?

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
gordonbrander: Approach LGTM. Would playtest a bit to make sure no infinite loops are created, but I've had to do two-way sync code like this before in SwiftUI to work around similar issues, and this looks correct.

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
```diff
@@ -114,6 +154,8 @@ struct FirstRunView: View {
                 for: FirstRunStep.self
             ) { step in
                 switch step {
+                case .initial:
+                    EmptyView()
```
gordonbrander, line 158: Seems odd. We never want to hit this case, right?

xcode/Subconscious/Shared/Components/Settings/GatewayURLSettingsView.swift
```diff
@@ -44,25 +44,54 @@ struct ResourceSyncBadge: View {
...
+        switch status {
+        case .failed:
+            return .red
+        default:
+            return .accentColor
+        }
+    }
+
+    var body: some View {
+        Label(title: {
```
gordonbrander, line 73: Minor nit: odd formatting. Prefer

```
Label(
    title: {
    },
    icon: {
    }
)
```

xcode/Subconscious/Shared/Components/Settings/GatewayURLSettingsView.swift
```diff
@@ -88,6 +117,129 @@ struct GatewayProvisionLabel: View {
     }
 }
 
+// MARK: Provisioning
+
+struct ValidatedInviteCodeFormField: View {
```
gordonbrander, line 122: We could consider breaking these section views out into separate files. Non-blocking.

gordonbrander (approved):
LGTM. Haven't test-driven, but nothing looks unusual or out-of-place to me. Good improvements!

xcode/Subconscious/Shared/Models/FormField.swift
bfollington: Agreed, I have been deferring a change like this until we inevitably need validation that can produce different messages (instead of binary true/false) but I don't like carrying the closure around like this. 

We might also be able to leverage a protocol to make validation cleaner?

xcode/Subconscious/Shared/Components/Common/Forms/ValidatedTextField.swift
bfollington: I didn't get it right the first time haha, my first _few_ attempts resulted in infinite loops 😅

This seems robust from my testing so far, but we'll see...

xcode/Subconscious/Shared/Models/FormField.swift
bfollington: Added an issue to track this https://github.com/subconsciousnetwork/subconscious/issues/779

xcode/Subconscious/Shared/Components/AppView.swift
bfollington: We should only get here if `.createSphere` has failed or wasn't run in the first place. 

We _could_ dispatch `.createSphere` here as an escape hatch and then retry... but we might end up in an infinite loop. I've slightly tweaked it to surface a UI error and try to create the sphere (once) if this happens.

xcode/Subconscious/Shared/Components/Settings/GatewayURLSettingsView.swift
bfollington: Good catch, this was an Xcode autocomplete I didn't reformat.

xcode/Subconscious/Shared/Components/Settings/GatewayURLSettingsView.swift
bfollington: Done!

xcode/Subconscious/Shared/Components/FirstRun/FirstRunView.swift
bfollington: On reflection, this was basically a laziness decision. I've remodelled the actions to remove the idea of `.initial` step.

PR #740, opened 2023-06-20
Report uncaught exceptions with sentry.io

justinabrahms:
We still need to upload debug symbols for release builds.

PR #742, opened 2023-06-20
Decorate Sentry errors with log stream

bfollington:
Extends #740 

Use `OSLogEnumerator` to pin the app logs to the reported error. This probably isn't super efficient but given the nature of reporting a crash, that's probably ok.


bfollington:
I should be appending to event.crumbs not setting it, I could be deleting existing info

justinabrahms (approved):

Issue #693, opened 2023-06-08
`Read corrupted data at file offset None backtrace ()`


bfollington:
A mysterious error that spontaneously occurred during development. Appears that somehow the on-disk data has been corrupted. After profiling disk use in Instruments and examining the files on disk it seems like the app can read data as expected,  but the data itself is malformed.

Full dump of the application data is attached, along with the (mostly unhelpful) logs:

[dump.zip](https://github.com/subconsciousnetwork/subconscious/files/11683453/dump.zip)

```
2023-06-08 12:24:55.034032+1000 Subconscious[79520:12172927] [NoosphereService] init NoosphereService
2023-06-08 12:24:55.034122+1000 Subconscious[79520:12172927] [NoosphereService] Global storage URL: file:///Users/ben/Library/Developer/CoreSimulator/Devices/01619CFE-B226-41AB-9391-7B85DA80F25E/data/Containers/Data/Application/4F973352-825A-4D6F-90DF-243925B389A6/Library/Application%20Support/noosphere
2023-06-08 12:24:55.034185+1000 Subconscious[79520:12172927] [NoosphereService] Sphere storage URL: file:///Users/ben/Library/Developer/CoreSimulator/Devices/01619CFE-B226-41AB-9391-7B85DA80F25E/data/Containers/Data/Application/4F973352-825A-4D6F-90DF-243925B389A6/Library/Application%20Support/sphere
2023-06-08 12:24:55.034235+1000 Subconscious[79520:12172927] [NoosphereService] Gateway URL: https://bdyqnyx47jvtoggasueyhqgoz7dirnl3.sphere.test.subconscious.cloud/
2023-06-08 12:24:55.034267+1000 Subconscious[79520:12172927] [NoosphereService] Sphere identity: did:key:z6MkemLNbFXrN6vUgQJtDUdiow2BLfSHbwtP7nNVNrNpKnEE
2023-06-08 12:24:55.037745+1000 Subconscious[79520:12172927] [app] Set sphere ID: did:key:z6MkemLNbFXrN6vUgQJtDUdiow2BLfSHbwtP7nNVNrNpKnEE
2023-06-08 12:24:55.109682+1000 Subconscious[79520:12172927] [app] Documents: file:///Users/ben/Library/Developer/CoreSimulator/Devices/01619CFE-B226-41AB-9391-7B85DA80F25E/data/Containers/Data/Application/4F973352-825A-4D6F-90DF-243925B389A6/Documents/
[...]

PR #755, opened 2023-06-26
Make Gateway URL update idempotent

bfollington:
Related to https://github.com/subconsciousnetwork/subconscious/issues/693

# Changes
- Clarify gateway URL lifecycle
- Make updates to gateway URL idempotent
- Prevent editing key values during sync

Unsure if this technically "fixes" the issue, since we still reset our noosphere instance as-soon-as there's a reason to update the gateway URL. Perhaps we should be queuing the reset operation (somehow) until after in-flight work is complete? Or perhaps we can find a way to update the gateway URL without tearing down and rebuilding noosphere?

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -1105,6 +1105,7 @@ struct AppModel: ModelProtocol {
         url: URL
     ) -> Update<AppModel> {
         guard state.gatewayURL != url.absoluteString else {
+            logger.log("Gateway URL is identical to current value, doing nothing")
             return Update(state: state)
         }
         
@@ -1135,6 +1136,7 @@ struct AppModel: ModelProtocol {
         environment: AppEnvironment
     ) -> Update<AppModel> {
         guard let url = state.gatewayURLField.validated else {
+            logger.log("Gateway URL field is invalid, doing nothing")
             return Update(state: state)
         }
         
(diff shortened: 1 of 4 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1093,6 +1104,10 @@ struct AppModel: ModelProtocol {
         environment: AppEnvironment,
         url: URL
     ) -> Update<AppModel> {
+        guard state.gatewayURL != url.absoluteString else {
```
gordonbrander, line 1107: Should log that we got a duplicate URL and are doing nothing.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1115,6 +1130,21 @@ struct AppModel: ModelProtocol {
         .mergeFx(fx)
     }
     
+    static func submitGatewayURLForm(
+        state: AppModel,
+        environment: AppEnvironment
+    ) -> Update<AppModel> {
+        guard let url = state.gatewayURLField.validated else {
+            return Update(state: state)
```
gordonbrander, line 1138: Worth adding a log that URL is invalid, so doing nothing

xcode/Subconscious/Shared/Components/Settings/GatewayURLSettingsView.swift
```diff
@@ -109,26 +109,26 @@ struct GatewayURLSettingsView: View {
...
-                                )
-                            }
-                        )
-                        .disabled(app.state.gatewayURL.count == 0)
-                    }
+                    Button(
+                        action: {
+                            // Ensure URL is updated before trying to sync
+                            app.send(.submitGatewayURLForm)
+                            app.send(.syncSphereWithGateway)
```
gordonbrander, line 120: Nit: would prefer sending a single action which dispatched these two actions via `actions:`

gordonbrander (approved):
A few minor suggestions. Otherwise LGTM.

PR #756, opened 2023-06-27
Fix loading bar animation

### LINKED ISSUE #709: Loading indicator transposed outside of the header bar
Occasionally when loading a sphere, the loading indicator appears (the border-stripe moving around the header bar) and is transposed such that it no longer aligns with the header bar

bfollington:
Fixes #709 

It really seems like switching from `.onAppear` to `.task` is enough. I could speculate as to why, but the short answer must be a race condition between layout and animation that is resolved by the time a `.task` runs.

I can trigger the issue occasionally by spamming navigation with `.onAppear` but not with `.task`.


gordonbrander (approved):
✨

Issue #726, opened 2023-06-16
"Random Note" suggestion can attempt to edit 3p note


bfollington:
Found playtesting #627, we're always presenting the editor view for 3p notes. This could be fixed by either excluding 3p notes from the candidate set or by updating the presentation logic to use the correct view.

Issue #725, opened 2023-06-16
"Random Note" suggestion returns address with DID peer


bfollington:
As of #627 selecting "Random Note" from the search-or-create menu will direct you to an unresolved peer address with a DID at the start instead of a petname.

PR #757, opened 2023-06-27
Fix "Random Note" suggestion

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/726
Fixes https://github.com/subconsciousnetwork/subconscious/issues/725

- Tweak loading animation trigger
- Fix random retrieval functions

I had to touch some of the old geist code but I just got it compiling and left them for now.

### DIFF
--- a/xcode/Subconscious/Shared/Services/DatabaseService.swift
+++ b/xcode/Subconscious/Shared/Services/DatabaseService.swift
@@ -1108,18 +1108,22 @@ final class DatabaseService {
         guard self.state == .ready else {
             return nil
         }
-
+        
+        let dids = [
+            Did.local.description,
+            owner.description]
+        
         return try? database.execute(
             sql: """
             SELECT id, title
             FROM memo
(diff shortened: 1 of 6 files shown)


xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1054,13 +1071,21 @@ final class DatabaseService {
             SELECT id, title
             FROM memo
             WHERE substr(memo.slug, 1, 1) != '_'
+                AND did = ?
```
bfollington, line 1074: Restrict random selection to only `our` notes.

gordonbrander: Note this doesn't include draft content, which has a "fake" local data did.

In other SQL queries, we pass in, and then unpack a JSON array to match against any DID in the array.

gordonbrander (approved):
Approve! One refinement requested.

Thanks for fixing!

PR #758, opened 2023-06-27
Disable Sentry logging while running tests

bfollington:
Introduced a `TEST` environment variable only set when the `SubconsciousTests` scheme runs.


gordonbrander (approved):
Approved, but flagging @justinabrahms in case I'm missing anything.

justinabrahms (commented):
Looks fine to me, but I don't really know if/when that config is loaded. As long as release builds don't set the var, I'm happy.

Issue #749, opened 2023-06-22
`MemoEditorDetail` EXC_BAD_ACCESS


bfollington:
```
Thread 1: EXC_BAD_ACCESS (code=1, address=0x10)

Thread 1 Queue : com.apple.main-thread (serial)
#0	0x000000018bcc721c in Array._getCount() ()
#1	0x000000010431009c in MemoEditorDetailAction.logDescription.getter at /Users/ben/code/subconscious/xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift:446
#2	0x00000001043171b0 in protocol witness for CustomLogStringConvertible.logDescription.getter in conformance MemoEditorDetailAction ()
#3	0x00000001046d3f8c in static String.loggable<τ_0_0>(_:) at /Users/ben/code/subconscious/xcode/Subconscious/Shared/Library/CustomLogStringConvertible.swift:21
#4	0x0000000104309fb0 in closure #7 in MemoEditorDetailView.body.getter at /Users/ben/code/subconscious/xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift:190
#5	0x000000010bbad534 in ___lldb_unnamed_symbol119067 ()
#6	0x000000010c0ac1b0 in ___lldb_unnamed_symbol159348 ()
#7	0x000000010be153c0 in ___lldb_unnamed_symbol139732 ()
#8	0x000000010be14e78 in ___lldb_unnamed_symbol139731 ()
#9	0x000000010bbad474 in ___lldb_unnamed_symbol119066 ()
#10	0x000000010bbae71c in ___lldb_unnamed_symbol119123 ()
#11	0x000000010bbad9d8 in ___lldb_unnamed_symbol119080 ()
#12	0x000000010bbad808 in ___lldb_unnamed_symbol119079 ()
#13	0x000000010bbada40 in ___lldb_unnamed_symbol119082 ()
#14	0x000000010c41b0c4 in ___lldb_unnamed_symbol186293 ()
#15	0x000000019b3490d4 in PassthroughSubject.Conduit.offer(_:) ()
#16	0x000000019b34a1fc in partial apply for closure #1 in PassthroughSubject.send(_:) ()
#17	0x000000019b38070c in partial apply for thunk for @callee_guaranteed (@guaranteed ConduitBase<τ_0_0, τ_0_1>) -> (@error @owned Error) ()
#18	0x000000018bd68588 in Sequence.forEach(_:) ()
[...]

PR #761, opened 2023-06-27
Fix memory santization error

### LINKED ISSUE #759: `error.pointee` sometimes has incorrect value
I believe that I've accidentally discovered a memory bug in the app. I stumbled across these diagnostics in Xcode today (under Scheme Settings) and thought I'd try them out. Docs are here for the different options: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/ManagingMemory/Articles/MallocDebug.html

Long story short: the app crashes under `Address Sanitization` and/or `Malloc Scribble` right here: https://github.com/subconsciousnetwork/subconscious/blob/2bc417c1a0838e2585a1475813f56de2d49dc8ff/xcode/Subconscious/Shared/Services/Noosphere/Noosphere.swift#L255

Thanks to the using `MallocPreScribble` and `MallocScribble` I can see that (sometimes?) `error.pointee` is a pointer to a garbage address. `MallocPreScribble` writes `0xAA` to any newly allocated bytes and my value for `error.pointee` is `.some(0xaaaaaaaaaaaaaaaa)`.

This _seems_ like we're allocating the space for the pointer but not actually writing to it on the NS side?

bfollington:
Fixes #759 

Discovered using the `Address Sanitizer` diagnostic in Xcode, which I've left enabled in Debug.

We must explicitly zero-out the `pointee` field after this unsafe allocation, otherwise it could literally be pointing _anywhere_.

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/Noosphere.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/Noosphere.swift
@@ -244,6 +244,9 @@ public actor Noosphere {
         _ perform: (UnsafeMutablePointer<OpaquePointer?>) -> Z
     ) throws -> Z {
         let error = UnsafeMutablePointer<OpaquePointer?>.allocate(capacity: 1)
+        // Explicitly clear this memory before use
+        error.pointee = nil
+        
         defer {
             error.deallocate()
         }
--- a/xcode/Subconscious/Subconscious.xcodeproj/xcshareddata/xcschemes/Subconscious (iOS).xcscheme
+++ b/xcode/Subconscious/Subconscious.xcodeproj/xcshareddata/xcschemes/Subconscious (iOS).xcscheme
@@ -26,7 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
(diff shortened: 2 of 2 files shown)


xcode/Subconscious/Shared/Services/Noosphere/Noosphere.swift
```diff
@@ -244,6 +244,9 @@ public actor Noosphere {
         _ perform: (UnsafeMutablePointer<OpaquePointer?>) -> Z
     ) throws -> Z {
         let error = UnsafeMutablePointer<OpaquePointer?>.allocate(capacity: 1)
+        // Explicitly clear this memory before use
+        error.pointee = nil
```
gordonbrander, line 248: This seems wild! Do we not overwrite the pointee when setting the error pointer on the Rust side?

I looked for more Swift resources explaining this in more detail, but couldn't find any. If you have them, please link.

gordonbrander (approved):
Approve with question

xcode/Subconscious/Shared/Services/Noosphere/Noosphere.swift
jsantell: In lieu of other (less idiomatic) solutions, explicitly initializing the error pointer as `nil` is necessary to determine whether or not a noosphere method set an error, otherwise checking `error.pointee` would dereference as neither an error object nor nil and throw a BAD_ACCESS

bfollington:
Possible that this will fix https://github.com/subconsciousnetwork/subconscious/issues/749

Issue #723, opened 2023-06-16
After migration my notes the notebook view was empty, but my notes were present on my profile


bfollington:
Per the title, after migrating my notes with #627 my notebook view was empty, but my notes were visible on my profile.
After creating a new note the notes rendered in the main list, but the count was still out of date until I pulled-to-refresh.

I suspect if I had just refreshed initially it would also have fixed the view, we must not be refreshing at the completion of indexing.

gordonbrander:
Tested migration from 0.0.7 by installing 0.0.7, then flashing with `main`.

Findings: local notes were migrated correctly.

gordonbrander:
Wild guess without any info: these were sphere notes, and when #627 was flashed, Noosphere was off. Thus, the gateway sync was not kicked off, and after it, the sphere sync was not kicked off. Since there were no local drafts, the app showed an empty state.

@bfollington does this sound plausible?

bfollington:
@gordonbrander yep that definitely sounds plausible, I'm not sure WHY the Noosphere setting turned off but I recall  having to turn it back on post-migration.

I have a faint memory of seeing the issue again later in testing but I can't recall if Noosphere was off the second time.

bfollington:
Possible duplicate of https://github.com/subconsciousnetwork/subconscious/issues/659

PR #777, opened 2023-06-28
Make notes empty state pull-to-refreshable

### LINKED ISSUE #599: Pull-to-refresh empty note state
We don't currently let you pull-to-refresh empty note state to sync, but we should.

gordonbrander:
Fixes #599

- Create group for EntryList views
- Break empty state out into separate view
- Wrap empty state in scrollview, implement refreshable

Creates an escape hatch, at least, if users run into https://github.com/subconsciousnetwork/subconscious/issues/723

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/EntryList/EntryListEmptyView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/EntryList/EntryListEmptyView.swift
@@ -1,0 +1,59 @@
+//
+//  EntryListEmptyView.swift
+//  Subconscious (iOS)
+//
+//  Created by Gordon Brander on 6/28/23.
+//
+
+import SwiftUI
+
+struct EntryListEmptyView: View {
+    var onRefresh: () -> Void
+
+    var body: some View {
+        GeometryReader { geom in
+            ScrollView {
+                VStack {
+                    Spacer()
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Components/Common/EntryList/EntryListEmptyView.swift
```diff
@@ -0,0 +1,59 @@
...
+                            .frame(width: 240)
+                            .font(.caption)
+                        }
+                        Spacer()
+                    }
+                    Spacer()
+                }
+                .multilineTextAlignment(.center)
+                .padding()
+                .frame(minHeight: geom.size.height)
```
bfollington, line 42: Assuming this is to ensure the user can drag from anywhere on the screen?

bfollington (approved):
One curious question, but LGTM!

xcode/Subconscious/Shared/Components/Common/EntryList/EntryListEmptyView.swift
gordonbrander: By default, views in scrollviews shrink vertically to a height defined by their contents.

We set an explicit height for the view so that the quote within will be vertically centered.

Issue #769, opened 2023-06-28
You can accidentally overwrite a petname when manually entering DID


bfollington:
Currently, you can overwrite an entry in your address book accidentally using the "Add User Form" (located in Profile -> Top Right Corner)

We should detect and prevent this collision the same way we do when following a user from their profile: by offering a `-1`, `-2`, `-3` etc. suffixed version.

PR #780, opened 2023-07-03
Prevent overwriting petnames via Follow User form

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/769

# Demo

https://github.com/subconsciousnetwork/subconscious/assets/5009316/74cbe177-07a9-4763-acf0-63d7d770e383

# Questions

- Is it better to:
    1. mark collision as a validation error and force the user to pick a new name
    2. replace the name with an available suffixed version _in the form_
    3. dismiss the form and find a unique name in the background, adding a suffix if needed

I tried (1) first and liked it but currently we have (2). (3) seems like a hard sell without #333.

# Changes
- Prevent overwriting petnames
- Surface overwrite petname error to UI
- Suggest a unique name if there is a collision

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Profile/FollowUserFormSheet.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/FollowUserFormSheet.swift
@@ -264,7 +264,7 @@ struct FollowUserFormView: View {
                         send: send,
                         tag: FollowUserFormAction.petnameField
                     ),
-                    caption: "Lowercase letters, numbers and dashes only."
+                    caption: state.failFollowMessage ?? "Lowercase letters, numbers and dashes only."
                 )
                 .lineLimit(1)
                 .textInputAutocapitalization(.never)
@@ -279,6 +279,7 @@ struct FollowUserFormView: View {
 enum FollowUserFormAction: Equatable {
     case didField(FormFieldAction<String>)
     case petnameField(FormFieldAction<String>)
+    case failFollow(_ error: String)
     case reset
 }
 
@@ -297,6 +298,8 @@ struct FollowUserFormModel: ModelProtocol {
         value: "",
         validate: Self.validatePetname
     )
+    
+    var failFollowMessage: String? = nil
     
     static func validateDid(key: String) -> Did? {
         Did(key)
@@ -324,6 +327,10 @@ struct FollowUserFormModel: ModelProtocol {
                 action: action,
                 environment: FormFieldEnvironment()
             )
+        case .failFollow(let error):
+            var model = state
+            model.failFollowMessage = error
+            return Update(state: model)
         case .reset:
             return update(
                 state: state,
--- a/xcode/Subconscious/Shared/Components/Common/Profile/FollowUserFormSheet.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/FollowUserFormSheet.swift
@@ -136,7 +136,7 @@ enum FollowNewUserFormSheetAction {
     case qrCodeScanned(scannedContent: String)
     case qrCodeScanError(error: String)
     
-    case failFollow(error: String, petname: Petname.Name)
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -400,7 +400,11 @@ actor AddressBookService {
         preventOverwrite: Bool = false
     ) -> AnyPublisher<Void, Error> {
         Future.detached {
-            try await self.followUser(did: did, petname: petname)
+            try await self.followUser(
+                did: did,
+                petname: petname,
+                preventOverwrite: preventOverwrite
```
bfollington, line 406: This parameter as not being passed before, which is why overwriting was even possible. It was a regression in the first place.

bfollington:
@gordonbrander Do you think it's better to:
1. mark collision as a validation error and force the user to pick a new name
2. replace the name with an available suffixed version in the form, or
3. dismiss the form and find a unique name in the background, adding a suffix if needed

I tried (1) first and liked it but currently we have (2). (3) seems like a hard sell without https://github.com/subconsciousnetwork/subconscious/issues/333.

xcode/Subconscious/Shared/Components/Common/Profile/FollowUserFormSheet.swift
```diff
@@ -183,6 +194,42 @@ struct FollowNewUserFormSheetModel: ModelProtocol {
...
+                state: state,
+                action: .form(
+                    .petnameField(.setValue(input: petname.verbatim))
+                ),
+                environment: environment
+            )
+        case .failToFindUniquePetname(let error):
+            // Not a huge deal, the user will have to enter a name themselves
+            logger.warning("Failed to find a unique petname: \(error)")
+            return Update(state: state)
```
gordonbrander, line 232: Do we display anything to the user in this case?

gordonbrander: I'm guessing the answer is yes, via the `failFollowMessage`

xcode/Subconscious/Shared/Components/Common/Profile/FollowUserFormSheet.swift
```diff
@@ -324,9 +374,21 @@ struct FollowUserFormModel: ModelProtocol {
                 action: action,
                 environment: FormFieldEnvironment()
             )
+        case .failFollowDueToPetnameCollision(let error):
+            var model = state
+            model.failFollowMessage = error
```
gordonbrander, line 379: Is this error going to provide a user-friendly message?

If not, in the past, We've talked about defining a `LocalizedUserMessage` protocol, as kind of a user-facing complement to [`LocalizedError`](https://developer.apple.com/documentation/foundation/localizederror). If needed, this might be an opportunity to introduce?

Picturing:

```swift
protocol LocalizedUserMessage {
    var userMessage: String { get }
}
```

gordonbrander: Update: NVM, looks like we're using LocalizedError for this. That SGTM. Let's make it a convention that LocalizedError always provides a user-friendly description.

gordonbrander (approved):
Looks great!

xcode/Subconscious/Shared/Components/Common/Profile/FollowUserFormSheet.swift
bfollington: I would argue no based on my comment:

> Not a huge deal, the user will have to enter a name themselves

To be clearer: when this occurs the input field will already be marked as invalid and have the caption changed to the message "Petname already in use" (which is the `localizedDescription` for the relevant `AddressBookServiceError`.

bfollington: I'll update my in-code comment with this context.

PR #782, opened 2023-07-05
2023 07 05 icon

### LINKED ISSUE #423: Adjust icon sizing in app icon
Currently we use the same size as e.g. mobile safari. However, it's too visually large, and most icons sit within a smaller inner circle bound. We should do a survey of other icons that have circular/close-to-circular marks, and align with the rough bounds of those app icons.

gordonbrander:
Fixes #423 

simulator_screenshot_8DA0E696-EFE5-4ED3-AD82-4DBAF0FB0A39


bfollington (approved):
LGTM!

Issue #641, opened 2023-05-15
Move profile to app tab bar


bfollington:
The more I use the app, the more it feels like we should have two tabs at the bottom: Notebook and Profile (and perhaps feed + discover in future)

It's annoying to reach the top-left corner for your profile when it's the portal to everyone else on the network. 

I imagine each tab having its own navigation stack, which gives you soft multitasking (like twitter).

PR #784, opened 2023-07-06
Move profile to app tab

gordonbrander:
Progress toward #641

## Architecture

- Move profile to own tab
- Factor out stack into DetailStack component
    - One stack for each tab
- Note deletion and other changes that have global significance are managed by the shared app store
    - This allows us to refresh lists across the app when a delete, add, or remove happens

## Decisions

- Continue to use AppStore to manage IO with global effects
    - The semantics we need are pub/sub with side-effects
    - This essentially describes store.
        - We could use the environment instead, and invent an abstraction that is like a store without state, but why?
        - Even though we do not need the store's state for some things, it makes sense to draft off of the already established tooling of the store


gordonbrander:
Was experiencing a mysterious crasher when factoring NavigationStack into a subview.

What I discovered was that, at some point, NavigationStack seems to be attempting to mutate the array that manages that stack. I believe the fact that we pass the stack down as a value type may be causing the problem. It's obscured because the issue happens in Apple's proprietary SwiftUI code. The crash has to do with an array manipulation though, so it is either to do with the array that manages the view stack, or perhaps it is to do with some sort of race condition in view rendering vs stack management on the SwiftUI DOM-patching side.

The workaround was giving NavigationStack a @State binding and replaying changes onto it.

However, I'm going to see if using a ViewStore with closure-based get semantics, like Binding may also work around the issue.

If so, it would provide a strong reason to complete the work on the next version of ObservableStore, which includes a Binding-like ViewStore abstraction.

gordonbrander:
I'm thinking this PR is probably too involved to roll into Beta, but I'm continuing to pull on the thread, because factoring the app into tabs will un-block our work on the Feed as well.

gordonbrander:
Update: ViewStore with closure get semantics does resolve the NavigationStack crasher.

So it seems the issue is passing state by value when NavigationStack expects a Binding. Note that Binding works by using a get and set closure to access state. If the state being accessed is a reference type (ObservableObject), then you get a live view of the data.

What this says to me is that we should be leaning into Binding-like approaches that use closure to read a live view of state, when using a SwiftUI view that expects a Binding.

gordonbrander:
On SwiftUI's update strategy:

> I’m not sure what you mean by connect, but SwiftUI only tracks `@State` and `@ObservedObject` variables inside a View. More accurately, Any views that reads from `@State` var foo are invalidated whenever foo changes, wherever foo is declared, and Any views that has `@ObservedObject` var foo are invalidated whenever foo.objectWillChange is triggered. Binding is just a way to read from/write to variables across views. If you mutate to ObservableObject, it will trigger `objectWillChange`, but if the view that mutates it does not observe it with `@ObservedObject` the view will not be invalidated. https://forums.swift.org/t/swiftui-bindings-does-reading-them-in-body-mean-view-will-re-render/34600/2

I think this description is correct. So what a Binding is doing is providing a window into a slice of state, but the binding itself does not trigger view invalidation. Rather, any binding relies on the parent `@State` or `@StateObject` or `@ObservedObject` in the parent to invalidate the parent, re-evaluate its body, which will in-turn re-evaluate the child.

What does this tell us? One is that Binding (and, by extension, ViewStore) are not re-rendering optimizations. They will re-render any time something in the root `ObservableObject` changes (Store, in our case). So we could achieve a similar effect by passing down the whole store, but offering a scoped view function over it. However, `ViewStore` does give us the advantage of encapsulation.

Some explorations of other approaches that are possible:

- There may be an opportunity to use `EquatableView` to customize the diffing. If we explore this path, we must take care to remember that the two instances of `ObservedObject` will be the same identity, so we would have to diff by another means, such as keeping the previous state in the Store.
- Another approach could be to create ObservableObjects for sub-components, and replay select events onto them from the parent store using Combine. This is the approach ComposableArchitecture takes. However, ObservableObjects must be reference types, and this means re-creating a new instance of the heap-allocated class every time the parent store changes, plus any escaping closures that are needed to map the Combine publisher.
    - ViewStore requires 2 escaping closures. These are heap-allocated too. We could alternatively use a protocol.

gordonbrander:
Closing this. Completing the refactor will depend upon updating ObservableStore to https://github.com/subconsciousnetwork/ObservableStore/pull/36

PR #794, opened 2023-07-11
Make QR Code shareable

### LINKED ISSUE #499: QR Code is drawn in Blue after re-render
On initial presentation the correct colour is used, but after re-rendering the view (in this case by unfollowing a user from the address book) the colour reverts to the default blue.

The reason for this seems to be that `Color.accentColor` is not always safe to call, see: https://stackoverflow.com/a/69123469

# Correct 
image

# Incorrect
image

### LINKED ISSUE #764: Share QR code to clipboard as image
As per the title. Give users the option to share the QR code itself, maybe as a little badge like so:

image

bfollington:
Fixes #764 
Fixes #499 

- Break apart DidQrCodeView
- Implement copyable QR code

image

image

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Profile/DidQrCodeView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/DidQrCodeView.swift
@@ -9,20 +9,21 @@ import Foundation
 import SwiftUI
 import CoreImage.CIFilterBuiltins
 
-struct DidQrCodeView: View {
-    var did: Did
-    var color: Color
-    
-    let context = CIContext()
-    let filter = CIFilter.qrCodeGenerator()
-    let colorInvertFilter = CIFilter.colorInvert()
-    let maskFilter = CIFilter.blendWithMask()
-    
-    func generateQRCode(from string: String) -> UIImage? {
-        filter.message = Data(string.utf8)
+extension DidQrCodeView {
+    static func generate(from did: String, color: Color) -> UIImage {
+        let context = CIContext()
+        let filter = CIFilter.qrCodeGenerator()
+        let colorInvertFilter = CIFilter.colorInvert()
+        let maskFilter = CIFilter.blendWithMask()
+        
+        let fallback = UIImage(systemName: "xmark.circle") ?? UIImage()
+        
+        filter.message = Data(did.utf8)
         
         colorInvertFilter.inputImage = filter.outputImage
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Components/Common/Profile/DidQrCodeView.swift
```diff
@@ -36,24 +37,80 @@ struct DidQrCodeView: View {
...
+struct ShareableDidQrCodeView: View {
+    var did: Did
+    var color: Color
+    
+    var img: UIImage {
+        DidQrCodeView.generate(from: did.did, color: color)
+    }
+    
+    var body: some View {
+        let shareable = Image(uiImage: img)
```
bfollington, line 96: This is the easiest way I found to get an imagine into a `Transferable` protocol-compatible form.

xcode/Subconscious/Shared/Components/Common/Profile/DidQrCodeView.swift
```diff
@@ -36,24 +37,80 @@ struct DidQrCodeView: View {
         let col = CIColor(red: r, green: g, blue: b, alpha: a)
         maskFilter.inputImage = CIImage(color: col)
         
-        guard let img = maskFilter.outputImage else { return nil }
-        guard let cgImg = context.createCGImage(img, from: img.extent) else { return nil }
+        guard let img = maskFilter.outputImage else {
+            return fallback
+        }
+        
+        let scaleFactor = 10.0
```
bfollington, line 44: The image was comically small upon sharing without this scaleup

xcode/Subconscious/Shared/Components/Common/Profile/DidQrCodeView.swift
```diff
@@ -36,24 +37,80 @@ struct DidQrCodeView: View {
...
+        let upscaleImg = img
+            .samplingNearest()
+            .transformed(
+                by: CGAffineTransformMakeScale(scaleFactor, scaleFactor)
+            )
```
bfollington, line 49: Couple hoops to jump through to ensure we get sharp pixels (nearest-neighbour magnification filter)

gordonbrander: Great stuff

gordonbrander (approved):
LGTM

Issue #793, opened 2023-07-11
All notes show the title as their excerpt by default


bfollington:
Our intention is to show previews of each note in the feed / transclusions so you can get a glimpse of content without clicking through. Currently, every preview will look like this by default:

<img width="375" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/4a1e9827-0938-49da-b996-fd3db8781a12">

When creating a note we autopopulate the first line with the title, which is fine as a pattern. The problem is that the title automatically becomes the first block and is the entirety of the excerpt as a result. To get the intended outcome I find myself making notes, writing them and then deleting the title afterwards.

<img width="428" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/35193da2-726a-4a97-9f66-0b30c1652d37">

I can see two fixes here:
1. Change the UX for naming a note to set the note title _without_ putting it in the body
2. Skip over the first block when picking the excerpt if (and only if) it matches the title

gordonbrander:
We should include the first 2 blocks in the excerpt (I thought we had made this change already 🤔)

bfollington:
I recall fixing the excerpt logic but not considering two blocks. Did a quick hack to test the first two block behaviour, it's better but it does still "double up" the title 

<img width="364" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/5eea7e74-350c-4b5c-af6a-b8588efbd3db">

gordonbrander:
That looks right to me. IMO the slug does not suffice as a title. 

We could also consider making the first block medium weight font. I believe the Figma mocks take this approach.

PR #795, opened 2023-07-11
Implement full transclusion design

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/799
Fixes https://github.com/subconsciousnetwork/subconscious/issues/793
Fixes https://github.com/subconsciousnetwork/subconscious/issues/328
Fixes https://github.com/subconsciousnetwork/subconscious/issues/811

- Display user pfp with slashlink on transcludes and backlinks
- Retire stopgap Transclude2View
- Cache user profiles to speed up loading
- Replace standalone slashlinks with transcludes
- Show up to the first _two_ blocks in the excerpt

Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 12 51 Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 14 44

Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 14 29 Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 14 17

https://github.com/subconsciousnetwork/subconscious/assets/5009316/71bf964d-7afd-406b-ba08-8b870cb89e64


xcode/Subconscious/Shared/Models/EntryStub.swift
```diff
@@ -7,6 +7,13 @@
 
 import Foundation
 
+struct AuthoredEntryStub: Identifiable, Equatable, Hashable {
```
bfollington, line 10: Annoying name, but this is accurate

bfollington:
Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 12 51 Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 14 44

Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 14 29 Simulator Screenshot - iPhone 14 Pro - 2023-07-17 at 11 14 17

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
```diff
@@ -38,17 +38,40 @@ struct SubtextView: View {
...
+            case _:
+                continue
+            }
+        }
+        
+        return block.body()
+            .trimmingCharacters(in: .punctuationCharacters)
+            .trimmingCharacters(in: .whitespaces)
+            .count <= count
+    }
```
bfollington, line 58: This was a suggestion in Discord from the team, for blocks where there is a slashlink on its own (aside from punctuation and whitespace) then we can replace it entirely with a transclude.

Unsure if this is actually a good idea or not from a principles POV, we've been showing the raw markup _and_ the rendered version so far.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeButtonStyle.swift
```diff
@@ -12,19 +12,40 @@ struct TranscludeButtonStyle: ButtonStyle {
...
+        default:
+            return Color.clear
+        }
+    }
+    
+    var stroke: Color {
+        colorScheme == .light
+            ? Color.separator
+            : Color.clear
+    }
```
bfollington, line 36: Vary the transclude style based on light vs dark mode.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
```diff
@@ -2,41 +2,95 @@
...
+    var spacing: CGFloat = AppTheme.unit
+    var excerptLines: [EnumeratedSequence<[String.SubSequence]>.Element] {
+        Array(excerpt.split(separator: "\n").enumerated())
+    }
+    
+    var body: some View {
+        VStack(alignment: .leading, spacing: spacing) {
+            ForEach(excerptLines, id: \.offset) { idx, line in
+                Text("\(String(line))")
+                    .fontWeight(idx == 0 && excerptLines.count > 1 ? .medium : .regular)
```
bfollington, line 22: Only one line? Regular weight, full opacity.

Two lines: the first is medium, full opacity, the second is regular weight and 0.8 opacity. On light mode the weight is enough but in dark mode we need more constrast.

xcode/Subconscious/Shared/AppTheme.swift
```diff
@@ -36,7 +36,7 @@ extension AppTheme {
     static let cornerRadius: Double = 8
     static let cornerRadiusLg: Double = 16
     
-    static let smPfpSize: CGFloat = 24
+    static let smPfpSize: CGFloat = 22
```
bfollington, line 39: In Figma the backlinks panel seems to have 24pt pfps but the transcludes have 22pt pfps. I think 22 looks better for transcludes and I'm tempted to use that for both cases.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -344,6 +345,39 @@ actor UserProfileService {
...
+            address: address,
+            resolutionStatus: resolutionStatus
+        )
+        
+        if let petname = address.petname {
+            self.cache.updateValue(profile, forKey: petname)
+        }
+        
+        return profile
+    }
```
bfollington, line 379: I needed a more general purpose way of fetching the minimal profile details and, in the process, I realised that our loading performance was gated by fetching profiles. I implemented a simple cache of petname to profile which is cleared whenever we sync, so theoretically it will always be up-to-date.

xcode/Subconscious/Shared/Components/Common/Byline/BylineSmView.swift
```diff
@@ -16,6 +16,7 @@ struct BylineSmView: View {
         HStack {
             ProfilePic(pfp: pfp, size: .small)
             SlashlinkDisplayView(slashlink: slashlink)
+                .theme(slug: .secondary)
```
gordonbrander, line 19: 👍

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
```diff
@@ -2,41 +2,95 @@
...
+                    .opacity(idx > 0 ? 0.8 : 1) // Helps massively in dark mode
+                    .lineLimit(lineLimit)
```
gordonbrander, line 24: I think we don't want to set a line limit on the transclude. Otherwise we get a strange effect where each line gets truncated.

Seems ok to let these get a bit long.

Edge case: we could set some high upper limit or truncation, as a sanity check against blocks that are > 50 lines long.

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
gordonbrander: Let's go with it and see how it feels in practice. (Aside: this is what I do on my website and it works pretty well)

In future when we have better across the board transclusion support, I want to consider hiding bare links and slashlinks in view mode. This is a good start.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeButtonStyle.swift
gordonbrander: Nice. I think we want to consider the wash style for light mode as well.

Perhaps tertiary background?

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeButtonStyle.swift
```diff
@@ -12,19 +12,40 @@ struct TranscludeButtonStyle: ButtonStyle {
...
     @Environment(\.isEnabled) private var isEnabled
-
+    @Environment(\.colorScheme) var colorScheme
+    
+    func background(_ configuration: Configuration) -> Color {
+        switch (colorScheme) {
+        case .dark:
+            return configuration.isPressed
+                ? Color.secondaryBackground
+                : Color.secondaryBackground.opacity(0.5)
```
gordonbrander, line 22: Should this be tertiaryBackground?

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
gordonbrander: Feels like we should find a treatment that can work in light and dark mode. Medium weight seems fine to me for first block.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
```diff
@@ -2,41 +2,95 @@
...
+//  Created by Gordon Brander on 2/22/23.
 //
 
 import SwiftUI
 
+struct ExcerptView: View {
+    var excerpt: String
+    var lineLimit: Int = 5
+    var spacing: CGFloat = AppTheme.unit
+    var excerptLines: [EnumeratedSequence<[String.SubSequence]>.Element] {
```
gordonbrander, line 14: Instead of flattening Subtext to a string and then splitting it again, we should have a method:

```
Subtext.truncate(maxBlocks: Int) -> Subtext
```

And hold a reference to this on the `EntryStub`. That way we're doing less work.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -643,16 +643,18 @@ extension Subtext {
...
     /// Derive an excerpt
     func excerpt(fallback: String = "") -> String {
-        for block in blocks {
-            switch block {
-            case .empty:
-                continue
-            default:
-                let block = String(block.body())
-                return block.isEmpty ? fallback : block
+        // Filter out empty blocks
```
gordonbrander, line 646: Mentioned above but we should introduce a `Subtext.truncate(maxBlocks: Int) -> Subtext` method and use that instead of flattening to a string and then splitting again.

gordonbrander (changes requested):
A few requests, but this generally looks great.

xcode/Subconscious/Shared/Models/EntryStub.swift
bfollington: Done. It's an optional field now and `TranscludeView` can handle the case where it's missing (hides the pfp). Currently we always populate `author` eagerly but this leaves room to lazy-load author information in the future while rendering transcludes immediately.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeButtonStyle.swift
bfollington: Maddeningly, `tertiaryBackground` is `white` on light mode, making it useless... Here's `tertiarySystemBackgroundGrouped`, which looks ok on light but is way too bright for dark mode, imo:

Simulator Screenshot - iPhone 14 - 2023-07-18 at 17 23 58 simulator_screenshot_FAE42048-77C3-4617-900A-F01C599A2A05

--- 

Here's my best attempt at a nice looking solution: using `tertiarySystemGroupedBackground` for light and `secondarySystemBackground` for dark.

Simulator Screenshot - iPhone 14 - 2023-07-18 at 17 29 47 Simulator Screenshot - iPhone 14 - 2023-07-18 at 17 29 57

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
bfollington: Removed the opacity tweak.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
bfollington: I'm not 100% clear on how this would actually work? We create the excerpt and save it to the DB as a string and then load that string to populate transclusions and backlinks, so it _has_ to be flatted to a string before we get access to it.

I considered re-parsing as Subtext on load but that seems _even more_ wasteful?

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
```diff
@@ -34,17 +34,44 @@ struct SubtextView: View {
...
 
     var body: some View {
-        VStack(alignment: .leading) {
+        VStack(alignment: .leading, spacing: AppTheme.tightPadding) {
```
gordonbrander, line 65: Nice

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeButtonStyle.swift
```diff
@@ -12,19 +12,33 @@ struct TranscludeButtonStyle: ButtonStyle {
...
     @Environment(\.isEnabled) private var isEnabled
-
+    @Environment(\.colorScheme) var colorScheme
+    
+    func background(_ configuration: Configuration) -> Color {
+        switch (colorScheme) {
+        case .dark:
+            return configuration.isPressed
+                ? Color.backgroundPressed
+                : Color.secondaryBackground
```
gordonbrander, line 22: Is tertiary too light for dark mode? Is secondary too dark for light mode?

Would be nice if we could unify these, but not strictly necessary.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeButtonStyle.swift
```diff
@@ -12,19 +12,33 @@ struct TranscludeButtonStyle: ButtonStyle {
...
+        case .light:
+            return configuration.isPressed
+                ? Color.backgroundPressed
+                : Color.tertiarySystemGroupedBackground
```
gordonbrander, line 26: Is tertiary too light for dark mode? Is secondary too dark for light mode? Would be nice if we could unify these, but not strictly necessary.

Non-blocking.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
gordonbrander: My mistake. You are correct.

xcode/Subconscious/Shared/Parsers/Subtext.swift
gordonbrander: NVM, per above.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -477,7 +477,8 @@ final class DatabaseService {
             return EntryStub(
                 address: address,
                 excerpt: excerpt,
-                modified: modified
+                modified: modified,
+                author: nil
```
gordonbrander, line 481: Nit: seems unfortunate we can't get the author info from the DB. DID and petname are technically in there.

Non-blocking.

xcode/Subconscious/Shared/Services/TranscludeService.swift
```diff
@@ -11,15 +11,22 @@ import Combine
 struct Transclusion {
     let displayAddress: Slashlink
     let address: Slashlink
+    let absoluteAddress: Slashlink
```
gordonbrander, line 14: Should we instead store a did on the Transclusion?

gordonbrander (approved):
A few questions, but LGTM

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Yah. We could build a DB cache for the profile information and then retrieve it as part of queries like this. It would simplify a large part of the loading process.

Tracked in https://github.com/subconsciousnetwork/subconscious/issues/813

xcode/Subconscious/Shared/Services/TranscludeService.swift
bfollington: Yes, that is better.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeButtonStyle.swift
bfollington: I caved and made it consistent, I think it's too bright on dark mode but I'll let it settle in.

PR #807, opened 2023-07-17
Remove "Top" tab from profile

bfollington:
Refactor tabbed column view to use TabView, much simpler internally


gordonbrander (approved):
LGTM

PR #809, opened 2023-07-18
Minor fixup on TabView

bfollington:
- Fix gap at the bottom of user profile TabView
- The loading state tab still said "Recent" but the loaded state said "Notes"


gordonbrander (approved):
LGTM

Issue #836, opened 2023-07-25
Cannot change audience of note, "file already exists"


bfollington:
Experienced by @cdata and @gordonbrander, hitting a "file already exists" error when changing the audience of a note.

Notes _without a slug_ (aka "Untitled") cannot have their audience or link updated (known issues https://github.com/subconsciousnetwork/subconscious/issues/582). This is a different case where changing audience fails unexpectedly, perhaps related to https://github.com/subconsciousnetwork/subconscious/issues/798?

Multiple of the same slug is the common element I am suspicious of.

PR #837, opened 2023-07-26
Use `Slashlink.verbatim` when changing audience

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/836

You were right @gordonbrander, we were looking at `description` here. It was actually quite subtle, since Slashlink uses `description` to meet the `Identifiable` protocol we were treating local addresses as being _equal to_ public ones.

`Slashlink.description` preserves the old behaviour, `Slashlink.markup` strips the local prefix (`did:subconscious:local`) but `verbatim` and `verbatimMarkup` preserve the prefix. My logic is:

- Description is human readable and detail preserving
- Verbatim is case preserving and detail preserving
- Markup is designed to be rendered in UI and inserted into text

### DIFF
--- a/xcode/Subconscious/Shared/Models/Slashlink.swift
+++ b/xcode/Subconscious/Shared/Models/Slashlink.swift
@@ -31,9 +31,9 @@ public struct Slashlink:
     
     public var description: String {
         guard let peer = peer else {
-            return slug.markup
+            return slug.description
         }
-        return "\(peer.markup)\(slug.markup)"
+        return "\(peer.description)\(slug.markup)"
     }
 
     public var verbatim: String {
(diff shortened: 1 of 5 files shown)


gordonbrander (changes requested):
@bfollington Slashlink `description` should not drop `did:subconscious:local`, since it is the literal string description of the slashlink. `markup` can drop `did:subconscious:local` since it is the representation meant to be used in Subtext markup.

I think this would resolve the `description` footgun?

xcode/Subconscious/Shared/Models/Slashlink.swift
```diff
@@ -46,7 +43,12 @@ public struct Slashlink:
     public var id: String { description }
     
     // The normalized markup form of the slashlink
-    public var markup: String { description }
+    public var markup: String {
+        guard let peer = peer else {
+            return slug.markup
+        }
+        return "\(peer.markup)\(slug.markup)"
+    }
```
bfollington, line 51: _Only_ `.markup` will strip the `local:subconscious:did` prefix, `description` preserves the old behaviour as-is.

bfollington:
@gordonbrander another round, I think this is clearer.

`Slashlink.description` preserves the old behaviour, `Slashlink.markup` strips the local prefix (`did:subconscious:local`) but `verbatim` and `verbatimMarkup` preserve the prefix. 

My logic is:
- Description is human readable and detail preserving
- Verbatim is case preserving and detail preserving
- Markup is designed to be rendered in UI and inserted into text

xcode/Subconscious/Shared/Models/Slashlink.swift
gordonbrander: That's perfect. We want `description` to be `LosslessStringConvertible, so it should include the full did or petname.

gordonbrander (approved):
LGTM

Issue #798, opened 2023-07-11
Opening the same note multiple times and then changing its audience results in two notes with same slug


gordonbrander:
Opening the same note multiple times and then changing its audience results in two notes with same slug.

### Steps to reproduce

Setup:

- Create local draft "/foo"
- Add a link to `/bar`
- Follow link to `/bar`
- Add link to `/foo` in `/bar`
- Follow link to`/foo`

You now have an editor stack that looks like:

1. `/foo`
2. `/bar`
3. `/foo`

Then...

- Change topmost foo (3) audience to public.
- Swipe back until you get to foo (1)

Note that foo is still private. When you swipe back to notes list, you will see two foos.

What should happen:

- foo (1) should be public, no second note should be created
    - We should be notifying the list when the audience changes and updating each entry in the stack. Also, when you access a stack item, it should refresh itself on-appear, and discover it has changed.

Issue #583, opened 2023-05-08
Deleting a note doesn't close current note, duplicates


jsantell:
When deleting a note, after confirming deletion, the view is of the same note in an editing state, and any change (e.g. adding a character) recreates the note in its entirety. I'd expect after deleting a note to not see the note, and probably go back to the notes view and avoid the duplication scenario.

PR #839, opened 2023-07-27
Fix stale audience in navigation stack

bfollington:
Relies on https://github.com/subconsciousnetwork/subconscious/pull/837
Fixes https://github.com/subconsciousnetwork/subconscious/issues/798

Prefer updated address to stale address after changing audience.

Interestingly, this _also_ fixes https://github.com/subconsciousnetwork/subconscious/issues/583

### DIFF
--- a/xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
+++ b/xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
@@ -987,7 +987,7 @@ struct MemoEditorDetailModel: ModelProtocol {
         info: MemoEditorDetailDescription
     ) -> Update<MemoEditorDetailModel> {
         // No address? This is a draft.
-        guard let address = state.address ?? info.address else {
+        guard let address = info.address ?? state.address else {
             return update(
                 state: state,
                 action: .setDraftDetail(


xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -987,7 +987,7 @@ struct MemoEditorDetailModel: ModelProtocol {
         info: MemoEditorDetailDescription
     ) -> Update<MemoEditorDetailModel> {
         // No address? This is a draft.
-        guard let address = state.address ?? info.address else {
+        guard let address = info.address ?? state.address else {
```
gordonbrander, line 990: Reasoning about what this inversion is doing is tricky. Can you explain what's happening?

gordonbrander (approved):
LGTM, but I have a question before merging.

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
bfollington: When we rewrite the navigation stack after a rename we rewrite the list of `details` to point to the new address (with updated audience). When we pop the navigation stack we dispatch `send(.appear(description))` in `onAppear {}` which contains the _updated_ address. This is our chance to reload the data and refresh the view.

The bug was that we were throwing the _new_ address away (with updated audience) and retaining the address stored in `state`, which no longer resolves. 

TL;DR, in `.appear`'s handler, `state` is stale and `info` is fresh.

gordonbrander: That scans. Let's get this merged 🙏

PR #843, opened 2023-08-03
Massively improve profile load times

bfollington:
We were _always_ reading the user's sphere to populate the list of notes on the profile. This PR introduces an optimisation to read notes from the DB index if we follow the user who's profile we're loading. 

Combining this with the existing cache of user profile metadata makes it practically instant to load any profile we follow.

### DIFF
--- a/xcode/Subconscious/Shared/Services/UserProfileService.swift
+++ b/xcode/Subconscious/Shared/Services/UserProfileService.swift
@@ -399,13 +399,32 @@ actor UserProfileService {
         let following = try await self.getFollowingList(
             address: address
         )
-        let notes = try await sphere.list()
         let followingStatus = await self.addressBook.followingStatus(did: did)
         
-        let entries = try await self.loadEntries(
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -399,13 +399,32 @@ actor UserProfileService {
...
-        let entries = try await self.loadEntries(
-            address: address,
-            slugs: notes
-        )
+        let entries = try await Func.run {
+            switch (followingStatus) {
+            // Read from DB if we follow this user
+            case .following(let name):
+                return try
+                    self.database.listRecentMemos(owner: did)
```
gordonbrander, line 409: Noting that it is technically possible to JOIN tables and map the DID to a petname, returning the petname with the data.

However, I do not think we should worry about this rn.

gordonbrander (approved):
LGTM

Issue #830, opened 2023-07-25
Local notes appearing in "Backllinks" / "Related" panel have no pfp


bfollington:
This makes sense, since they're not part of the sphere, but it does look odd. We should fall back to the icon for `local` audience if there's no pfp available instead.

PR #844, opened 2023-08-07
Show draft icon as pfp for local transcludes

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/830

If `pfp` is `.none` then the icon for "local" audience is used in its place.

image

image


gordonbrander (approved):
LGTM

Issue #825, opened 2023-07-25
Profile tabs are only selectable on the text, not the outside "pill box"


jsantell:
Clicking tabs in the profile view, only when selecting the text ("Notes", "Following") does the tab change, not any of the white space around it

PR #847, opened 2023-08-07
Fix profile tab tap targets

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/825

- Fix hit target
- Drop superfluous VStack


gordonbrander (approved):
LGTM

PR #852, opened 2023-08-08
Refresh profile after follow, indexing and protect against stale data

bfollington:
- Use version to detect stale profile cache entries
- Refresh profile when indexing complete


gordonbrander (approved):
LGTM

Issue #851, opened 2023-08-08
Local note is appearing on 3P user's profile


bfollington:
Reading notes from the DB allows for massive speedup in profile loading (see #843) but our query seems to be fetching local drafts and returning them as part of the list. 

This is because `DatabaseService.listRecentMemos` implicitly includes `Did.local` in the list of DID's to look up.

PR #853, opened 2023-08-08
Exclude local drafts from recent memos list

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/851

Maybe not the cleanest API, we have to differentiate between the case where we are listing the _user's_ recent notes vs someone they follow. We should only include local drafts when dealing with _our_ content.


gordonbrander (approved):
LGTM

Issue #817, opened 2023-07-21
Autogenerated slugs are using first two blocks, making them extremely long


bfollington:
This was introduced when we extended the excerpt. We need to handle slug generation differently, perhaps with a dedicated method.

bfollington:
Maybe we relax the `.medium` font weight and redux the maximum number of characters?

@gordonbrander also suggests that parsing the first 1-2 blocks as subtext for rendering might be a good idea to enable title styles etc.

Probably best to load the raw string and only parse it when the view is rendered, we'll eventually have to start using virtualized lists of notes and shouldn't parse until it hits the viewport.

Issue #804, opened 2023-07-11
Improve note excerpt formatting and display, render subtext markup


bfollington:
The excerpt shown for backlinks and trancsludes can include formatting characters like `[[`, `*`, `_` etc. without actually rendering them as they appear in the editor and viewer. We should strip down to the "essential" characters for the excerpt.

Consider fixing https://github.com/subconsciousnetwork/subconscious/issues/793 at the same time.

<img width="399" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/39061774-0ea9-475f-bd79-8117c65c3ed6">

bfollington:
@gordonbrander suggests that maybe we actually render these out to an attributed string.

We also need to restrict the size of the excerpt to a max number of blocks and characters, the max is very high right now.

PR #855, opened 2023-08-08
Restrict auto-slug generation to first non-empty block

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/817
Related to https://github.com/subconsciousnetwork/subconscious/issues/804

- Introduce Subtext.generateSlug
- Reduce excerpt length, it's too long in practice imo
- Introduce `truncatedPrefix()` to add ellipsis (…) if string is truncated

280 char excerpt:

image

512 char excerpt (old value):

image

### DIFF
--- a/xcode/Subconscious/Shared/Library/StringUtilities.swift
+++ b/xcode/Subconscious/Shared/Library/StringUtilities.swift
@@ -74,3 +74,14 @@ extension StringProtocol {
         self.prefix(1).capitalized + self.dropFirst()
     }
 }
+
+extension String {
+    /// Similar to `prefix` but inserts an ellipsis (…) if truncation occurs
+    func truncatedPrefix(_ count: Int) -> String {
+        let truncated =
+            String(self.prefix(count - (self.count > count ? 1 : 0)))
+                .trimmingCharacters(in: .whitespacesAndNewlines)
+        
+        return self.count > count ? truncated + "…" : truncated
+    }
+}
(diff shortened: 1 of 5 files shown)


xcode/Subconscious/Shared/Library/StringUtilities.swift
```diff
@@ -74,3 +74,14 @@ extension StringProtocol {
...
+extension String {
+    /// Similar to `prefix` but inserts an ellipsis (…) if truncation occurs
+    func truncatedPrefix(_ count: Int) -> String {
+        let truncated =
+            String(self.prefix(count - (self.count > count ? 1 : 0)))
+                .trimmingCharacters(in: .whitespacesAndNewlines)
+        
+        return self.count > count ? truncated + "…" : truncated
+    }
+}
```
gordonbrander, line 87: We already have `String.truncate` in `Shared/Library/Prose.swift`. Does this not cover the same usecase?

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -663,10 +663,30 @@ extension Subtext {
         return output.isEmpty ? fallback : output
     }
     
+    static let maxExcerptSize = 280
+    static let maxGeneratedSlugSize = 128
```
gordonbrander, line 667: These should probably be private

gordonbrander (changes requested):
Small change requested. We already have a truncation utility function.

xcode/Subconscious/Shared/Library/StringUtilities.swift
bfollington: ty! I had a feeling I'd done this before

gordonbrander (approved):
LGTM

Issue #857, opened 2023-08-09
Stale data in backlinks view when navigating "back" in history


bfollington:
Backlinks in the past history items can go out-of-date if a note is renamed or updated by a view higher in the navigation stack. This data should refresh along with everything else when the view is presented, so there must be a caching issue that results in the same data for transcludes.

Likely fix: use the sphere `cid` as the cache key when dealing with cached transcludes (same approach we use for expiring the user profile metadata cache)

PR #859, opened 2023-08-09
Refresh backlinks regardless of whether editor content is stale

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/857

# Changes
- Backlinks are now loaded independently of the memo in both the viewer and editor
- Backlinks are refreshed regardless of whether the editor content is refresh on `.appear`
- `DataService.readMemoDetail` and `DataService.readEditorMemoDetail` are both much simpler due to the introduction of `DataService.readMemoBacklinks`
- Both the viewer an editor will refresh backlinks and transcludes (in the case of the viewer) if indexing completes in the BG
    - To accomplish this I had to pass `app` down to the viewer and editor so they can replay select actions
    - I'd already done this on the profile, so I refactored into a common `.fromAppAction` pattern


bfollington:
After chatting to @gordonbrander we think it's more sensible to hold off merging this and split apart backlink refresh from editor content refresh.

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -1000,11 +1048,13 @@ struct MemoEditorDetailModel: ModelProtocol {
...
-                fallback: info.fallback,
-                autofocus: false
-            ),
+            actions: [
+                .loadDetail(
+                    address: address,
+                    fallback: info.fallback,
+                    autofocus: false
+                ),
+            ],
```
bfollington, line 1057: revert

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -1165,7 +1215,7 @@ struct MemoEditorDetailModel: ModelProtocol {
         }).eraseToAnyPublisher()
         
         let model = prepareLoadDetail(state)
-        return Update(state: model, fx: fx)
+        return update(state: model, action: .refreshBacklinks(address), environment: environment).mergeFx(fx)
```
bfollington, line 1218: break line

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -1281,6 +1330,7 @@ struct MemoEditorDetailModel: ModelProtocol {
             left: FileFingerprint(state),
             right: FileFingerprint(detail)
         )
+
```
bfollington, line 1333: drop

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -261,6 +272,25 @@ extension MemoViewerDetailAction: CustomLogStringConvertible {
     }
 }
 
+/// React to actions from the root app store
+extension MemoViewerDetailAction {
+    static func fromAppAction(
+        action: AppAction,
+        description: MemoViewerDetailDescription
```
gordonbrander, line 279: Same comment as above. Best to avoid passing in state. Instead, include relevant state in action being mapped.

gordonbrander (changes requested):
A couple notes

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -441,15 +441,12 @@ private struct EditProfileSheetModifier: ViewModifier {
...
-                case .succeedIndexPeer(let peer) where peer.identity == store.state.user?.did:
-                    store.send(.refresh(forceSync: false))
-                    break
-                case _:
-                    break
-                }
-            })
+            .onReceive(
+                app.actions.compactMap { action in
+                    UserProfileDetailAction.fromAppAction(action: action)
```
gordonbrander, line 446: Nit: you could pass this function directly to compactMap instead of wrapping in closure.

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -196,6 +198,12 @@ struct MemoEditorDetailView: View {
         ) { action in
             notify(action)
         }
+        .onReceive(
+            app.actions.compactMap { action in
+                MemoEditorDetailAction.fromAppAction(action: action, description: description)
```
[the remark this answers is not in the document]
gordonbrander: Still need to address

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -477,6 +490,26 @@ extension MemoEditorDetailAction {
...
 
+/// React to actions from the root app store
+extension MemoEditorDetailAction {
+    static func fromAppAction(
+        action: AppAction,
+        description: MemoEditorDetailDescription
+    ) -> MemoEditorDetailAction? {
+        guard let address = description.address else {
+            return nil
+        }
```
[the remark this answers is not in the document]
gordonbrander: Still need to address

xcode/Subconscious/Shared/Models/MemoDetailResponse.swift
```diff
@@ -10,7 +10,6 @@ import Foundation
 /// Response when reading info for a view-only memo detail from data source.
 struct MemoDetailResponse: Hashable {
     var entry: MemoEntry
-    var backlinks: [EntryStub] = []
```
gordonbrander, line 13: Since this struct only carries an entry now, should we just respond with a MemoEntry?

xcode/Subconscious/Shared/Models/MemoEditorDetailResponse.swift
```diff
@@ -11,7 +11,6 @@ import Foundation
 struct MemoEditorDetailResponse: Hashable {
     var saveState: SaveState
     var entry: MemoEntry
-    var backlinks: [EntryStub] = []
```
gordonbrander, line 14: Since this struct only carries an entry now, should we just respond with a MemoEntry?

gordonbrander (changes requested):
Direction looks good to me, but I have a few requests.

xcode/Subconscious/Shared/Models/MemoEditorDetailResponse.swift
bfollington: I think we still want to attach `saveState` here, so the struct makes sense?

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -1164,8 +1204,14 @@ struct MemoEditorDetailModel: ModelProtocol {
             Just(MemoEditorDetailAction.failLoadDetail(error.localizedDescription))
         }).eraseToAnyPublisher()
         
-        let model = prepareLoadDetail(state)
-        return Update(state: model, fx: fx)
+        var model = prepareLoadDetail(state)
+        model.address = address
```
bfollington, line 1208: Address must be set on model to fetch backlinks as early as possible.

xcode/Subconscious/Shared/Models/MemoEditorDetailResponse.swift
gordonbrander: Oops, missed that. You're right.

gordonbrander (approved):
LGTM

Would suggest playtesting since this touches the editor:

- Creating a note
- Deleting a note
- Renaming a note
- Creating a new note A, linking off to another new note B from it, following link to create B... does A show up in the related notes section?
- Opening the same note multiple times, changing the top one, paging back to the lower one. What happens?

PR #863, opened 2023-08-10
Log privacy (protect PII)

bfollington:
This has been a bit of an adventure. Things I've learned:

[`OSLogPrivacy`](https://developer.apple.com/documentation/os/oslogprivacy) does what we want already, masking sensitive values in release builds. It will only censor values when running on hardware and not connected to a debugging session. 

Additionally, anything which interpolates a _string_ or _object_ will be treated as `OSLogPrivacy.private`, logs containing static text and interpolating _numbers_ are treated as public. See https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code#3665948

> The unified logging system uses privacy options to hide or show interpolated variables in a message. By default, the system doesn’t redact integer, floating-point and Boolean values, but it does redact the contents of dynamic strings and complex dynamic objects. To make a private value public again, configure the privacy of the variable using appropriate modifiers in your message string or interpolated variable. For example, the following code shows how to make a dynamic string visible again using the public modifier:

_This means that, before this PR, almost all of our interpolations & logs will automatically be classed as `private`._

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -128,7 +128,7 @@ enum AppAction: CustomLogStringConvertible {
     case failCreateSphere(_ message: String)
 
     /// Set identity of sphere
-    case setSphereIdentity(String?)
+    case setSphereIdentity(Did?)
     /// Fetch the latest sphere version, and store on model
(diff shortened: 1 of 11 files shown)


bfollington:
Dead in favour of #864

Issue #858, opened 2023-08-09
Scrub potential PII (Personally Identifiable Information) from sentry logs


bfollington:
https://docs.sentry.io/platforms/javascript/data-management/sensitive-data/

Prevent leaking:
- DID
- Slashlinks/slugs
- CID

I am thinking we use the beforeSend/beforeBreadcrumb methods to regex match on slashlinks, CIDs and DIDs appearing in the logs & replace them? Maybe we can preserve the last few characters of  DID/CIDs for debugging purposes, or hash them?

PR #864, opened 2023-08-10
Logging, PII and Sentry

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/858

This has been a bit of an adventure. Things I've learned:

[`OSLogPrivacy`](https://developer.apple.com/documentation/os/oslogprivacy) does what we want already, masking sensitive values in release builds. It will only censor values when running on hardware and not connected to a debugging session. 

Additionally, anything which interpolates a _string_ or _object_ will be treated as `OSLogPrivacy.private`, logs containing static text and interpolating _numbers_ are treated as public. See https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code#3665948

> The unified logging system uses privacy options to hide or show interpolated variables in a message. By default, the system doesn’t redact integer, floating-point and Boolean values, but it does redact the contents of dynamic strings and complex dynamic objects. To make a private value public again, configure the privacy of the variable using appropriate modifiers in your message string or interpolated variable. For example, the following code shows how to make a dynamic string visible again using the public modifier:

_This means that, even before this PR, almost all of our interpolations & logs will automatically be classed as `private`._ 

**We'd have to _opt-in_ to making these values public in our logs and, since Sentry reads from the os logs it follows the same rules.**

image

# Changes

`logger.log(level:message:metadata)` now explicitly formats its log message to make the `message` public and the payload private, without this change we would see `"<private> <mask.hash 12321>"` but with it we see `"My message here <mask.hash 12321>"`

I would be nice to configure privacy for each value in the metadata dictionary but I've burned _way_ too much time on that. I don't think it's possible without exploding each KV pair out on to its own log line. Why? You can only specify privacy in an `OSLogMessage` and `OSLogMessage` can _only_ be constructed by passing a statically typed string interpolation to a logger (which we can't do because the metadata dictionary is of course dynamic). **You cannot compose these `OSLogMessage` at all or combine them with standard string interpolation and make use of the privacy system.**

# TL;DR

The app will do the right thing (block PII from Sentry logs) in production as-is, this PR slightly improves readability.

### DIFF
--- a/xcode/Subconscious/Shared/Library/LogFmt.swift
+++ b/xcode/Subconscious/Shared/Library/LogFmt.swift
@@ -15,9 +15,9 @@ struct LogFmt {
     
     private static func formatParameter(
         key: String,
-        value: String
+        value: String?
     ) -> String {
-        let value = Self.escapeValuePart(value)
+        let value = Self.escapeValuePart(value ?? "nil")
         return #"\#(key)="\#(value)""#
     }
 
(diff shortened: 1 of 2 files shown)


gordonbrander (approved):
LGTM

PR #869, opened 2023-08-11
Manage cache validation by comparing version CID

bfollington:
- Use version to test for cache validity
- Remove all manual cache invalidation
- Rename field

### DIFF
--- a/xcode/Subconscious/Shared/Services/AddressBookService.swift
+++ b/xcode/Subconscious/Shared/Services/AddressBookService.swift
@@ -311,8 +311,8 @@ actor AddressBookService {
         .eraseToAnyPublisher()
     }
     
-    /// Associates the passed DID with the passed petname within the sphere, clears the cache,
-    /// clears the cache, saves the changes and updates the database.
+    /// Associates the passed DID with the passed petname within the sphere,
+    /// saves the changes and updates the database.
     func followUser(
         did: Did,
         petname: Petname,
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Services/AddressBookService.swift
```diff
@@ -68,37 +69,36 @@ extension AddressBookError: LocalizedError {
...
+        
         for petname in petnames {
+            // Re-use the cached result if the version matches
+            if let entry = self.cache[petname],
+               version == entry.version {
+                entries.append(entry)
+                
+                continue
+            }
+
```
gordonbrander, line 101: It seems we are patching the cache entry-by-entry.

A more straightforward approach may be to check if the version differs from the cached version, and if so, wholesale replace the petname entries with those we get back from `sphere.listPetnames()`.

We have to iterate over the full list of petnames anyway, so this should be the same time complexity. Additionally, replacing the cache has the advantage of being able to exit early if the sphere versions match.

gordonbrander (changes requested):
One comment/question

xcode/Subconscious/Shared/Services/AddressBookService.swift
bfollington: Hm, I didn't want to pay the cost of rebuilding the entire cache in one hit if we didn't need to but I suppose that's a negligible amount of work in practice.

The address book differs slightly from the user profile cache though in that we never fetch an item directly from the cache - we always list it. That has me leaning towards your suggestion for the address bool but leaving the user profile cache as-is? I'll do another pass on Monday.

gordonbrander (approved):
LGTM, with one minor suggestion

Issue #776, opened 2023-06-28
Ability to rename someone in my address book


cdata:
Currently it is sort-of possible to rename someone in my address book, but it is a multi-step process:

1. Somehow know the DID of the person being renamed
2. Remove that person from your address book
3. Re-add them under the new name

It would be nice if we could make this a direct verb in the UX, so that as a user I can modify an entry that appears in my address book and set a new name for it directly.

PR #870, opened 2023-08-11
Rename users in address book

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/776

This is a quick way to implement to repurpose follow sheet for renaming users.

image 

image


xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -359,6 +368,46 @@ private struct FollowModifier: ViewModifier {
     }
 }
 
+private struct RenameModifier: ViewModifier {
```
gordonbrander, line 371: Note that `private` here means "private to my project", not "private to this file".

- Recommend moving this to a separate file, or into `FollowUserSheet`
- Recommend giving it specific name, perhaps `RenameSheetModifier`

gordonbrander (changes requested):
Small request. Otherwise LGTM.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
bfollington: Do you think this applies to the other modifiers in this file too? Should we break them all out? I was trying to avoid polluting the namespace with view modifiers so I actually would almost prefer `fileprivate` and keeping them at the use-site.

I'll try moving things around and see where we land. Agreed RE: the naming of the modifier.

bfollington: > Note that private here means "private to my project", not "private to this file".

Based on my testing this isn't true, it means "private to the containing lexical scope" which in this case _is_ the file. `private` at the root of a file is equivalent to `fileprivate`, no other file can access the type.

https://forums.swift.org/t/calling-private-methods-from-extensions-in-separate-files/54029

```swift
// A.swift

fileprivate struct One {}
private struct Two {}
internal Three {}
struct Four {} // implicitly `internal` by default
public struct Five {}
```

```swift
// B.swift

let one: One? = nil // Error: Cannot find 'One' in scope
let two: Two? = nil  // Error: Cannot find 'One' in scope
let three: Three? = nil
let four: Four? = nil
let five: Five? = nil
```

bfollington: I've moved all these sheets into `UserProfileView+SheetModifier` and clarified the naming convention.

bfollington: Huh, when trying to use something marked as `private` at the file root the error says that it's actually _`fileprivate`_. I guess internally `private` at file root === `fileprivate`.

image

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView+SheetModifier.swift
gordonbrander, line 1: Could call this UserProfileSheets.swift. Or, we could split these sheets into multiple files.

`+` is usually used to mean "extension".

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -263,14 +264,21 @@ private extension View {
         state: UserProfileDetailModel,
```
gordonbrander, line 264: Extend 'UserProfileView`, not `View`.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -734,6 +784,87 @@ struct UserProfileDetailModel: ModelProtocol {
...
+        from: Petname,
+        to: Petname
+    ) -> Update<Self> {
+        var model = state
+        model.renameCandidate = nil
+        model.failRenameMessage = nil
+        return update(state: model, actions: [
+            .presentRenameSheet(false),
+            .refresh(forceSync: true)
+        ], environment: environment)
```
gordonbrander, line 854: Formatting

gordonbrander (approved):
Approved, with a couple small requests.

Issue #634, opened 2023-05-12
Memo viewer should have a button to view the author's profile


bfollington:
This could be a toolbar icon or in the metasheet. Currently if you're reading a note it's difficult to navigate to the author of that note if you wanted to read more.

(workaround: you can copy the link and then slice of the slug, which I have done, which is why I logged this issue)

PR #878, opened 2023-08-17
View author profile from metasheet

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/634

- View author profile from memo viewer metasheet

Simulator Screenshot - iPhone 14 - 2023-09-27 at 14 25 00

### DIFF
--- a/xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
+++ b/xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
@@ -12,8 +12,7 @@ import ObservableStore
 
 struct MemoViewerDetailMetaSheetView: View {
     @Environment(\.dismiss) private var dismiss
-    var state: MemoViewerDetailMetaSheetModel
-    var send: (MemoViewerDetailMetaSheetAction) -> Void
+    var store: ViewStore<MemoViewerDetailMetaSheetModel>
     var onViewAuthorProfile: () -> Void
     
     var body: some View {
@@ -21,7 +20,7 @@ struct MemoViewerDetailMetaSheetView: View {
             HStack {
                 VStack(alignment: .leading, spacing: AppTheme.unit2) {
                     HStack {
-                        if let slashlink = state.address {
+                        if let slashlink = store.state.address {
                             SlashlinkDisplayView(slashlink: slashlink).theme(
                                 base: Color.primary,
                                 slug: Color.secondary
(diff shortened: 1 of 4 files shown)


xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
```diff
@@ -14,7 +14,8 @@ struct MemoViewerDetailMetaSheetView: View {
     @Environment(\.dismiss) private var dismiss
     var state: MemoViewerDetailMetaSheetModel
     var send: (MemoViewerDetailMetaSheetAction) -> Void
-
+    var onViewAuthorProfile: () -> Void
```
gordonbrander, line 17: I would like to funnel this through `send` so we don't have to call two things and pass two things, and everything goes through one path.

I think the right way to do this is:

- pass down a `ViewStore<MemoViewerDetailMetaSheetModel>`
- send up a new action `MemoViewerDetailMetaSheetAction.requestUserProfileDetail`
  - handle as a no-op in `MemoViewerDetailMetaSheetModel.update`
- Use cursor to tag into `MemoViewerDetailAction.requestUserProfileDetail` in 
- Use `onReceive` and `compactMap` to replay select actions from `Store<MemoViewerDetailMetaSheetModel>` on `notify`.
    - Map `MemoViewerDetailAction.requestUserProfileDetail` to an equivalent `MemoViewerDetailNotification`

This way only `MemoViewerDetailView` has to know about `notification`. The children are just consuming ordinary state or ViewStore.

WDYT?

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
```diff
@@ -45,6 +46,20 @@ struct MemoViewerDetailMetaSheetView: View {
                             item: state.shareableLink ?? ""
                         )
                         .disabled(state.shareableLink == nil)
+
```
gordonbrander, line 49: There should be a `Divider()` here.

gordonbrander (changes requested):
A couple changes requested.

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
bfollington: Done

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailMetaSheetView.swift
bfollington: Yep, I like it. I believe I've implemented this as you explained above, let me know if I missed something.

gordonbrander (approved):
LGTM!

Issue #767, opened 2023-06-28
Following the same DID under multiple names leads to confusing UX


bfollington:
Instead, if we have multiple names for one DID we could show all the names in the AKA section on **one** entry.

@cdata reminded me there is an FFI method for `DID -> [Petname]` that can help us here, but we will also need to remodel the `PetnameView` adding another `NameVariant` case for multiple names.

<img width=320 src=https://github.com/subconsciousnetwork/subconscious/assets/5009316/0e0e339f-bc54-4731-bd86-ab26960f3fbc>

<img width=320 src=https://github.com/subconsciousnetwork/subconscious/assets/5009316/08da3d87-db98-4e8f-80e3-a90e2e2425f7>

PR #879, opened 2023-08-19
Introduce `expectedName` for following status

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/767

Before:

image

After:

Screenshot 2023-08-19 at 3 48 02 pm


bfollington:
> Direction LGTM, but don't we want to keep the AKA underneath?
> 
> E.g.
> 
> ```
> @another-one
> AKA @dev-b
> 
> @dev-b
> AKA @another-one
> ```

I agree, did another pass on it. Now the list of aliases is displayed in the following list + on the profile header.

PR #883, opened 2023-08-30
Fix loadEntries so sphere is only opened once

gordonbrander:
`UserProfileService.loadEntries` was previously opening a sphere for the address on every turn of the for loop.

This PR fixes the logic so the sphere is only opened once, at beginning of function call.


bfollington (approved):
LGTM!

PR #887, opened 2023-09-06
Create workspace and SubconsciousCore framework

gordonbrander:
- Create a workspace
- Add Subconscious (app .xcproj) to that workspace
- Add a SubconsciousCore framework to that workspace


gordonbrander:
Spoke with Jordan:

- Workspaces can contain multiple projects and frameworks
- Projects can contain frameworks, but not other projects

We probably want to have a project for the time-being.

PR #888, opened 2023-09-06
Change bundle identifier via config

bfollington:
Customise bundle identifier based on whether we're building dev vs release.


bfollington:
@jsantell FYI this is a (very small) first step before we bring the changes in #841, this allows us to customise the bundle identifier immediately giving us sandboxing between development and release.

jsantell (approved):

Issue #874, opened 2023-08-15
Profile tabs should use same weight in active and inactive states


gordonbrander:
What happens: font animates between weights creating awkward-looking antialiasing states as the weight animates.

What should happen: visual transition should be relatively lightweight, focusing on color.

Proposed intervention: make tab text always medium weight.

Issue #829, opened 2023-07-25
Profile header sizing is incorrect after editing bio


bfollington:
If your bio increases in length after editing it then the header will grow in size _but_ the tabview below stays full height, resulting in clipped content within. Might be a GeometryReader problem.

We ideally want a sticky header that collapses to a minimal version on scroll to preserve screen space. Perhaps we can work around the issue by implementing that.

bfollington:
https://exyte.com/blog/swiftui-tutorial-scaling-header

bfollington:
@justinabrahms found an additional wonky behaviour:

> note for later: display jitter bug when you have a bio, then click into a follower then go back.

It's not just editing the bio, seemingly _having_ one causes strange layout jitter.

PR #895, opened 2023-09-15
Rework profile layout 

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/829
Fixes https://github.com/subconsciousnetwork/subconscious/issues/874

# Demo

https://github.com/subconsciousnetwork/subconscious/assets/5009316/7b2a2bee-0a27-4064-b478-c38bd9be0ef8

# Changes
- Entire profile is wrapped in `ScrollView`, no sticky header
    - In the fullness of time we will re-introduce this idea via UIKit
- Profile tabs do not animation font-weight between selected and default appearance
- Remove `TabView` and replace with custom implementation
    - We lose out on gestures to swipe between tabs but not a huge loss
- Re-style profile note list to match `Feed` style
- Changing bio no-longer leads to stuttering animation + broken profile layout 

# Tasks
- [x] testing
- [x] max height?
- [x] tidy code

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Layout/TabbedColumnView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Layout/TabbedColumnView.swift
@@ -35,20 +35,19 @@ struct TabbedTwoColumnView<A: View, B: View>: View {
             focusedTabIndex: selectedColumnIndex
         )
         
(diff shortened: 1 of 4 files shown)


gordonbrander (approved):
LGTM

Issue #894, opened 2023-09-15
Notebook view does not refresh after creating a new note


bfollington:
Possibly introduced in https://github.com/subconsciousnetwork/subconscious/pull/885?

1. Tap FAB
2. Create new Public Note
3. Write anything
4. Navigate back (saving the note)
5. Notice that the note is not visible in the notebook root
6. Refresh notebook, note becomes visible

Seems we're not triggering the refresh action correctly (anymore?)

PR #885, opened 2023-09-05
Extract DetailStackView

bfollington:
Porting the work from https://github.com/subconsciousnetwork/subconscious/pull/784, moving towards https://github.com/subconsciousnetwork/subconscious/issues/641 and https://github.com/subconsciousnetwork/subconscious/issues/322.

This PR extracts the `NavigationStack` logic from `NotebookNavigationView` and shifts it to the re-usable component `DetailStackView`. This is a drop-in replacement in terms of the user experience, a stepping stone towards introducing the profile and feed as app tabs (next PRs).

I also snuck a couple fixes to warnings in, sorry if the diff is slightly more confusing as a result.

PR #896, opened 2023-09-17
Replay notebook DetailStack actions on app root

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/894

As of https://github.com/subconsciousnetwork/subconscious/pull/885 are currently swallowing the actions that should refresh the `NotebookView` inside `DetailStack`. I would normally replay the events from `DetailStack`'s store on the `Notebook` store _but_ we (seemingly) can't do this with a `ViewStore`, there's no `actions` publisher exposed.

Instead, I am listening for `.detailStack()` actions in the `NotebookView` and replaying those on the `App` store and _then_ replaying those actions on the `Notebook` store. 

It fixes the problem but feels janky to me, can you think of a better approach @gordonbrander?

### DIFF
--- a/xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
+++ b/xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
@@ -253,13 +253,6 @@ extension AppAction {
         switch action {
         case let .requestDeleteMemo(address):
             return .deleteMemo(address)
-        case let .detailStack(action):
-            switch (action) {
-            case .succeedDeleteMemo(let address):
-                return .succeedDeleteMemo(address)
-            case _:
-                return nil
-            }
         default:
             return nil
         }
@@ -566,8 +559,15 @@ struct NotebookModel: ModelProtocol {
         action: DetailStackAction
     ) -> Update<NotebookModel> {
         switch action {
-        case .succeedMergeMemo, .succeedUpdateAudience, .succeedMoveMemo, .succeedSaveMemo:
-            return update(state: state, action: .refreshLists, environment: environment)
+        case .succeedMergeMemo,
+                .succeedUpdateAudience,
+                .succeedMoveMemo,
+                .succeedSaveMemo:
+            return update(
+                state: state,
+                action: .refreshLists,
+                environment: environment
+            )
         case _:
             return NotebookDetailStackCursor.update(
                 state: state,
--- a/xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
+++ b/xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
@@ -236,14 +236,6 @@ extension NotebookAction {
             return .ready
         case .succeedIndexOurSphere(_):
             return .refreshLists
-        case .succeedSaveMemo:
-            return .refreshLists
-        case .succeedUpdateAudience:
-            return .refreshLists
-        case .succeedMoveMemo:
-            return .refreshLists
-        case .succeedMergeMemo:
-            return .refreshLists
         case let .succeedDeleteMemo(address):
             return .succeedDeleteMemo(address)
         case let .failDeleteMemo(error):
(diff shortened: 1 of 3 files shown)


gordonbrander:
@bfollington in the case of ViewStore, I believe we should be able to handle all communication through tagging/unwrapping actions. If I understand the problem correctly, I think we can use the cursor to do this communication:

The reason we have an actions publisher on Store, and replay those actions across stores is that stores are each separate sources of truth, so we need to replay messages across them to have any kind of state synchronization between them.

OTOH, ViewStore operates on the same principle as Binding. It's not a source of truth. It uses closures to read from a source of truth that exists elsewhere. When we read the state of DetailStack's ViewStore, we're actually reading the state directly from Notebook's `Store` through a closure. When we send an action to DetailStack's ViewStore, we're actually wrapping it with a closure, and then passing it directly to Notebook's store.

Because ViewStore is just a view over a store, we don't need to replay actions from it to the parent store, we can use the cursor code to reroute communication between parent and child.

If I understand correctly, DetailStack has some local actions (succeedSaveMemo, etc) we want handled by the parent component (Notebook/Feed). The ViewStore way to do this would be to handle these actions in DetailStack's update function, but make them no-ops. Then, in the parent's cursor, use a switch statement in the action tagging function to intercept certain child actions, and transform them into the equivalent parent actions. You can then handle these parent-level actions in the parent's update function.

bfollington:
@gordonbrander I've reworked the approach inline with your comment. The reason I didn't do this originally is that I can't see how to intercept / react to these actions _and_ have them pass through the `NotebookDetailStackCursor` (in `NotebookModel.detailStack`). Ideally I would like to still allow the inner `DetailModel` to handle these actions while also dispatching `.refreshLists`

gordonbrander:
That totally makes sense. The way to accomplish this with cursor is:

- in `tag`, intercept the relevant child action, and replace it with a corresponding parent action
- in the parent `update`, handle the parent action, AND in that arm, ALSO pass down the child-level action

Example:

```swift
enum ChildAction {
    case inc
}

struct ChildModel: Model {
    var count = 0

    static func update(state: Self, action: ChildAction, environment: Void) -> Update<Self> {
        switch action {
        case .inc:
            var model = state
            model.count = model.count + 1
            return Update(state: model)
        }
    }
}

enum ParentAction {
    case child(ChildAction)
    case inc
}

struct ParentChildCursor: CursorProtocol {
    func get(state: ParentModel) -> ChildModel {
        state.child
    }

    func set(state: ParentModel, inner: ChildModel) -> ChildModel {
        var model = state
        model.child = inner
        return model
    }

    func tag(_ action: ChildAction) -> ParentAction {
        switch action {
        case .inc:
            return .inc
        }
    }
}

struct ParentModel: Model {
    var count = 0
    var child = ChildModel()

    static func update(state: Self, action: ParentAction, environment: Void) -> Update<Self> {
        switch action {
        case .child(let action):
            return ParentChildCursor.update(
                state: state,
                action: action,
                environment: environment
            )
        case .inc:
            var model = state
            model.count = model.count + 1
            return update(
                state: model,
                action: .child(.inc),
                environment: environment
            )
        }
    }
}
```

gordonbrander:
@bfollington LMK if you want to further edit the PR in response to above

bfollington:
@gordonbrander updated using the approach above, nice. It's clean.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -221,7 +221,18 @@ struct NotebookDetailStackCursor: CursorProtocol {
...
+        switch action {
+        case let .succeedMergeMemo(parent: parent, child: child):
+            return .succeedMergeEntry(parent: parent, child: child)
+        case let .succeedMoveMemo(from: from, to: to):
+            return .succeedMoveEntry(from: from, to: to)
+        case let .succeedUpdateAudience(receipt):
+            return .succeedUpdateAudience(receipt)
+        case let .succeedSaveMemo(address: address, modified: modified):
+            return .succeedSaveEntry(slug: address, modified: modified)
+        case _:
```
gordonbrander, line 233: Nit: prefer `default`

PR #906, opened 2023-09-27
Refactor logging to avoid leaking sensitive data

### LINKED ISSUE #905: Recovery phrase leaks through input field actions
What happens: recovery phrase leaks through input field actions. These are logged at debug level.

What should happen: recovery phrases should not leak.

## Ideas for mitigating

- Create a special recovery phrase input field
- Override description protocol for action to prevent text from leaking out
- Other?

## Example

Screenshot 2023-09-27 at 11 36 38

gordonbrander:
Fixes #905.

Refactors logging to make it easier to avoid logging sensitive data.

Changes:

- Remove `CustomLogStringConvertible`. This was too fancy, and had fallbacks which made it difficult to reason about where and how to redact.
    - Replace all calls to `String.loggable()` with `String(describing:)`.
    - Use default string representation for actions in more places.
    - Relatedly, give actions their own loggers. Actions will be chattier. Giving them a separate logger makes for easier filtering.
- Introduce `@Redacted` property decorator. Apply to any property to implement `CustomStringConvertible` and `CustomDebugStringConvertible` to return the literal string "--redacted"
- Refactor `RecoveryPhrase` to use `@Redacted`
    - Previously this was redacting `debugDescription`, but not `description`. This change redacts both, so you have to be very specific that you want the string value of the mnemonic by referencing the property directly. 
- Update `FormField` to redact all strings for value actions. We should not be logging user's input for most fields anyway, and validated form fields are most likely to contain sensitive information.
- Update Noosphere and NoosphereService to receive mnemonic as String (this matches other parts of the API)


xcode/Subconscious/Shared/Components/FirstRun/RecoveryPhraseView.swift
```diff
@@ -43,11 +43,11 @@ struct RecoveryPhraseView: View {
 }
 
 enum RecoveryPhraseAction: Hashable {
-    case setPhrase(_ phrase: String)
+    case setPhrase(_ phrase: RecoveryPhrase?)
```
gordonbrander, line 46: Use RecoveryPhrase in action to prevent accidentally leaking string via action logs.

xcode/Subconscious/Shared/Components/FirstRun/RecoveryPhraseView.swift
```diff
@@ -43,11 +43,11 @@ struct RecoveryPhraseView: View {
...
 }
 
 struct RecoveryPhraseModel: ModelProtocol {
-    var phrase: String = ""
+    var phrase: RecoveryPhrase?
```
gordonbrander, line 50: Use RecoveryPhrase in model to prevent accidentally leaking if stringifying model.

xcode/Subconscious/Shared/Library/CustomLogStringConvertible.swift
gordonbrander, line 1: Removing this protocol. The way it worked was too clever by half. If `logDescription` was implemented, it would use that, otherwise it would use `debugDescription` or `description` or the magic `String(describing:)`.

This is too many code paths to think about when redacting.

- We remove it
- Standardize on logging actions through `String(describing:)`
- Use `@Redact` to implement redaction for both `String(describing:)` and `String(reflecting:)` for good measure

xcode/Subconscious/Shared/Models/FormField.swift
```diff
@@ -15,13 +15,31 @@ typealias FormFieldValidator<Input, Output> = (Input) -> Output?
...
+    var description: String {
+        switch self {
+        case .reset:
+            return "reset"
+        case .markAsTouched:
+            return "markAsTouched"
+        case .focusChange(let focused):
+            return "focusChange(\(focused)"
+        case .setValue:
+            return "setValue(--redacted--)"
```
gordonbrander, line 38: Manually redact `setValue`. It's too hard to get away from strings for validated form fields, since most of our custom types return `nil` for invalid values, and we need the input to support intermediate values.

Validated form fields are more likely than other places to contain sensitive information which could leak through this action, so the easiest thing to do is redact all `setValue()` actions.

xcode/Subconscious/Shared/Models/FormField.swift
```diff
@@ -38,7 +56,7 @@ struct FormField<Input: FormFieldInput, Output>: ModelProtocol {
         )
     }
     
-    var value: Input
+    @Redacted var value: Input
```
gordonbrander, line 59: Redact model value as well, for good measure.

xcode/Subconscious/Shared/Models/RecoveryPhrase.swift
```diff
@@ -11,26 +11,17 @@ import Foundation
 public struct RecoveryPhrase:
     Hashable,
     Equatable,
-    Identifiable,
-    Codable,
-    LosslessStringConvertible,
-    CustomDebugStringConvertible {
```
gordonbrander, line 17: RecoveryPhrase should not be Codable, LosslessStringConvertible, or CustomDebugStringConvertible, since we don't ever want it serialized.

We don't need it to be identifiable either. Hashable and Equatable is enough.

xcode/Subconscious/Shared/Models/RecoveryPhrase.swift
```diff
@@ -11,26 +11,17 @@ import Foundation
...
-    
+    Codable
+{
     private static let recoveryPhraseRegex = /^(?:\w+\s+){23}\w+$/
 
-    public let description: String
-    public let verbatim: String
-    public var id: String { description }
+    // !!!: Mnemonic is secret and should never be logged or persisted
+    @Redacted public private(set) var mnemonic: String
```
gordonbrander, line 19: Create an explicit field for the secret mnemonic and redact it. This ensures you must access the property explicitly.

Calling `String(describing: recoveryPhrase)` will not accidentally reveal the mnemonic.

xcode/Subconscious/Shared/Models/RecoveryPhrase.swift
```diff
@@ -11,26 +11,17 @@ import Foundation
...
 
     public init?(_ description: String) {
         guard description.wholeMatch(of: Self.recoveryPhraseRegex) != nil else {
             return nil
         }
-        self.description = description.lowercased()
-        self.verbatim = description
+        self.mnemonic = description.lowercased()
```
gordonbrander, line 25: We don't need the verbatim value.

xcode/Subconscious/Shared/Services/Noosphere/Noosphere.swift
```diff
@@ -200,13 +200,13 @@ public actor Noosphere {
         .eraseToAnyPublisher()
     }
     
-    func recover(identity: Did, localKeyName: String, mnemonic: RecoveryPhrase) async throws -> Bool {
+    func recover(identity: Did, localKeyName: String, mnemonic: String) async throws -> Bool {
```
gordonbrander, line 203: Update this method to take a string, since that's what all the other Noosphere methods that take a mnemonic take.

This should be fine, since the value of `RecoveryPhrase` is to prevent accidental logging, which isn't a concern in the scope of this method. We would have to unwrap it to a string here anyway.

xcode/Subconscious/SubconsciousTests/Tests_RecoveryPhrase.swift
```diff
@@ -9,31 +9,19 @@ import XCTest
...
-
-        var string: String? {
-            get { _string }
-            set {
-                self._string = newValue
-                self.sets = self.sets + 1
-            }
-        }
-    }
-
```
gordonbrander, line 24: No longer used

bfollington (approved):
LGTM, I checked it out and playtested and couldn't leak anything. I definitely prefer this approach.

xcode/Subconscious/Shared/Models/RecoveryPhrase.swift
bfollington: Nice.

xcode/Subconscious/Shared/Models/FormField.swift
bfollington: Good call, I don't think we need to see the value anyway. It's always reflected on the UI in the context that these actions are generated.

Issue #902, opened 2023-09-27
BAD_ACCESS error on boot when running on device with Address Sanitization


bfollington:
May have been introduced by #892 but not currently sure. Need to do some git history bisection to find where this was introduced, it almost certainly means we're handling memory / pointers subtly wrong somewhere important.

bfollington:
Can't reproduce anymore, so... let's just hope Xcode had a bad day.

PR #907, opened 2023-09-28
Extract stray switch-case bodies to functions

bfollington:
I'm paranoid after experiencing https://github.com/subconsciousnetwork/subconscious/issues/902 so I did a pass over our most complex models and extracted any switch-case bodies to simplify typechecking.


gordonbrander (approved):
LGTM

PR #909, opened 2023-09-28
Retrieve sphere ID from UserDefaults and run recovery check earlier

gordonbrander:
Previously, we were reading the known sphere identity from the SQLite database. This is suboptimal for a couple reasons:

- The database is an index, not a source of truth, and can be thrown away and re-built at any time. 
- The source of truth for the known sphere identity is `AppDefaults.standard.sphereIdentity` (UserDefaults)
    - We store the identity here upon sphere creation
    - UserDefaults persist for the life of the app container
- Relying on the database meant we had to wait until database migration was complete before checking for recovery state.

This PR makes the following changes:

- Read known sphere identity from `AppDefaults.standard.sphereIdentity`
- Run recovery check as early as possible (`.start`)

Additionally, I added some logs to `NoosphereService` to make it clearer when we experience a cache miss due to failure to construct sphere.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -2453,18 +2453,19 @@ struct AppModel: ModelProtocol {
         environment: Environment
     ) -> Update<Self> {
         let fx: Fx<Action> = Future.detached {
-            let noosphereIdentity = try await environment.noosphere.identity()
             // Get any existing sphere identity stored in UserDefaults
-            let userDefaultsIdentity = AppDefaults.standard.sphereIdentity?
-                .toDid()
+            // If none, it's the first run, nothing to recover
+            guard let userDefaultsIdentity = AppDefaults.standard.sphereIdentity?
+                .toDid() else {
+                return AppAction.presentRecoveryMode(false)
+            }
+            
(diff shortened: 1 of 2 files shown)


gordonbrander:
Playtested recovery on simulator successfully.

Steps:

- Create sphere with app client at Noosphere `v0.14.x`
- Redeem invite code
- Make some notes
- Sync
- Update to this branch, with client at Noosphere `v0.15.0`
- Recover

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -2452,9 +2453,19 @@ struct AppModel: ModelProtocol {
...
         let fx: Fx<Action> = Future.detached {
-            let identity = try await environment.noosphere.identity()
-            let did = try environment.database.readOurSphere()?.identity
-            if did != nil && identity != did {
+            // Get any existing sphere identity stored in UserDefaults
+            // If none, it's the first run, nothing to recover
+            guard let userDefaultsIdentity = AppDefaults.standard.sphereIdentity?
+                .toDid() else {
+                return AppAction.presentRecoveryMode(false)
+            }
```
bfollington, line 2461: Check the persisted DID first, if there isn't one we should skip recovery mode.

bfollington (approved):

PR #912, opened 2023-09-28
Refresh Feed + Profile view after successful recovery

bfollington:
Ensures the feed is populated when dismissing recovery


xcode/Subconscious/Shared/Components/Feed.swift
```diff
@@ -288,9 +305,48 @@ struct FeedModel: ModelProtocol {
...
+            return detailStackNotification(
+                state: state,
+                environment: environment,
+                action: .succeedSaveEntry(address: address, modified: modified)
+            )
         }
     }
-
+    
+    static func detailStackNotification(
```
gordonbrander, line 335: This seems somewhat surprisingly named to me. AFAICT, it's not sending up a notification, but forwarding down an intercepted action?

It's also somewhat unusual to have an update fn that takes an action from another context.

I think it might be more straightforward to have these update arms inlined, or given separate static methods. I think that might make them easier to test, or to handle behavior differently in future.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -497,7 +497,11 @@ struct NotebookModel: ModelProtocol {
         state: NotebookModel,
         environment: AppEnvironment
     ) -> Update<NotebookModel> {
-        return Update(state: state)
+        return update(
+            state: state,
+            action: .refreshLists,
```
gordonbrander, line 502: We don't want to send this until `.ready`, since database will not be ready yet. We already do send this during `.ready`.

Do you recall why this was added?

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -679,190 +683,131 @@ struct NotebookModel: ModelProtocol {
...
+    static func succeedMoveEntry(
+        state: NotebookModel,
+        environment: AppEnvironment,
+        from: Slashlink,
+        to: Slashlink
+    ) -> Update<NotebookModel> {
+        return update(
+            state: state,
+            actions: [
+                .detailStack(.succeedMoveEntry(from: from, to: to)),
```
gordonbrander, line 772: If I understand correctly, we removed the update/setDetails(details) logic from this method because it is already handled by `.detailStack(.succeedMoveEntry(from: from, to: to))`.

Is that correct?

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -679,190 +683,131 @@ struct NotebookModel: ModelProtocol {
...
+    static func succeedMergeEntry(
+        state: NotebookModel,
+        environment: AppEnvironment,
+        parent: Slashlink,
+        child: Slashlink
+    ) -> Update<NotebookModel> {
+        return update(
+            state: state,
+            actions: [
+                .detailStack(.succeedMergeEntry(parent: parent, child: child)),
```
gordonbrander, line 788: Ditto question above.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
```diff
@@ -679,190 +683,131 @@ struct NotebookModel: ModelProtocol {
...
+    /// Updates UI in response.
+    static func succeedUpdateAudience(
+        state: NotebookModel,
+        environment: AppEnvironment,
+        receipt: MoveReceipt
+    ) -> Update<NotebookModel> {
+        return update(
+            state: state,
+            actions: [
+                .detailStack(.succeedUpdateAudience(receipt)),
```
gordonbrander, line 805: Ditto question above

gordonbrander (changes requested):
A few questions.

It may be worth adding tests for DetailStack move/merge/audience update functions, since the logic involved in these is easy to get wrong.

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
bfollington: Yes, we need to refresh the notebook when the tab is presented by switching to it and this is the action dispatched when the view is presented. 

We can add an if check to skip the refresh if the DB isn't ready yet?

gordonbrander: @bfollington that makes sense. Yeah, I think we should add that check. We have an `isDatabaseReady` field on `Notebook` to flag this.

xcode/Subconscious/Shared/Components/Feed.swift
bfollington: I've inlined the `update` calls, but I'm not sure that this makes sense to me:

> It's also somewhat unusual to have an update fn that takes an action from another context.

It's just extracting common functionality in a way that's almost identical to a cursor's update function, perhaps if it were named `forwardWithRefresh` it wouldn't seem so odd? Given the footgun opportunity to omit `.detailStack()` around the forwarded actions I thought _one_ mapping site was safer & more testable (I made this mistake during the refactor).

I suppose I don't see a reason to repeatedly write the same code 4 times upfront, you can always branch one out if/when it's called for.

bfollington:
@gordonbrander I've added tests for `DetailStack` for:

- update audience
- delete
- move
- merge

xcode/Subconscious/Shared/Components/Feed.swift
gordonbrander: Fair point. If you like, we can leave as-was, but I would prefer a different name, since notification is used elsewhere to mean "sending a message up into another parent context we don't know anything about"

xcode/Subconscious/Shared/Components/Notebook/Notebook.swift
bfollington: Done!

gordonbrander (approved):
LGTM!

PR #913, opened 2023-09-29
Prevent accidental double-recovery

bfollington:


xcode/Subconscious/Shared/Components/Common/Recovery/RecoveryModeFormPanelView.swift
```diff
@@ -75,8 +75,10 @@ struct RecoveryModeFormPanelView: View {
             
             Button(
                 action: {
+                    // Dismiss recovery if already succeeded
                     if store.state.recoveryStatus == .succeeded {
                         store.send(.requestPresent(false))
+                        return
```
gordonbrander, line 81: LOL 🤦 

One way we could be more sure we aren't hitting cases like this is to factor this code out into an update function. Say `.recoveryButtonPressed`. In the `recoveryButtonPressed` update function we could perform this same logic.

The advantage would be that we could create a unit test for the update logic to make very sure we don't run recovery twice.

gordonbrander (approved):
Approved, but with a suggestion to improve further.

xcode/Subconscious/Shared/Components/Common/Recovery/RecoveryModeFormPanelView.swift
gordonbrander: I've created a PR to implement this suggestion, along with tests https://github.com/subconsciousnetwork/subconscious/pull/917

bfollington: Merged your suggestions into my PR, definitely a better approach.

PR #915, opened 2023-09-29
Fix merging notes

bfollington:
Working on https://github.com/subconsciousnetwork/subconscious/pull/912 I realised that merging notes was broken. 

The addresses for merge suggestions were not being relativized when retrieved from SQLite, so they had a `peer` of `.did()` which made `writeMemo` blow up thinking it couldn't write to that path.

I've changed `searchRenameSuggestions` to relativize addresses.


gordonbrander (approved):
LGTM

Would be worthwhile to add a test as well.

PR #922, opened 2023-10-03
Integrate NoosphereService with Sentry logging

gordonbrander:
Log errors generated by the FFI wrappers to Sentry.

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -57,13 +57,10 @@ actor NoosphereService:
     SpherePublisherProtocol,
     NoosphereServiceProtocol
 {
-    /// Default logger for NoosphereService instances.
-    private static let logger = Logger(
+    private var logger = Logger(
         subsystem: Config.default.rdns,
         category: "NoosphereService"
     )
-
-    private var logger: Logger
     var globalStorageURL: URL
     var sphereStorageURL: URL
     var gatewayURL: GatewayURL?
@@ -74,6 +71,7 @@ actor NoosphereService:
     private var _sphereIdentity: String?
     /// Memoized Sphere instance
     private var _sphere: Sphere?
+    private var errorLoggingService: any ErrorLoggingServiceProtocol
     
     init(
         globalStorageURL: URL,
@@ -81,7 +79,7 @@ actor NoosphereService:
         gatewayURL: GatewayURL? = nil,
         sphereIdentity: String? = nil,
         noosphereLogLevel: Noosphere.NoosphereLogLevel = .basic,
-        logger: Logger = logger
+        errorLoggingService: any ErrorLoggingServiceProtocol
     ) {
         logger.debug(
             "init NoosphereService",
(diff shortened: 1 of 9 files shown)


xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -74,14 +71,15 @@ actor NoosphereService:
     private var _sphereIdentity: String?
     /// Memoized Sphere instance
     private var _sphere: Sphere?
+    private var errorLoggingService: any ErrorLoggingServiceProtocol
```
gordonbrander, line 74: We don't need to know about Sentry, just that we have an error logging service.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -96,22 +94,24 @@ actor NoosphereService:
...
     }
     
     /// Create a default sphere for user and persist sphere details
     /// This creates, but does not save the sphere as default.
     /// - Returns: SphereReceipt
     func createSphere(ownerKeyName: String) async throws -> SphereReceipt {
-        try await self.noosphere().createSphere(
-            ownerKeyName: ownerKeyName
-        )
+        try await errorLoggingService.capturing {
```
gordonbrander, line 105: In all public methods that call to Noosphere or Sphere FFI wrappers, we wrap those calls with `errorLoggingService.capturing`. This will log any errors that happen as a side-effect, and then re-throw the error.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -96,22 +94,24 @@ actor NoosphereService:
...
+            try await noosphere().createSphere(
+                ownerKeyName: ownerKeyName
+            )
+        }
     }
     
     /// Set a new default sphere
     func resetSphere(_ identity: String?) {
-        logger.debug("Reset sphere identity: \(identity ?? "none")")
+        logger.log("Reset sphere identity: \(identity ?? "none")")
```
gordonbrander, line 114: Increasing log level to log in a few places.

xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -164,113 +164,110 @@ actor NoosphereService:
...
+        }
     }
     
     nonisolated func versionPublisher() -> AnyPublisher<Cid, Error> {
-        self.spherePublisher().flatMap({ sphere in
-            sphere.versionPublisher()
-        })
+        Future {
+            try await self.version()
+        }
```
gordonbrander, line 199: Use our Future helper to expose publishers. This lets us call methods from `NoosphereService`, rather than `Sphere` or `Noosphere`, ensuring we get the error logging.

xcode/Subconscious/SubconsciousTests/Tests_DataService.swift
```diff
@@ -133,7 +133,8 @@ final class Tests_DataService: XCTestCase {
             let noosphere = NoosphereService(
                 globalStorageURL: globalStorageURL,
                 sphereStorageURL: sphereStorageURL,
-                gatewayURL: GatewayURL("http://unavailable-gateway.fakewebsite")
+                gatewayURL: GatewayURL("http://unavailable-gateway.fakewebsite"),
+                errorLoggingService: SentryIntegration()
```
bfollington, line 137: Will this result in sentry errors being logged during testing? I believe sentry is globally disabled during testing but maybe we should stub out a dummy integration here?

bfollington (approved):
Approved with a question

xcode/Subconscious/SubconsciousTests/Tests_DataService.swift
gordonbrander: Good catch! Yes, we should.

PR #923, opened 2023-10-04
Persist selected `AppTab` in `UserDefaults`

bfollington:
Tested on simulator and confirmed it works. 

I think the app should actually default to the `.notebook` view on first run, because the feed is always empty.


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -574,7 +574,7 @@ struct AppModel: ModelProtocol {
         self.databaseMigrationStatus == .succeeded
     }
     
-    var selectedAppTab: AppTab = .feed
+    var selectedAppTab: AppTab = .notebook
```
bfollington, line 577: I think the app should actually default to the .notebook view on first run, because the feed is always empty.

gordonbrander (approved):
LGTM!

Issue #924, opened 2023-10-04
Improve profile loading performance & offline behaviour


bfollington:
After chatting with @gordonbrander we've spotted a few nasty app behaviours on the profile view that occur when network content is unreachable. 

We should be able to render a (possibly stale) profile view for ourselves and any user we follow without touching the network. 

- Index followed (and own) user profiles in DB (#813)
    - This would make the `UserProfileService` cache redundant
- Render our profile using purely cached data
- Render followed users’ profiles with purely cached data
    - Online / offline indicator on profile?
    - Figma mockups have an "online" dot in the omnibar we could use
- Render followed list as just names at first, then we if we have the profiles cached, render full version (otherwise wait for indexing)
- Tapping on a 2nd degree+ peer -> actually hit the network
    - Do not preview their bio & nickname
- Profile-specific not found state (with header)
    - Retry button on failed-to-load states

PR #925, opened 2023-10-05
Fetch basic user metadata from SQLite if available

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/924

This turned into quite a meaty refactor. The goal is to avoid hitting the network to produce the feed or our local profile view. To accomplish this we had to revisit our entire approach to loading and caching user profiles. While a lot of code changed, the new approach is much simpler: we have two code paths, one for loading user data for list views and another for the profile detail view.

I've tested loading the feed with several hundred posts in it from various peers and it loads instantly for me without reading any of the peer's `_profile_` memos.

# Changes
- Introduce `UserProfileService.identifyUser` which consults the DB to produce a `UserProfile` and never traverses (intended for list views which need user metadata)
    - If we don't follow this user or their profile is missing from the DB we fall back to a "shallow" profile display
- Remove `UserProfileService.buildUserProfile`
- Roll `buildUserProfile` into `UserProfileService.loadFullProfileData` which _always_ fetches the latest profile from noosphere (intended for the profile detail view)
    - We do not fetch profiles for any users in the following list _unless_ we follow them and have their profile indexed in the DB
- Consolidate and fix `UserProfile.aliases` logic
- `EntryStub` now comes with a `did` field

# TODO
- [x] explicitly read from memo for the actual users profile, fall back to DB
- [x] when listing users if _we_ follow them, check the DB for their profile, fallback to nothing
- [x] stop passing user profiles around everywhere
- [x] do a join to list the feed?
- [x] show AKA blocks on followed users

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
@@ -8,9 +8,8 @@
 import SwiftUI
 
 public enum NameVariant {
-    case petname(Slashlink, Petname.Name)
-    case proposedName(Slashlink, Petname.Name)
-    case selfNickname(Slashlink, Petname.Name)
+    case known(Slashlink, Petname.Name)
+    case unknown(Slashlink, Petname.Name)
 }
 
 extension UserProfile {
@@ -22,13 +21,13 @@ extension UserProfile {
             self.address.petname?.leaf
         ) {
         case (.ourself, _, .some(let selfNickname), _):
-            return NameVariant.petname(Slashlink.ourProfile, selfNickname)
+            return NameVariant.known(Slashlink.ourProfile, selfNickname)
         case (_, .following(let petname), _, _):
-            return NameVariant.petname(self.address, petname)
+            return NameVariant.known(self.address, petname)
         case (_, .notFollowing, .some(let selfNickname), _):
-            return NameVariant.selfNickname(self.address, selfNickname)
+            return NameVariant.unknown(self.address, selfNickname)
         case (_, .notFollowing, _, .some(let proposedName)):
-            return NameVariant.proposedName(self.address, proposedName)
+            return NameVariant.unknown(self.address, proposedName)
         case _:
             return nil
         }
@@ -68,7 +67,7 @@ struct PetnameView: View {
     
     var uniqueAliases: [Petname] {
         switch name {
-        case .petname(_, let name):
+        case .known(_, let name):
             return aliases.filter { alias in
                 name != alias.leaf
             }
@@ -80,7 +79,7 @@ struct PetnameView: View {
     var body: some View {
         VStack(alignment: .leading, spacing: AppTheme.unit) {
             switch name {
-            case .petname(_, let name):
+            case .known(_, let name):
                 Text(name.toPetname().markup)
                     .fontWeight(.medium)
                     .foregroundColor(.accentColor)
@@ -88,8 +87,7 @@ struct PetnameView: View {
                 if !uniqueAliases.isEmpty {
                     AliasView(aliases: uniqueAliases)
                 }
-            case .selfNickname(let address, let name),
-                 .proposedName(let address, let name):
+            case .unknown(let address, let name):
                 Text(showMaybePrefix
                      ? "Maybe: \(name.description)"
                      : name.description
@@ -104,32 +102,24 @@ struct PetnameView: View {
         }
     }
     
-    
-    
     struct PetnameView_Previews: PreviewProvider {
         static var previews: some View {
             VStack {
                 PetnameView(
-                    name: .selfNickname(
-                        Slashlink(petname: Petname("melville.bobby.tables")!),
-                        Petname.Name("melville")!
-                    )
-                )
-                PetnameView(
-                    name: .proposedName(
+                    name: .unknown(
                         Slashlink(petname: Petname("melville.bobby.tables")!),
                         Petname.Name("melville")!
                     ),
                     showMaybePrefix: true
                 )
                 PetnameView(
-                    name: .petname(
+                    name: .known(
                         Slashlink(petname: Petname("robert")!),
                         Petname.Name("robert")!
                     )
                 )
                 PetnameView(
-                    name: .petname(
+                    name: .known(
                         Slashlink(petname: Petname("robert")!),
                         Petname.Name("robert")!
                     ),
--- a/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
@@ -61,11 +61,12 @@ struct FollowTabView: View {
         ForEach(state.following) { follow in
             StoryUserView(
                 story: follow,
-                action: { _ in
-                    onNavigateToUser(Slashlink(petname: follow.entry.petname)) },
+                onNavigate: {
+                    onNavigateToUser(follow.user.address)
+                },
                 profileAction: onProfileAction,
                 onRefreshUser: {
-                    send(.requestWaitForFollowedUserResolution(follow.entry.petname))
+                    send(.requestWaitForFollowedUserResolution(follow.user.entry.petname))
                 }
             )
         }
--- a/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
@@ -54,21 +54,18 @@ struct RecentTabView: View {
 struct FollowTabView: View {
     var state: UserProfileDetailModel
     var send: (UserProfileDetailAction) -> Void
-    var onNavigateToUser: (UserProfile) -> Void
+    var onNavigateToUser: (Slashlink) -> Void
     var onProfileAction: (UserProfile, UserProfileAction) -> Void
     
     var body: some View {
         ForEach(state.following) { follow in
             StoryUserView(
                 story: follow,
-                action: { _ in onNavigateToUser(follow.user) },
+                action: { _ in
+                    onNavigateToUser(Slashlink(petname: follow.entry.petname)) },
                 profileAction: onProfileAction,
                 onRefreshUser: {
-                    guard let petname = follow.user.address.toPetname() else {
-                        return
-                    }
-                    
-                    send(.requestWaitForFollowedUserResolution(petname))
+                    send(.requestWaitForFollowedUserResolution(follow.entry.petname))
                 }
             )
         }
@@ -93,7 +90,7 @@ struct UserProfileView: View {
     }
     
     var onNavigateToNote: (Slashlink) -> Void
-    var onNavigateToUser: (UserProfile) -> Void
+    var onNavigateToUser: (Slashlink) -> Void
     
     var onProfileAction: (UserProfile, UserProfileAction) -> Void
     var onRefresh: () async -> Void
--- a/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
@@ -66,7 +66,7 @@ struct FollowTabView: View {
                 },
                 profileAction: onProfileAction,
                 onRefreshUser: {
-                    send(.requestWaitForFollowedUserResolution(follow.user.entry.petname))
+                    send(.requestWaitForFollowedUserResolution(follow.entry.petname))
                 }
             )
         }
--- a/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
@@ -31,16 +31,13 @@ struct RecentTabView: View {
     var onNavigateToNote: (Slashlink) -> Void
     
     var body: some View {
-        if let user = state.user {
-            ForEach(state.recentEntries) { entry in
-                StoryEntryView(
-                    story: StoryEntry(
-                        author: user,
-                        entry: entry
-                    ),
-                    action: { address, _ in onNavigateToNote(address) }
-                )
-            }
+        ForEach(state.recentEntries) { entry in
+            StoryEntryView(
+                story: StoryEntry(
+                    entry: entry
+                ),
+                action: { address, _ in onNavigateToNote(address) }
+            )
         }
         
         if state.recentEntries.count == 0 {
--- a/xcode/Subconscious/Shared/Components/Common/Story/StoryEntryView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Story/StoryEntryView.swift
@@ -84,7 +84,8 @@ struct StoryPlainView_Previews: PreviewProvider {
                             But do we have to go to distant worlds to find other kinds of replicator and other, consequent, kinds of evolution? I think that a new kind of replicator has recently emerged on this very planet. It is staring us in the face.
                             """
                         )
-                    )
+                    ),
+                    did: Did.dummyData()
                 )
             ),
             action: { _, _ in }
--- a/xcode/Subconscious/Shared/Components/Common/Story/StoryEntryView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Story/StoryEntryView.swift
@@ -27,11 +27,8 @@ struct StoryEntryView: View {
                     
                     HStack(alignment: .center, spacing: AppTheme.unit) {
                         BylineSmView(
-                            pfp: story.author.pfp,
-                            slashlink: Slashlink(
-                                peer: story.author.address.peer,
-                                slug: story.entry.address.slug
-                            )
+                            pfp: .generated(story.entry.did),
+                            slashlink: story.entry.address
                         )
                         
                         Spacer()
@@ -68,7 +65,6 @@ struct StoryPlainView_Previews: PreviewProvider {
     static var previews: some View {
         StoryEntryView(
             story: StoryEntry(
-                author: UserProfile.dummyData(),
                 entry: EntryStub(
                     MemoEntry(
                         address: Slashlink("@here/meme")!,
--- a/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
@@ -57,7 +57,7 @@ struct StoryUserView: View {
                     
                     Spacer()
                     
-                    switch story.user.resolutionStatus {
+                    switch story.addressBookEntry.status {
                     case .unresolved:
                         Image(systemName: "person.fill.questionmark")
                             .foregroundColor(.secondary)
@@ -77,7 +77,7 @@ struct StoryUserView: View {
                         }
                     }
                 }
-                .disabled(!story.user.resolutionStatus.isReady)
+                .disabled(!story.addressBookEntry.status.isReady)
                 
                 Menu(
                     content: {
@@ -134,7 +134,7 @@ struct StoryUserView: View {
         }
         .contentShape(.interaction, RectangleCroppedTopRightCorner())
         .onTapGesture {
-            switch (story.user.ourFollowStatus, story.user.resolutionStatus) {
+            switch (story.user.ourFollowStatus, story.addressBookEntry.status) {
             case (.following(_), .unresolved):
                 onRefreshUser()
             case (.following(let name), _):
@@ -160,9 +160,14 @@ struct StoryUserView_Previews: PreviewProvider {
                         pfp: .image("pfp-dog"),
                         bio: UserProfileBio("Ploofy snooflewhumps burbled, outflonking the zibber-zabber."),
                         category: .human,
-                        resolutionStatus: .unresolved,
                         ourFollowStatus: .notFollowing,
                         aliases: []
+                    ),
+                    addressBookEntry: AddressBookEntry(
+                        petname: Petname("ben")!,
+                        did: Did("did:key:123")!,
+                        status: .resolved("ok"),
+                        version: "ok"
                     )
                 ),
                 action: { _ in }
@@ -176,9 +181,14 @@ struct StoryUserView_Previews: PreviewProvider {
                         pfp: .image("pfp-dog"),
                         bio: UserProfileBio("Ploofy snooflewhumps burbled, outflonking the zibber-zabber."),
                         category: .human,
-                        resolutionStatus: .pending,
                         ourFollowStatus: .following(Petname.Name("lol")!),
                         aliases: []
+                    ),
+                    addressBookEntry: AddressBookEntry(
+                        petname: Petname("ben")!,
+                        did: Did("did:key:123")!,
+                        status: .resolved("ok"),
+                        version: "ok"
                     )
                 ),
                 action: { _ in }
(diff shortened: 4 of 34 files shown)


xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -345,60 +375,56 @@ actor UserProfileService {
...
-            guard let petname = address.petname else {
-                let sphere = try await self.noosphere.sphere(address: address)
-                let cid = try await sphere.version()
-                return .resolved(cid)
-            }
-            
-            return try await self.addressBook.resolutionStatus(petname: petname)
+        var identity = did
+        if identity == nil {
+            identity = try await self.noosphere.identity()
```
bfollington, line 413: doesn't make a lot of sense to me, I think we want resolve here, but this code might go away anyway

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
```diff
@@ -22,13 +21,13 @@ extension UserProfile {
...
+            return NameVariant.known(Slashlink.ourProfile, selfNickname)
         case (_, .following(let petname), _, _):
-            return NameVariant.petname(self.address, petname)
+            return NameVariant.known(self.address, petname)
         case (_, .notFollowing, .some(let selfNickname), _):
-            return NameVariant.selfNickname(self.address, selfNickname)
+            return NameVariant.unknown(self.address, selfNickname)
         case (_, .notFollowing, _, .some(let proposedName)):
-            return NameVariant.proposedName(self.address, proposedName)
+            return NameVariant.unknown(self.address, proposedName)
```
bfollington, line 30: Simplified the modelling for name display, we either know a user or we don't. 

If we know them, we use our name for them. If we don't know them, we use their nickname (if available) or fall back to the leaf of their current address.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -119,7 +119,7 @@ enum DetailStackAction: Hashable {
     case pushDetail(MemoDetailDescription)
 
     case requestOurProfileDetail
-    case pushOurProfileDetail(UserProfile)
```
bfollington, line 122: Since it's cheap to retrieve UserProfile's now I think we should stop passing them around. There was a point where this cut down on load times but not anymore.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -441,15 +439,10 @@ struct DetailStackModel: Hashable, ModelProtocol {
...
-        environment: Environment,
-        user: UserProfile
+        environment: Environment
     ) -> Update<Self> {
         let detail = UserProfileDetailDescription(
             address: Slashlink.ourProfile,
-            user: user,
-            // Focus following list by default
-            // We can already see our recent notes in our notebook so no
-            // point showing it again
```
bfollington, line 452: Put this back

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -643,28 +635,10 @@ extension DetailStackAction {
...
-                    return user.overrideAddress(Slashlink(petname: name.toPetname()))
-                case _:
-                    return user
-                }
-            }
-
-
-            guard user.resolutionStatus.isReady else {
-                return .failPushDetail("Attempted to navigate to unresolved user")
-            }
```
bfollington, line 663: Instead of trying to rewrite the addresses on tap we now calculate the preferred navigation address for a user when initially querying data for the view.

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -561,10 +561,11 @@ struct MemoViewerDetailModel: ModelProtocol {
                         .profile
                 }
                 
+                let did = try await environment.noosphere.resolve(peer: state.address?.peer)
+                
                 return try await environment
                     .userProfile
-                    .requestUserProfile(petname: petname)
```
bfollington, line 566: The viewer was hitting the author's profile memo just to load optional metadata, I've changed it to try and read from the DB.

xcode/Subconscious/Shared/Models/Slashlink.swift
```diff
@@ -211,7 +211,11 @@ extension Slashlink {
     /// Given an address, re-base relative to the passed peer.
     /// If `@alice.bob/test` appears in a note owned by `@donna.charlie`
     /// then we want `@alice.bob.donna.charlie/test`
-    func rebaseIfNeeded(peer: Peer) -> Slashlink {
+    func rebaseIfNeeded(peer: Peer?) -> Slashlink {
+        guard let peer = peer else {
+            return self
+        }
```
bfollington, line 217: This cleans up a lot of noise at the use-site.

xcode/Subconscious/Shared/Models/EntryStub.swift
```diff
@@ -16,6 +16,7 @@ struct EntryStub:
     CustomDebugStringConvertible,
     Codable
 {
+    let did: Did
```
bfollington, line 19: All `EntryStub`s now come with a `did`. This is always available when reading from the DB and means we can trivially work out which user profile should be associated with a given entry.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -220,7 +218,6 @@ struct UserProfile: Equatable, Codable, Hashable {
     let pfp: ProfilePicVariant
     let bio: UserProfileBio?
     let category: UserCategory
-    let resolutionStatus: ResolutionStatus
```
bfollington, line 223: Dropped this field because I felt we were mixing concerns.

`UserProfile`s contain a user's "display" metadata.
`AddressBookEntry`s contain noosphere-level user data like name resolution status.

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -663,13 +663,7 @@ actor DataService {
         let identity = try await noosphere.identity()
         var feed: [EntryStub] = []
         for entry in try self.database.listFeed(owner: identity) {
```
bfollington, line 665: We can go a step further here and actually do a query with a `JOIN` to ensure each `EntryStub` comes with the user's profile data already.

A `JOIN` has the _disadvantage_ of loading a bunch of duplicate data but it would do it all in one sweep. Currently, we maintain an in-memory cache of the `UserProfile`s for a given `(Did, Petname)` combination so while we do have to call `identifyUser` for every item, the calls become cheap quickly and never hit the underlying sphere.

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -663,13 +663,7 @@ actor DataService {
...
-            let author = try await Func.run {
-                if entry.address.isOurs {
-                    return try await self.userProfile.loadOurProfileFromMemo()
-                } else {
-                    return try await self.userProfile.buildUserProfile(address: entry.address)
-                }
-            }
+            let author = try await self.userProfile.identifyUser(did: entry.did, address: entry.address, context: nil)
```
bfollington, line 666: We could go further here and retrieve the user profile as part of `listFeed` using a `JOIN`. 

This would result in a lot of duplicate data in the query results but would resolve in one sweep. I didn't think this was necessary yet because, while we do call `identifyUser` for each entry, the in-memory cache in `UserProfileService` will make these cheap when building the feed.

xcode/Subconscious/Shared/Components/Feed.swift
```diff
@@ -25,7 +25,10 @@ struct FeedNavigationView: View {
         DetailStackView(app: app, store: detailStack) {
             ScrollView {
                 LazyVStack(spacing: 0) {
-                    if let feed = store.state.entries {
+                    switch (store.state.status, store.state.entries) {
```
bfollington, line 28: Introduced an explicit `status` for the feed instead of checking whether `entries` was `nil`. This lets us be smarter about when to show the loading vs empty state.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1003,15 +1032,19 @@ final class DatabaseService {
                 .queryFTS5(link.slug.description),
                 .text(link.id)
             ]
-        ).compactMap({ row in
-            let did = row.col(0)?.toString()?.toDid()
+        ).compactMap({ row -> EntryStub? in
+            guard let did = row.col(0)?.toString()?.toDid() else {
+                return nil
+            }
```
bfollington, line 1038: The `did` column of the `memo_search` table is marked as nullable, unlike the `did` column of `memo`. I'm not sure why, so flagging it in-case I'm missing something.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -112,6 +118,11 @@ struct UserProfileCacheEntry {
     let version: Cid
 }
 
+struct UserProfileCacheKey: Hashable {
+    let did: Did
+    let petname: Petname?
+}
```
bfollington, line 124: The same user with the same `Did` can have different `UserProfile`s depending on the context you encounter them.

A trivial example: following the same `Did` under two `Petname`s changes which name is chosen as the "primary" name, depending on which version you're looking at.

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -663,13 +663,11 @@ actor DataService {
...
-                    return try await self.userProfile.loadOurProfileFromMemo()
-                } else {
-                    return try await self.userProfile.buildUserProfile(address: entry.address)
-                }
-            }
+            let author = try await self.userProfile.identifyUser(
+                did: entry.did,
+                address: entry.address,
+                context: nil
+            )
```
bfollington, line 670: We could go further here and retrieve the user profile as part of listFeed using a JOIN.

This would result in some duplicate data in the query results but would resolve in one sweep. I didn't think this was necessary yet because, while we do call identifyUser for each entry, the in-memory cache in UserProfileService will make these cheap when building the feed.

xcode/Subconscious/Shared/Services/UserProfileService.swift
bfollington: Worth saying: I'm open to the idea of dropping the in-memory cache. The main drawback is that we'd be sending a query to SQLite for each post in the feed except in the case of _our_ profile which is read from our sphere.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -344,73 +356,163 @@ actor UserProfileService {
...
+    func identifyUser(
+        did: Did,
+        address: Slashlink,
+        context: Peer?
+    ) async throws -> UserProfile {
+        switch address.peer {
+        case .petname(let petname):
+            return try await self.identifyUser(did: did, petname: petname, context: context)
+        case .did, .none:
+            return try await self.identifyUser(did: did, petname: nil, context: context)
```
bfollington, line 396: Convenience signatures

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
gordonbrander: I like the clarity, however, I want to note that this loses us information.

It seems likely at some point we will want to be able to distinguish proposed names from self-proposed names (https://github.com/cwebber/rebooting-the-web-of-trust-spring2018/blob/petnames/draft-documents/petnames.md). For example,

- We may want to prefer a self-proposed name to a 3p-proposed name when following and suggesting a name.
- (Speculative) We may wish, in future, to show the user's self-proposed name in context menus or place a badge next to self-proposed names to show user's preferred name.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -54,21 +54,19 @@ struct RecentTabView: View {
...
+                    onNavigateToUser(follow.user.address)
+                },
                 profileAction: onProfileAction,
                 onRefreshUser: {
-                    guard let petname = follow.user.address.toPetname() else {
-                        return
-                    }
-                    
-                    send(.requestWaitForFollowedUserResolution(petname))
+                    send(.requestWaitForFollowedUserResolution(follow.entry.petname))
```
gordonbrander, line 69: Nice. It's a good smell to move guard logic out of event handlers and into update functions.

xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
gordonbrander, line 1: Most of the changes in this file seem unrelated to the PR. Do I have that right? This is fine, I just want to make sure I'm not missing anything.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
gordonbrander: Nice. I like the idea of cutting down on caches that could get stale. SQLite is our cache, and it should be synchronized in lock-step with Noosphere syncs, so it should have the most up-to-date information, yet be fast to read.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -484,6 +485,32 @@ final class DatabaseService {
         .first
     }
     
+    func loadUserProfile(did: Did) throws -> Data? {
```
gordonbrander, line 488: Let's add a test for this

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -541,7 +570,7 @@ final class DatabaseService {
         
         let results = try database.execute(
             sql: """
-            SELECT id, modified, excerpt
+            SELECT did, id, modified, excerpt
```
gordonbrander, line 573: I would like to add a test for `listRecentMemos` that ensures we're getting the data out of the DB correctly.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1048,7 +1083,7 @@ final class DatabaseService {
         
         return try? database.execute(
             sql: """
-            SELECT id, modified, excerpt
+            SELECT did, id, modified, excerpt
```
gordonbrander, line 1086: I would like to add/update a test for readRandomEntryInDateRange to ensure we're getting the data out correctly.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1062,17 +1097,18 @@ final class DatabaseService {
         )
         .compactMap({ row in
             guard
-                let address = row.col(0)?
+                let did = row.col(0)?.toString()?.toDid(),
+                let address = row.col(1)?
                     .toString()?
-                    .toLink()?
```
gordonbrander, line 1067: Why was this step removed?

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -554,17 +583,18 @@ final class DatabaseService {
         )
         return results.compactMap({ row in
             guard
-                let address = row.col(0)?
+                let did = row.col(0)?.toString()?.toDid(),
+                let address = row.col(1)?
                     .toString()?
-                    .toLink()?
```
gordonbrander, line 559: Why was this step removed?

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1130,7 +1168,7 @@ final class DatabaseService {
         
         return try? database.execute(
             sql: """
-            SELECT id, modified, excerpt
+            SELECT did, id, modified, excerpt
```
gordonbrander, line 1171: I would like to update/add a test for readRandomEntryMatching to ensure we're getting the data out correctly.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1143,17 +1181,18 @@ final class DatabaseService {
         )
         .compactMap({ row in
             guard
-                let address = row.col(0)?
+                let did = row.col(0)?.toString()?.toDid(),
+                let address = row.col(1)?
                     .toString()?
-                    .toLink()?
```
gordonbrander, line 1148: Same question as above

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -344,73 +356,163 @@ actor UserProfileService {
...
     }
     
-    func buildUserProfile(
-        address: Slashlink
+    /// Build a UserProfile suitable for list views, transcludes etc.
+    /// This will attempt to read from the database and also maintain an in-memory cache of profiles
+    /// we have encountered.
+    ///
+    /// `context` will be used to determine the preferred navigation address for a user.
+    func identifyUser(
```
gordonbrander, line 405: There's a lot going on in this function. Probably deserves a test.

gordonbrander (changes requested):
On the whole, looks great. Requests for some tests given this touches DB queries, which is error prone.

xcode/Subconscious/Shared/Components/Common/Story/StoryUserView.swift
bfollington: Yep this file only changed because of some data structure changes elsewhere, no intended view changes.

xcode/Subconscious/Shared/Services/UserProfileService.swift
bfollington: Yep! I was considering that in this comment: https://github.com/subconsciousnetwork/subconscious/pull/925#discussion_r1351439703

**But** I realised today that since `EntryStub` now had the `Did` attached, I don't need to load a profile at all for the feed. In fact, I managed to drop `author` from `EntryStub` entirely (I wasn't a fan of having to introduce it in the first place).

Now, we can produce the feed + backlinks + transcludes without having to fetch the user's profile in the process. _If_ we introduce something like customisable PFPs in the future we will have to revisit this but I would argue that that's a whole seperate domain.

As a result, there's really no benefit coming from the in-memory cache so... I removed it. One less invalidation layer to care about.

xcode/Subconscious/Shared/Components/Common/Byline/PetnameView.swift
bfollington: It doesn't actually lose us any information in the model itself, just in the mapping to an (unused) view construct. We were rendering `proposedName` and `selfNickname` identically before this change so nothing visibly changes as a result of this.

If the kind of cases you point out do arise, we can derive all that we need from a `UserProfile` regardless.

xcode/Subconscious/Shared/Services/DataService.swift
bfollington: No need, see: https://github.com/subconsciousnetwork/subconscious/pull/925#discussion_r1353735394

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: This is covered by `Tests_DatabaseService.testListRecentMemos` & `Tests_DatabaseService.testListRecentMemosWithoutOwner` but I've added a few more assertions.

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Good call, added `Tests_DatabaseService.testReadRandomEntryInDateRange`, turns out this query was broken but it's also not in use by anything in the app currently (it was once used by the "On This Day" geist).

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Added `Tests_DatabaseService.testReadRandomEntryMatching`

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Accidental, restored it, but I'm realising I don't fully grasp the importance of this step.

To my eye, the "benefit" of going via `toLink()` first is that we're asserting that there's a DID present in the address, going straight to `toSlashlink()` would allow a peerless address to pass through... but is that actually the job of this code? 

All entries in the `memo` table _should_ have a wellformed `id` already (ideally unit tested on the write-side), and we're throwing away the additional constraint by converting to a `Slashlink` right afterwards.

Maybe the core of my confusion comes down to this: when should we prefer to select `id` vs `slashlink` from the `memo` table if we're ultimately converting _either_ to a `Slashlink` before returning? My guess is this comes down to how `Did.local` is handled somehow, but now that we have `EntryStub.did` it would probably be better to consult _that_ to determine if content is local? It feels brittle to me to be implicitly depending on a particular `Peer` case within a `Slashlink`...

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Restored it, but answered here: https://github.com/subconsciousnetwork/subconscious/pull/925#discussion_r1353825665

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Restored it, but answered here: https://github.com/subconsciousnetwork/subconscious/pull/925#discussion_r1353825665

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Added `Tests_UserProfileService.testReadUserProfile` and renamed this method to `readUserProfile` for consistency with other methods in the file.

xcode/Subconscious/Shared/Services/UserProfileService.swift
bfollington: I had thought this would be too difficult to test without traversal, but I came up with some reasonable test cases that cover the logic within this function itself. 

See `Tests_UserProfileService.testIdentifyUserCallSignatures` and `Tests_UserProfileService.testIdentifyUserRebaseOnContext`

xcode/Subconscious/Shared/Components/Common/Story/StoryEntryView.swift
```diff
@@ -27,11 +27,8 @@ struct StoryEntryView: View {
                     
                     HStack(alignment: .center, spacing: AppTheme.unit) {
                         BylineSmView(
-                            pfp: story.author.pfp,
-                            slashlink: Slashlink(
-                                peer: story.author.address.peer,
-                                slug: story.entry.address.slug
-                            )
+                            pfp: .generated(story.entry.did),
```
bfollington, line 30: This is all we were using `author` for, which can be derived from just the DID.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -143,10 +136,31 @@ actor UserProfileService {
...
+
+            throw UserProfileServiceError.failedToDeserializeProfile(error, string)
+        }
+    }
+    
     /// Attempt to read & deserialize a user `_profile_.json` at the given address.
     /// Because profile data is optional and we expect it will not always be present
     /// any errors are logged & handled and nil will be returned if reading fails.
-    private func readProfileMemo(
+    func readProfileMemo(
```
bfollington, line 163: Public so we can access it in tests.

xcode/Subconscious/SubconsciousTests/TestUtilities.swift
```diff
@@ -56,6 +56,7 @@ struct TestUtilities {
             path: "database.sqlite",
             directoryHint: .notDirectory
         )
+        print("Test SQLite Path: \(databaseURL.path(percentEncoded: false))")
```
bfollington, line 59: Helpful for debugging tests.

gordonbrander (approved):
LGTM. Feel free to land.

Have you play tested? It's late my time, but will playtest myself in the AM.

Issue #927, opened 2023-10-06
When a sphere fails to index, we don't see its did in the logs


gordonbrander:
We should log the did when a fail happens.

Example of a successful index log message (I've truncated the identifiers):

```
2023-10-06 16:00:21.633816-0400 Subconscious[34425:11791299] [AppAction] indexPeer(jordan)
2023-10-06 16:00:21.633835-0400 Subconscious[34425:11791299] [app] Indexing peer petname="jordan"
2023-10-06 16:00:21.645535-0400 Subconscious[34425:11799839] [DataService] Indexing peer petname="jordan" identity="did:key:z6Mk..." version="bafy..." since="bafy..." changes="0"
2023-10-06 16:00:21.646910-0400 Subconscious[34425:11799839] [DataService] Indexed peer petname="jordan" identity="did:key:z6Mk..." version="bafyr..." since="bafy..."
```

Example of a fail index log message:

```
2023-10-06 16:00:21.633710-0400 Subconscious[34425:11791299] [AppAction] indexPeer(burrito)
...
2023-10-06 16:00:54.609963-0400 Subconscious[34425:11791299] [AppAction] failIndexPeer(petname: burrito, error: Subconscious.NoosphereError.foreignError("Failed to parse CAR file: Unexpected cbor code `0x3c` when decoding `libipld_core::ipld::Ipld`."))
2023-10-06 16:00:54.610752-0400 Subconscious[34425:11791299] [app] Failed to index peer petname="burrito" error="Foreign Error: Failed to parse CAR file: Unexpected cbor code `0x3c` when decoding `libipld_core::ipld::Ipld`."
```

PR #930, opened 2023-10-12
Attempt to include the DID if `indexPeer` fails

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/927

We cannot always guarantee that the identity is available (or even exists) but we can make a best-effort to include the identity for debugging purposes.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -213,7 +213,7 @@ enum AppAction {
     /// Index the contents of a sphere in the database
     case indexPeer(_ petname: Petname)
     case succeedIndexPeer(_ peer: PeerRecord)
-    case failIndexPeer(petname: Petname, error: Error)
+    case failIndexPeer(petname: Petname, identity: Did?, error: Error)
 
     /// Purge the contents of a sphere from the database
     case purgePeer(_ did: Did)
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -2003,8 +2004,12 @@ struct AppModel: ModelProtocol {
                 )
                 return Action.succeedIndexPeer(peer)
             } catch {
+                // We may be unable to resolve a peer, but attempt to grab the DID
+                let sphere = try? await environment.noosphere.traverse(petname: petname)
+                let identity = try? await sphere?.identity()
```
gordonbrander, line 2009: @bfollington This would result in a double-traversal in case of failure. This seems problem-prone, especially if the failure was with traversal (two network requests?).

Rather than adding the missing did to this action, I would consider adding information to the logging within `indexPeer`.

For example, if we moved this log below the call to `sphere.identity()`, we should be able to log the identity along with the petname. https://github.com/subconsciousnetwork/subconscious/blob/b4b843519e0356304187cad76f3f5717283800f5/xcode/Subconscious/Shared/Services/DataService.swift#L123-L129

IIRC if we have successfully traversed, then we should have the did available. Is my assumption correct?

While in there, I would consider changing all the logging in indexPeer from `logger.debug` to `logger.log`.

gordonbrander (changes requested):
Suggestion

gordonbrander (approved):
LGTM

Issue #928, opened 2023-10-11
Renaming a peer results in a stale SQLite record (and failed attempts to index)


bfollington:
Reproduction:

1. Follow a user `@test` who you can resolve and has at least one note
2. Let indexing complete
3. Rename them to `@another-test` (using the ... -> Rename function)
    a. This unfollows and refollows under a new name
4. Let indexing complete
5. Check the DB and logs, see that we're trying to index both `@test` and `@another-test`

We will always fail to index `@test` because it's no longer resolvable.

bfollington:
This works correctly when unfollowing, initial investigation makes me think that the way `UserProfileDetailView.attemptRename` calls unfollow then follow right after one another is _somehow_ leading to this behaviour. More digging needed.

PR #931, opened 2023-10-13
Fix stale peer record in SQLite

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/928

These updates to the database interfere with indexing and can result in a desync between the truth and the `peer` table in SQLite. We should let the indexing process be the sole way that the DB is updated.


gordonbrander (approved):
LGTM

PR #932, opened 2023-10-13
Update to Noosphere 0.15.3

bfollington:
Playtesting blocked by stormy network conditions but the tests pass.

### DIFF
--- a/xcode/Subconscious/Subconscious.xcodeproj/project.pbxproj
+++ b/xcode/Subconscious/Subconscious.xcodeproj/project.pbxproj
@@ -2658,7 +2658,7 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/subconsciousnetwork/noosphere";
 			requirement = {
-				branch = "swift-noosphere-v0.15.0";
+				branch = "swift-noosphere-v0.15.3";
 				kind = branch;
 			};
 		};
--- a/xcode/Subconscious/Subconscious.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/xcode/Subconscious/Subconscious.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -14,8 +14,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/subconsciousnetwork/noosphere",
       "state" : {
-        "branch" : "swift-noosphere-v0.15.0",
-        "revision" : "d1fd253414cf6852653baa08ab1074bde7e57538"
+        "branch" : "swift-noosphere-v0.15.3",
+        "revision" : "9e5c83b0d1f8c2ec46e61c604811b7e243b6d385"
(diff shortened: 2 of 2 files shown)


gordonbrander:
We'll need to make sure that sync logic works with the new versioning behavior.

gordonbrander:
> Make sure the sphere you are indexing is at the version you expected when you traversed to it

bfollington:
Plan:

Generate sphere history with a few revisions, noting the `CID`s along the way. Follow that sphere from a second sphere and test what happens if we call `changes(since:)` with an older `CID` than expected.

bfollington:
@cdata I am still not quite connecting the dots on this one. I'll outline how our indexing works and  my understanding of the problem and you can tell me where the gap is. Currently our indexing process looks like:

1. Traverse to peer
2. Read the version from the sphere we traversed to
3. Check SQLite using the identity of this peer
4. If we have seen this sphere before, call `sphere.changes(since:)` with the last version we read, otherwise list all slugs
5. Iterate over the changes, `sphere.read`ing each memo and updating SQLite

As I understand it Noosphere 0.15.3 will sometimes fall back to an older version of a given memo when we call `read` if the blocks for the latest version of that memo cannot be read (or time out). 

I see the potentially problematic scenario as: 

1. We have a peer with three sphere revisions: 1, 2 and 3
2. We index them at version 3
3. Some blocks become unreachable
4. We attempt to re-index and for one memo noosphere falls back to version 2
5. We end up replacing our cached copy (version 3) with version 2

I’m not actually sure how to adjust our approach to avoid this, but I also don’t think it’s strictly wrong to keep the index in sync with whatever noosphere returns?

---

I may be missing something obvious, your previous advice:
> Make sure the sphere you are indexing is at the version you expected when you traversed to it

Is hard for me to conceive of, we don’t know the version we’ll get until we traverse do we? We can know if the version of a sphere matches the last version we indexed, but how can I know what I expect from a traversal?

cdata:
@bfollington it is sufficient to check that you didn't already index the version you got. The problem you want to avoid is indexing a version you already saw, thinking it is some version you haven't seen yet.

You can see the "intended" version (the one that we fail to traverse to in the hypothetical) by using [ns_sphere_petname_resolve](https://github.com/subconsciousnetwork/noosphere/blob/9e5c83b0d1f8c2ec46e61c604811b7e243b6d385/rust/noosphere/src/ffi/petname.rs#L176), which gives you the expected version without traversing (this would not fail in the hypothetical because the data it accesses is part of a sphere you already have).

cdata:
I should note that if you aren't already using `ns_sphere_petname_resolve` in the indexing algorithm, then you probably aren't in danger of accidentally recording that you indexed a version you never saw (because you won't have had an expectation of a specific version until the traversal completes successfully).

gordonbrander (approved):
LGTM

Let's playtest this on-device with some peers before merging.

bfollington:
Tested on device with three peers, all behaving nicely. I'm going to merge this.

PR #934, opened 2023-10-13
Update search to use ViewStore

gordonbrander:
PR updates search view to use a ViewStore. In my playtesting, this significantly improves the perceived perf of type-ahead suggestions.

Why? I suspect this may have something to do with the way inputs work in SwiftUI. Perhaps the binding-style closer access that goes all the way up to store allows the input field to render text BEFORE the search view is invalidated by the parent.

Background:

- I've been bothered by the perceived performance of type-ahead search in the app.
- At first, I suspected this was due to SQLite full text search running up against large amounts of indexed peer content. So I tried searching only post titles instead of full text search.
    - However, upon investigating the app container, it turned out I was searching across relatively low amounts of content
    - SQLite was not the bottleneck.
- I also tried limiting the amount of serializing needed for logging by implementing a `CustomStringConvertible` for `SearchAction`. This wasn't the bottleneck, but I kept it anyway, because IMO its probably a good idea not to serialize and log all suggestions at every keystroke.
- The improvement came with using ViewStore. Surprising, but then again, perhaps not, given the issues we have had with input fields in the past that were fixed by ViewStore.

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/Forms/Search.swift
+++ b/xcode/Subconscious/Shared/Components/Common/Forms/Search.swift
@@ -42,6 +42,41 @@ enum SearchAction: Hashable {
     case entryDeleted(Slug)
 }
 
+extension SearchAction: CustomStringConvertible {
+    var description: String {
+        switch self {
+        case .requestPresent(let bool):
+            return "requestPresent(\(bool))"
+        case .hideAndClearQuery:
+            return "hideAndClearQuery"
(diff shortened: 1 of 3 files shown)


bfollington (approved):
Wow, I'm surprised it makes such a difference but that's good news. In my mind this bumps the priority on refactoring other `state` + `send` to `ViewStore` cases a little... Even if nothing else is showing performance issues yet.

Issue #873, opened 2023-08-15
View-mode transcludes do not horizontally fill space


gordonbrander:
What happens: the transclude in the note body is as wide as its content, and does not fill the available horizontal space. Note that the transclude in the related section does fill the horizontal space.

What should happen: the transclude should fill the available horizontal space.

![7082CC34-619D-4EE9-9F43-7FB730589748](https://github.com/subconsciousnetwork/subconscious/assets/58421/30569aae-5c2d-43f1-99a6-07f51cce6d41)

PR #935, opened 2023-10-13
Set size of spacer in expandAlignedLeading to 0

gordonbrander:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/873

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
+++ b/xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
@@ -13,7 +13,7 @@ struct ExpandAlignLeadingViewModifier: ViewModifier {
     func body(content: Content) -> some View {
         HStack(spacing: 0) {
             content
-            Spacer()
+            Spacer(minLength: 0)
         }
     }
 }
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Components/Common/ExpandAlignedLeadingViewModifier.swift
```diff
@@ -13,7 +13,7 @@ struct ExpandAlignLeadingViewModifier: ViewModifier {
     func body(content: Content) -> some View {
         HStack(spacing: 0) {
             content
-            Spacer()
+            Spacer(minLength: 0)
```
gordonbrander, line 16: It turns out that the spacing problem was not caused by the transclude failing to fill horizontal space. Rather, it was the spacer in this modifier taking up a minimum amount of space.

Setting the minimum to zero fixes the issue.

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
```diff
@@ -79,7 +79,6 @@ struct SubtextView: View {
                 }
             }
         }
-        .transition(.slide)
```
gordonbrander, line 82: Unrelated change for issue discovered while troubleshooting the SubtextView. We should not be decorating SubtextView with transition(.slide). This is a general-purpose view, and slide is a very opinionated transition (slides left to right).

bfollington: I'm not even sure why this is here in the first place, any transition should be on the container...

bfollington (approved):
LGTM!

PR #936, opened 2023-10-16
Refactor excerpt generation and rendering

bfollington:
- Implement updated story styles
- Catch user profile loading error
- Show leading sigil in excerpt
- Implement ExcerptView via SubtextView
- Update notebook entry row spacing
- Handle the case that a user has no profile memo
- Lift Divider()s outside Story views
- Implement FAB + Search for profile tab
- Add ShareLink to context menu
- Support tapping slashlinks in excerpt
- Remove unused callback
- Lots of fiddling to place context menu in the correct spot
- Tidy and comment code
- Fix indentation
- Introduce truncate vs excerpt, restore old ExcerptView in notebook
- Restore spacing in notebook
- Improve handling of link taps
- Revert to old method name
- Fall back to findAndPushDetail
- Rework & fix tests
- Fixing build errors after rebasing
- Preserve link (if we have one) to us as fallback
- Fix indentation
- Rework truncate() logic
- Use a LazyVStack in our profile view
- Rework Feed empty view + structure
- Improve profile empty states
- Pass store and notify down directly
- Rename closure to onRequestDetail
- Add punctuation to empty state
- Pass excerpt as fallback
- Rework TranscludeView
- Pass excerpt through again


bfollington:
With the proposed changes _alone_ we are prone to displaying excessively large stories in the feed if the second block is large.

Screenshot 2023-10-17 at 2 10 03 pm

If we lower the character limit of the excerpt itself we'd be at risk of slicing into markup but we can use `lineLimit().truncationMode()` in the view to cap the max height preserving the formatting.

Screenshot 2023-10-17 at 2 09 41 pm

(Note the italics being truncated)

bfollington:
We can fix this by using a smarter approach to truncation:

```swift
var length = min(560 * maxBlocks, 4000)
let dom = Subtext(markup: text)
var blocks: [Subtext.Block] = []

for block in dom.blocks {
    if block.isEmpty {
        continue
    }
    
    if length > block.span.count {
        length -= block.span.count
        blocks.append(block)
    } else {
        break
    }
}

return Subtext(base: dom.base, blocks: Array(blocks))
```

This will never cut a block in half and can never produce invalid markup as a result.

Combined with the `lineLimit` approach above:

image

The Notebook previews are still working well too:

image

bfollington:
Sidenote: It feels tempting to add the "Show More..." prompt that twitter has when this truncation occurs but... all the posts may have more if you tap because we only render 2 blocks worth. We'd actually need to detect if truncation has occurred OR if this document has more than 2 blocks.

xcode/Subconscious/Shared/Models/StoryEntry.swift
```diff
@@ -11,8 +11,7 @@ import Foundation
 struct StoryEntry:
     Hashable,
     Identifiable,
-    CustomStringConvertible,
-    Codable
```
bfollington, line 15: `Subtext` is not `Codable` and `EntryStub.excerpt` is `Subtext` now, so we have to drop this protocol.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -202,6 +202,10 @@ enum AppAction {
     case indexOurSphere
     case succeedIndexOurSphere(OurSphereRecord)
     case failIndexOurSphere(String)
+    /// Developer utility function, for now
+    case clearIndex
```
bfollington, line 206: "resetIndex" might be better, we don't actually delete the data

xcode/Subconscious/Shared/Components/BacklinksView.swift
```diff
@@ -9,7 +9,8 @@ import SwiftUI
 
 struct BacklinksView: View {
     var backlinks: [EntryStub]
-    var onSelect: (EntryLink) -> Void
+    var onRequestDetail: (EntryLink) -> Void
+    var onLink: (_ context: Slashlink, SubSlashlinkLink) -> Void
```
bfollington, line 13: This pattern comes up a lot in this PR: we need to know what context a link was tapped in. It could be the viewer or a transclude or a backlink which changes the actual navigation address.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -638,9 +638,9 @@ extension DetailStackAction {
             return .requestDeleteMemo(address)
         case let .requestDetail(detail):
             return .pushDetail(detail)
-        case let .requestFindLinkDetail(link):
+        case let .requestFindLinkDetail(context, link):
             return .findAndPushLinkDetail(
-                address: Slashlink.ourProfile,
+                address: context,
                 link: link
             )
```
bfollington, line 645: We needed additional context to support links within Backlinks

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -202,6 +202,10 @@ enum AppAction {
     case indexOurSphere
     case succeedIndexOurSphere(OurSphereRecord)
     case failIndexOurSphere(String)
+    /// Developer utility function, for now
+    case resetIndex
+    case succeedresetIndex
+    case failresetIndex(String)
```
gordonbrander, line 208: This seems like an unrelated change.

Also a potential footgun w/r/t syncing database with sphere which is currently pretty carefully made to be atomic. I haven't worked out all the implications, so could be wrong about that.

xcode/Subconscious/Shared/Components/BacklinksView.swift
gordonbrander: Q: how does the slashlink communicate whether context is viewer or transclude or backlink? Should we introduce an enum to encode it?

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -253,6 +253,12 @@ actor DataService {
         
         return results
     }
+    
+    func resetIndex() async throws {
+        let did = try await self.noosphere.identity()
+        try self.database.resetOurSphere(identity: did)
+        try self.database.resetPeers()
+    }
```
gordonbrander, line 261: Mentioned above, but how does this work with sphere->database synchronization?

We're pretty careful right now to keep database updates atomic and lock-step with sphere updates. Naively, to me, this would seem to violate those assumptions.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -173,6 +173,17 @@ final class DatabaseService {
...
+    /// Drop the record of our sphere to force re-indexing
+    func resetOurSphere(identity: Did) throws {
+        try database.execute(
+            sql: """
+            DELETE FROM our_sphere
+            WHERE did = ?
+            """,
+            parameters: [.text(identity.description)]
+        )
+    }
```
gordonbrander, line 185: Same question about sphere -> database sync.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -325,6 +336,20 @@ final class DatabaseService {
...
+            throw DatabaseServiceError.notReady
+        }
+        
+        // set the "since" column to NULL for this peer if it exists
+        try database.execute(
+            sql: """
+            UPDATE peer SET since = NULL
+            """
+        )
+    }
```
gordonbrander, line 352: Same question about sphere -> database sync.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -662,19 +705,11 @@ extension Subtext {
...
-        let validBlocks = blocks
-            .filter { block in !block.isEmpty }
-            .prefix(maxBlocks)
-        
-        return Subtext(base: self.base, blocks: Array(validBlocks))
-    }
-    
     static func excerpt(markup: String, fallback: String = "") -> String {
         let prefix = markup.truncate(maxLength: maxExcerptSize)
         return Subtext(markup: String(prefix)).excerpt(fallback: fallback)
```
gordonbrander, line 715: We should remove or update this function to use the logic from `truncate()`.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -662,19 +705,11 @@ extension Subtext {
             .prefix(2) // Take first two blocks
         
         let output = validBlocks.joined(separator: "\n")
+            .truncate(maxLength: Self.maxExcerptSize)
```
gordonbrander, line 708: The differences between this and `Subtext.truncate` are a source of inconsistencies.

I think we should factor the string truncation logic out of `Subtext.truncate`, and update `String.truncate` to use it, then use `String.truncate` in both places for consistent string truncation behavior.

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -451,7 +476,7 @@ final class DatabaseService {
 
         let results = try database.execute(
             sql: """
-            SELECT slashlink, modified, excerpt
+            SELECT slashlink, modified, length(body) AS length, excerpt
```
gordonbrander, line 479: Noting that the excerpt field is created during write by way of `Memo.excerpt()` for `MemoRecord()`.

We truncate this field so that we can reduce the amount of text that has to be returned for list queries. We'll want to make sure that `Memo.excerpt()` uses the same truncation length/logic as `EntryStub`'s excerpt generation code.

I believe this is the case.

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
```diff
@@ -7,51 +7,40 @@
...
-                    ExcerptView(excerpt: entry.excerpt)
+                    SubtextView(
+                        subtext: entry.excerpt,
+                        transcludePreviews: [:],
+                        onViewTransclude: { _ in
+                            // Nested transcludes are not supported
+                        },
+                        onTranscludeLink: { _, _ in
+                            // Nested transcludes are not supported
+                        }
```
gordonbrander, line 33: Perhaps we should make these closures Optional or give them a no-op default?

xcode/Subconscious/Shared/Components/Common/Transclude/TranscludeView.swift
```diff
@@ -7,51 +7,40 @@
...
                 VStack(alignment: .leading, spacing: AppTheme.unit2) {
                     BylineSmView(
                         pfp: .generated(entry.did),
                         slashlink: entry.address
                     )
                     
-                    ExcerptView(excerpt: entry.excerpt)
+                    SubtextView(
+                        subtext: entry.excerpt,
+                        transcludePreviews: [:],
```
gordonbrander, line 27: Perhaps we should give this a default of empty?

gordonbrander (changes requested):
A bit of feedback + a few questions. Overall looks good!

xcode/Subconscious/Shared/Components/BacklinksView.swift
bfollington: It doesn't need to know, it just needs the address that the link sits within. We can add a wrapping enum but we would just discard that information later anyway.

xcode/Subconscious/Shared/Models/Entry.swift
```diff
@@ -31,8 +31,9 @@ extension EntryLink {
 extension EntryStub {
     init(_ entry: MemoEntry, did: Did) {
         self.address = entry.address
-        self.excerpt = entry.contents.excerpt()
+        self.excerpt = Subtext(markup: entry.contents.body)
```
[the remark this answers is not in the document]
bfollington: Yes, good call, this is a type.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -45,6 +45,13 @@ struct Subtext: Hashable, Equatable, LosslessStringConvertible {
     var description: String {
         String(base)
     }
+    
+    /// Return a snapshotted String of the contents of `blocks`, joined by newlines.
+    func toString() -> String {
+        blocks
+            .map { b in b.span.toString() }
+            .joined(separator: "\n")
+    }
```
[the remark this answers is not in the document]
bfollington: No, because in the instance of `Subtext.truncate` we create a `Subtext` instance that uses the same `base` but only includes _some_ of the blocks. This avoids re-allocating the string but if we read directly from `base` we will get the original document not the filtered set of blocks.

We _could_ just create a new `base` String and point all the blocks at that though? I got the impression from the API we should avoid directly reading from `base` _because_ not every char in `base` has to correspond to a `block`.

xcode/Subconscious/Shared/Components/AppView.swift
bfollington: > This seems like an unrelated change.

It's near impossible to test excerpt generation without it and it's a developer utility... I don't see how it's much different to resetting the first run experience? 

We know the indexing process is designed to read the entire sphere if the `since` field is NULL, which is all this does.

xcode/Subconscious/Shared/Services/DataService.swift
bfollington: As mentioned here: https://github.com/subconsciousnetwork/subconscious/pull/936#discussion_r1367990163 this is no different to it being the first time we index something. We're only resetting a pointer to force re-indexing from the beginning of history, not actually deleting any indexed data. I don't see how that can lead to a desync.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -653,6 +660,42 @@ extension Subtext {
     private static let maxExcerptSize = 280
     private static let maxGeneratedSlugSize = 128
     
+    static func truncate(text: String, maxBlocks: Int, fallback: String = "") -> Subtext {
+        let length = min(280 * maxBlocks, 4000)
```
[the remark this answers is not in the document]
bfollington: WFM, the 4000 limit is practically never relevant since we are always asking for 2 blocks right now anyway.

xcode/Subconscious/Shared/Models/EntryStub.swift
```diff
@@ -13,12 +13,12 @@ struct EntryStub:
...
     Identifiable,
-    CustomDebugStringConvertible,
-    Codable
+    CustomDebugStringConvertible
 {
     let did: Did
     let address: Slashlink
-    let excerpt: String
+    let excerpt: Subtext
+    let contentLength: Int
```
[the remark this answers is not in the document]
bfollington: Good call, done.

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: See https://github.com/subconsciousnetwork/subconscious/pull/936#discussion_r1367990163

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: See https://github.com/subconsciousnetwork/subconscious/pull/936#discussion_r1367990163

xcode/Subconscious/Shared/Parsers/Subtext.swift
bfollington: I've reflowed the API surface there, now there's:

- `Subtext.truncate` (static) -> takes a string, truncates it at `Subtext.maxExcerptLength*maxBlocks` chars, parses it and takes `maxBlocks` from it
- `Subtext.truncate` (member) -> takes the first `maxBlocks` from this `Subtext`
- `Subtext.excerpt` (static) -> takes a string, calls `Subtext.truncate` (static) with `maxBlocks: Self.maxExcerptBlocks`
- `Subtext.excerpt` (member) -> calls `Subtext.truncate` (member) with `maxBlocks: Self.maxExcerptBlocks`

These make use of `String.truncateAtWordBounary`. I'm not sure what the implications of changing `String.truncate` everywhere are yet, will find out.

xcode/Subconscious/Shared/Parsers/Subtext.swift
bfollington: I've refactored the API surface, see https://github.com/subconsciousnetwork/subconscious/pull/936#discussion_r1368053902

xcode/Subconscious/Shared/Components/BacklinksView.swift
gordonbrander: If all it needs is the Peer, we could consider passing that.

xcode/Subconscious/Shared/Parsers/Subtext.swift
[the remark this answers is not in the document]
gordonbrander: @bfollington `base` is a Substring rather than a string. This means multiple Subtext instances may share the same underlying string, but slice from different parts of it.

The Substring in `Subtext` should always correspond 1:1 with the parsed dom. We make use of this invariant in a number of ways.

We attempt to enforce this via the API. This is the reason that both `base` and `blocks` are `let` properties, and also the reason that `init(base:blocks:)` is a private constructor (it's an internal shortcut for the parser).

We want to make sure we uphold this invariant. I think we should get rid of `init(base:blocks:)` and force everything through the string/substring code path. The footgun is not worth any marginal perf gain we may have had through avoiding a re-parse.

In the case of `excerpt()`, I think that we should take the relevant blocks, stringify them, pass the resulting string to `Subtext(markup:)` and allow that to re-parse (it's cheap). This means that the excerpt doesn't need to hold on to the full document in memory, since its `base` points to the stringified excerpt.

xcode/Subconscious/Shared/Parsers/Subtext.swift
```diff
@@ -651,33 +658,47 @@ extension Subtext {
...
     }
     
-    func truncate(_ maxBlocks: Int = 2, fallback: String = "") -> Subtext {
-        // Filter out empty blocks
-        let validBlocks = blocks
-            .filter { block in !block.isEmpty }
-            .prefix(maxBlocks)
-        
-        return Subtext(base: self.base, blocks: Array(validBlocks))
+    static func excerpt(markup: any StringProtocol, fallback: String = "") -> Subtext {
```
gordonbrander, line 692: My understanding is that `any` has a runtime cost vs generics. May want to write this as a generic with a `StringProtocol` constraint.

gordonbrander (changes requested):
We must make sure that `Subtext.base` corresponds 1:1 with the parsed markup (see comments).

Otherwise LGTM!

xcode/Subconscious/Shared/Parsers/Subtext.swift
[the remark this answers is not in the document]
bfollington: SGTM, I'll drop Subtext.toString after that and use the description instead.

xcode/Subconscious/Shared/Parsers/Subtext.swift
bfollington: Change to `<S: StringProtocol>` and tweaked the API, I believe this should avoid the `any` tax.

gordonbrander (approved):
LGTM!

xcode/Subconscious/Shared/Library/Prose.swift
```diff
@@ -44,6 +44,37 @@ extension String {
         )
         return !truncated.isEmpty ? "\(truncated)\(ellipsis)" : ""
     }
+    
+    /// Take the first `maxBlocks` blocks or `Subtext.maxExcerptSize * maxBlocks` characters worth, terminating with an
+    /// ellipsis. If the string is truncated, it will be sliced to the nearest word boundary to avoid breaking markup.
+    /// This is used to implement `Subtext.excerpt`.
+    static func truncateAtWordBoundary(markup: any StringProtocol, maxChars: Int, fallback: String = "") -> String {
```
gordonbrander: This could be more efficiently defined as a generic instead of `any` arg.

No action necessary.

PR #944, opened 2023-10-18
Index peers serially (and other performance enhancements)

### LINKED ISSUE #943: App seems to be requesting too often, resulting in long timeouts
On main, the app is taking longer (30s-60s) to resolve and fallback to cached 3p sphere content than it should (5s). It seems that we are hitting the gateway too often somewhere and causing redundant requests, stacking timeouts.

@cdata says:

>  It definitely looks like the app is hammering the gateway over and over again trying to get a thing it will never get. My guess is that there is a hot path here where you traverse to a sphere to read one thing and then discard that sphere, repeat for every slug (or something like that where there is N>1 things to do for that sphere). If this is the case, you'll have N*T seconds of wait time most likely, where T is the max timeout (~5s). So, 35s could be 7 things timing out one after another. It's not outright bad to discard a sphere N times, but it would certainly be better not to do that if possible.

> I was able to reproduce Gordon's case in prod for my "following" view. From the gateway logs, I saw a request to traverse to the same sphere multiple times one after another. (And, in this case, it was a sphere that referenced a missing block, so it timed out over and over)

bfollington:
# Changes

1. Index peers in serial, never concurrently (introduce `Task.yield()` between each peer)
2. Pass `sphere` reference around in `UserProfileService` to minimise excessive init/deinit
3. Prevent excessive profile refreshing
4. Prevent refreshing profile if it's already loading
5. Log each step of producing the profile view
6. Fix bug: we were loading the data for the entire profile view just to get our nickname on startup

None of these dramatically changed anything for me in testing but they do all incrementally improve performance.

Fixes #943

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -1975,10 +1975,11 @@ struct AppModel: ModelProtocol {
         peers: [PeerRecord]
     ) -> Update<Self> {
         // Transform list of peers into fx publisher of actions.
-        let fx: Fx<Action> = Future.detached {
-            AppAction.indexPeers(peers
-                .map({ peer in peer.petname }))
-        }
+        let fx: Fx<Action> = Just(
+            AppAction.indexPeers(
+                peers.map({ peer in peer.petname })
+            )
+        )
         .eraseToAnyPublisher()
         
         return Update(state: state, fx: fx)
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -2051,6 +2051,7 @@ struct AppModel: ModelProtocol {
                         petname: petname
                     )
                     success.append(peer)
+                    await Task.yield()
                 } catch {
                     fail.append(petname)
                 }
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -2014,37 +2014,11 @@ struct AppModel: ModelProtocol {
     ) -> Update<Self> {
         
         let fx: Fx<Action> = Future.detached(priority: .background) {
-            var results: [PeerIndexResult] = []
-            
-            for petname in petnames {
-                logger.log(
-                    "Indexing peer",
-                    metadata: [
-                        "petname": petname.description
-                    ]
-                )
-                
-                do {
-                    let peer = try await environment.data.indexPeer(
-                        petname: petname
-                    )
-                    
-                    results.append(.success(peer))
-                    await Task.yield()
-                } catch {
-                    results.append(
-                        PeerIndexResult.failure(
-                            PeerIndexError(
-                                error: error,
-                                petname: petname
-                            )
-                        )
-                    )
-                }
-            }
-            
+            let results = await environment.data.indexPeers(petnames: petnames)
             return AppAction.completeIndexPeers(results: results)
-        }.eraseToAnyPublisher()
+        }
+        .eraseToAnyPublisher()
+        
         return Update(state: state, fx: fx)
     }
     
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -98,8 +98,6 @@ enum AppAction {
         category: "AppAction"
     )
 
-    case writeTestData
-    
     /// Sent immediately upon store creation
     case start
 
@@ -595,11 +593,6 @@ struct AppModel: ModelProtocol {
         environment: AppEnvironment
     ) -> Update<AppModel> {
         switch action {
-        case .writeTestData:
-            return writeTestData(
-                state: state,
-                environment: environment
-            )
         case .start:
             return start(
                 state: state,
@@ -1089,24 +1082,6 @@ struct AppModel: ModelProtocol {
     ) -> Update<AppModel> {
         logger.log("\(message)")
         return Update(state: state)
-    }
-    
-    static func writeTestData(
-        state: AppModel,
-        environment: AppEnvironment
-    ) -> Update<Self> {
-        
-        let fx: Fx<AppAction> = Future.detached {
-            for _ in 0..<100 {
-                try? await environment.noosphere.write(slug: Slug.dummyData(), contentType: "text/subtext", additionalHeaders: [], body: String.dummyDataMedium().data(using: .utf8)!)
-            }
-            
-            try? await environment.noosphere.save()
-            return AppAction.start
-        }
-        .eraseToAnyPublisher()
-        
-        return Update(state: state, fx: fx)
     }
     
     static func start(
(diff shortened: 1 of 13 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -2006,63 +2003,67 @@ struct AppModel: ModelProtocol {
...
+                do {
+                    let peer = try await environment.data.indexPeer(
+                        petname: petname
+                    )
+                    success.append(peer)
+                    await Task.yield()
+                } catch {
+                    fail.append(petname)
+                }
             }
```
bfollington, line 2033: Running indexing serially instead of concurrently. We should factor this into a method.

xcode/Subconscious/Shared/Components/Common/Profile/UserProfileView.swift
```diff
@@ -80,9 +80,9 @@ struct FollowTabView: View {
         ForEach(store.state.following) { follow in
             StoryUserView(
                 story: follow,
-                action: { _ in
+                action: { address in
                     notify(
-                        .requestNavigateToProfile(follow.user.address)
+                        .requestNavigateToProfile(address)
```
bfollington, line 85: Extra bug I caught while investigating. This could result in using a non-optimal address to reach a peer before.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -167,26 +167,26 @@ actor UserProfileService {
     /// Because profile data is optional and we expect it will not always be present
     /// any errors are logged & handled and nil will be returned if reading fails.
     func readProfileMemo(
-        address: Slashlink
+        sphere: SphereProtocol
```
bfollington, line 170: Avoid traversal in `noosphere.read` by passing the sphere instance.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -206,23 +206,24 @@ actor UserProfileService {
     
     /// Load the underlying `_profile_` memo for a user and construct a `UserProfile` from it.
     private func loadProfileFromMemo(
-        did: Did,
+        sphere: SphereProtocol,
```
bfollington, line 209: Thread `sphere` through

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -251,10 +252,10 @@ actor UserProfileService {
     
     /// Takes a list of slugs and prepares an `EntryStub` for each, excluding hidden slugs.
     private func loadEntries(
+        sphere: SphereProtocol,
```
bfollington, line 255: Avoid opening the sphere again, pass reference through

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1257,12 +1252,12 @@ struct AppModel: ModelProtocol {
...
     
     static func fetchNicknameFromProfileMemo(
-        state: AppModel,
+        state: AppModel, 
         environment: AppEnvironment
     ) -> Update<AppModel> {
         let fx: Fx<AppAction> = Future.detached {
-            let response = try await environment.userProfile.requestOurProfile()
-            if let nickname = response.profile.nickname {
+            let response = try await environment.userProfile.readOurProfile(alias: nil)
```
bfollington, line 1259: We were asking for **way** more data than we needed before just to get our nickname.

xcode/Subconscious/Shared/Services/UserProfileService.swift
```diff
@@ -167,26 +167,26 @@ actor UserProfileService {
...
+        sphere: SphereProtocol
     ) async -> UserProfileEntry? {
+        let identity = try? await sphere.identity()
         do {
-            guard address.isProfile else {
-                throw UserProfileServiceError.attemptToReadProfileFromInvalidAddress
-            }
-            
-            let data = try? await noosphere.read(slashlink: address)
+            let data = try? await sphere.read(slashlink: Slashlink.ourProfile)
```
bfollington, line 174: Since we have a sphere reference "ourProfile" is always "this sphere's profile"

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -951,7 +944,7 @@ struct UserProfileDetailModel: ModelProtocol {
     ) -> Update<Self> {
         update(
             state: state,
-            action: .refresh(forceSync: true),
+            action: .refresh(forceSync: false),
```
bfollington, line 947: We shouldn't need to sync after resolution, because resolution comes as result of a sync.

xcode/Subconscious/Shared/Components/Detail/UserProfileDetailView.swift
```diff
@@ -964,7 +957,7 @@ struct UserProfileDetailModel: ModelProtocol {
         logger.log("Failed to resolve followed user: \(error)")
         return update(
             state: state,
-            action: .refresh(forceSync: true),
+            action: .refresh(forceSync: false),
```
bfollington, line 960: We shouldn't sync after a failed resolution either.

gordonbrander (approved):
Straightforward improvements.

I left a couple of suggestions. The only important one is to factor out the indexing code into a method of DataService.

Playtested in dev.

xcode/Subconscious/Shared/Services/UserProfileService.swift
bfollington: "ourProfile" just means a peerless slashlink of `/_profile_` which should always point to _that sphere's_ profile. I can change the code to read more clearly though and use `Slashlink(slug: Slug.profile)`?

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1257,12 +1249,12 @@ struct AppModel: ModelProtocol {
     }
     
     static func fetchNicknameFromProfileMemo(
-        state: AppModel,
+        state: AppModel, 
         environment: AppEnvironment
     ) -> Update<AppModel> {
         let fx: Fx<AppAction> = Future.detached {
-            let response = try await environment.userProfile.requestOurProfile()
```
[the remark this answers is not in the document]
bfollington: Yah, I thought something similar while working, good call.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -211,9 +210,8 @@ enum AppAction {
     case failCollectPeersToIndex(_ error: String)
 
     /// Index the contents of a sphere in the database
-    case indexPeer(_ petname: Petname)
-    case succeedIndexPeer(_ peer: PeerRecord)
-    case failIndexPeer(petname: Petname, error: Error)
+    case indexPeers(_ petnames: [Petname])
+    case completeIndexPeers(succeeded: [PeerRecord], failed: [Petname])
```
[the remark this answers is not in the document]
bfollington: Done!

xcode/Subconscious/Shared/Components/AppView.swift
bfollington: I managed to add some tests to assert concurrency behaviour around yielding, but I can't think of a way to _actually_ index a peer without a gateway in the mix.

It might be possible if we have mocked versions of every service `DataService` calls out to but... I'm not a huge fan of writing 100s of lines of mocks which can just as easily introduce problems.

PR #951, opened 2023-10-19
Update StoryUser description to remove newlines

### LINKED ISSUE #948: StoryUser `description` includes newlines, breaks up logs
StoryUser `description` includes newlines and breaks up logs.

We should use the following:

```
    var description: String {
        let nickname = String(describing: user.nickname)
        return "StoryUser(\(user.did), \(user.address), \(nickname), \(user.ourFollowStatus))"
    }
```

gordonbrander:
We should not be including newlines in the text description of this type since these get logged and break up logging.

Fixes #948

### DIFF
--- a/xcode/Subconscious/SubconsciousTests/Tests_StoryUser.swift
+++ b/xcode/Subconscious/SubconsciousTests/Tests_StoryUser.swift
@@ -1,0 +1,41 @@
+//
+//  Tests_StoryUser.swift
+//  SubconsciousTests
+//
+//  Created by Gordon Brander on 10/19/23.
+//
+
+import XCTest
+@testable import Subconscious
+
+final class Tests_StoryUser: XCTestCase {
+    func testDescription() throws {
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/SubconsciousTests/Tests_StoryUser.swift
```diff
@@ -0,0 +1,41 @@
...
+        )
+        XCTAssertEqual(story.description, "StoryUser(did:key:abc123, @bob/foo, Optional(bob), following(bob))")
+    }
+
+    func testPerformanceExample() throws {
+        // This is an example of a performance test case.
+        self.measure {
+            // Put the code you want to measure the time of here.
+        }
+    }
```
bfollington, line 39: Can we drop this unused case?

bfollington (approved):
LGTM if we drop the unused `testPerformanceExample`

PR #952, opened 2023-10-19
Set NoosphereService log level depending on DEBUG

### LINKED ISSUE #949: Set Noosphere log level to DEBUG in Development scheme
When doing development, we often want fine-grained logs of what Noosphere is doing. We should automatically set `NoosphereService`'s Noosphere log level to `DEBUG` in the Development scheme.

gordonbrander:
If DEBUG, set academic, otherwise basic.

Fixes #949.


bfollington (approved):
LGTM

Issue #940, opened 2023-10-18
Swiping back to 3p content reloads it


gordonbrander:
Steps to reproduce:

- Go to profile tab > Following
- Tap on a profile
- Tap on 3p note (A) (loads > displays)
- Tap on a link to some other note (B)
- Go back (B -> A)

Result: A is reloaded. Loading indicator also spins faster than usual, suggesting bug in animation code?

Expected: A should be the cached result from last load.

Defining the right behavior:

- We should only load read-only panels once at creation.
    - In future we could consider some sort of time-elapsed cache expiry
- Panel can be refreshed via pull-to-refresh.

This is similar to browser behavior with page loading.

gordonbrander:
Yes, the editor must always refresh so as not to create sync conflicts.

However the viewer should load once and then hold onto this cached content unless there is a good reason to invalidate it (such as loading the same resource somewhere else at a newer version, pull to refresh, etc). The viewer is view-only, so there are not sync conflict issues, and it loads 3P content, so there are latencies involved in reloading content.

I think the expected behavior when swiping back is seeing the content you just left behind, rather than getting a loading screen. I would expect to swipe back and see cached content, in the same way that when I hit back on the browser I see a cached page.

bfollington:
@gordonbrander that scans. We should just invalidate the viewer after a sync then I suppose?

bfollington:
Finding: this only seems to happen when we fall back to an older revision of content due to missing blocks, otherwise the content is already cached locally and noosphere returns instantly.

PR #956, opened 2023-10-23
Fix double speed loading animation

bfollington:
Popping views off the stack could trigger `.task{}` again and double the speed of the animation.

Originally mentioned in https://github.com/subconsciousnetwork/subconscious/issues/940


gordonbrander (approved):
LGTM

PR #969, opened 2023-10-26
Do not refresh viewer if content is indentical

### LINKED ISSUE #940: Swiping back to 3p content reloads it

bfollington:
Fixes #940 

Introduce new `MemoViewer` action `.refreshAll` that actually does the loading. `.appear` will only trigger loading if the address changes and the `MemoViewerDetailView` listens for `AppAction.succeedSyncSphereWithGateway` to trigger `.refreshAll`.

My thinking is that 3P content can only change after a sync, so that's the only explicit trigger we need.

### DIFF
--- a/xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
+++ b/xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
@@ -473,7 +473,7 @@ struct MemoViewerDetailModel: ModelProtocol {
         description: MemoViewerDetailDescription
     ) -> Update<Self> {
         guard state.address != description.address else {
-            logger.warning("Attempted to appear with same address, doing nothing")
+            logger.log("Attempted to appear with same address, doing nothing")
             return Update(state: state)
         }
         
@@ -496,7 +496,7 @@ struct MemoViewerDetailModel: ModelProtocol {
         environment: Environment
     ) -> Update<Self> {
         guard let address = state.address else {
-            logger.warning("Attempted to refresh with nil address")
+            logger.log("Attempted to refresh with nil address")
             return Update(state: state)
         }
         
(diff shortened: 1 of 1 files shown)


xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -464,20 +472,45 @@ struct MemoViewerDetailModel: ModelProtocol {
         environment: Environment,
         description: MemoViewerDetailDescription
     ) -> Update<Self> {
+        guard state.address != description.address else {
+            logger.warning("Attempted to appear with same address, doing nothing")
```
bfollington, line 476: Might be a bit noisy to log?

gordonbrander: `logger.log` seems more appropriate here, since this is expected behavior that is being handled gracefully.

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -464,20 +472,45 @@ struct MemoViewerDetailModel: ModelProtocol {
...
+            environment: environment
+        )
+    }
+    
+    static func refreshAll(
+        state: Self,
+        environment: Environment
+    ) -> Update<Self> {
+        guard let address = state.address else {
+            logger.warning("Attempted to refresh with nil address")
```
gordonbrander, line 499: `logger.log` seems more appropriate here, since this is expected behavior that is being handled gracefully.

gordonbrander (changes requested):
Two suggestions, otherwise LGTM.

gordonbrander (approved):
LGTM assuming logging changes land

PR #971, opened 2023-10-27
Display empty circle glyph for local did

### LINKED ISSUE #958: Draft notes notes have generative PFP in backlinks instead of empty circle
I believe this is because `Did.local` is getting interpreted as a seed for the generative PFP.

In other parts of the app we detect `Did.local` and display the empty circle glyph instead. We should do that here too.

gordonbrander:
Fixes #958 

Screenshot 2023-10-27 at 14 55 21


bfollington (approved):
LGTM

PR #972, opened 2023-10-27
Fix delete animation in notes list

### LINKED ISSUE #955: Swipe-to-delete on notebook does not have correct animation
When we pop up the delete confirmation, the item is briefly removed from the list and then re-appears.

Perhaps this has to do with `NotebookAction.stageDeleteEntry`?

gordonbrander:
Fixes #955 

Using `role: .destructive` causes the list to preemptively remove the row. However, at this stage we have not removed it, only staged the removal, resulting in an animation jank as the item jumps back.

Instead, we do not assign a role, but give the action a tint. This causes it to show up as a destructive action (visually), but stages the deletion, avoiding the animation until the item is actually removed.

https://github.com/subconsciousnetwork/subconscious/assets/58421/e70c0afd-bbff-4ceb-8977-b3c163ae51f2

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
@@ -33,13 +33,13 @@ struct EntryListView: View {
                             allowsFullSwipe: false
                         ) {
                             Button(
-                                role: .destructive,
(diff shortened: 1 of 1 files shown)


bfollington (approved):
LGTM, I have a feeling this will come up again...

PR #973, opened 2023-10-31
Fix (possible) delay on tapping links

### LINKED ISSUE #957: Tapping on a link can have long delay before navigating
When we tap a link we seem to be `traversing` to calculate the address before pushing a new detail onto the stack. This results in a long wait where you can tap repeatedly and queue up multiple navigations.

Steps:

1. View a 3P note with a peered link (i.e. `@peer/slashlink`) in it
2. Tap the link multiple times (may only be possible with sufficient network latency)
3. Observe as nothing happens, then after a while, multiple copies of the detail are pushed onto the stack

bfollington:
Fixes #957 

- Ensure DID is known at link traversal time
    - Thread the DID through from whatever context the link is being tapped in, lets us skip `noosphere.resolve`
- Don't let `context` make its way to the `DetailStackModel`

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -545,8 +545,7 @@ struct AppModel: ModelProtocol {
     ///
     /// This property is updated at `.start` with the corresponding value
     /// stored in `AppDefaults`.
-    var sphereIdentity: String?
-    var sphereDid: Did?
+    var sphereIdentity: Did?
     
     /// Default sphere version, if any.
     var sphereVersion: String?
@@ -1174,7 +1173,7 @@ struct AppModel: ModelProtocol {
         state: AppModel,
         environment: AppEnvironment
     ) -> Update<AppModel> {
-        let sphereIdentity = state.sphereIdentity ?? "nil"
+        let sphereIdentity = state.sphereIdentity?.description ?? "nil"
         logger.debug(
             "appear",
             metadata: [
@@ -1427,8 +1426,7 @@ struct AppModel: ModelProtocol {
         sphereIdentity: String?
     ) -> Update<AppModel> {
         var model = state
-        model.sphereIdentity = sphereIdentity
-        model.sphereDid = Did(sphereIdentity ?? "")
+        model.sphereIdentity = Did(sphereIdentity ?? "")
         if let sphereIdentity = sphereIdentity {
             logger.debug("Set sphere ID: \(sphereIdentity)")
         }
@@ -2249,8 +2247,7 @@ struct AppModel: ModelProtocol {
         environment: AppEnvironment,
         inviteCode: InviteCode
     ) -> Update<AppModel> {
-        guard let did = state.sphereIdentity,
-              let did = Did(did) else {
+        guard let did = state.sphereIdentity else {
             // Attempt to create the sphere if it's missing.
             // We could retry redeeming the code automatically but
             // if .createSphere fails we'll end up in an infinite loop
@@ -2520,7 +2517,7 @@ struct AppModel: ModelProtocol {
                 .presentRecoveryMode(true),
                 .recoveryMode(
                     .populate(
-                        state.sphereDid,
+                        state.sphereIdentity,
                         GatewayURL(state.gatewayURL),
                         context
                     )
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -546,6 +546,9 @@ struct AppModel: ModelProtocol {
     /// This property is updated at `.start` with the corresponding value
     /// stored in `AppDefaults`.
     var sphereIdentity: String?
+    var sphereDid: Did? {
+        Did(sphereIdentity ?? "")
+    }
     /// Default sphere version, if any.
     var sphereVersion: String?
     /// State for rendering mnemonic/recovery phrase UI.
@@ -2517,7 +2520,7 @@ struct AppModel: ModelProtocol {
                 .presentRecoveryMode(true),
                 .recoveryMode(
                     .populate(
-                        Did(state.sphereIdentity ?? ""),
+                        state.sphereDid,
                         GatewayURL(state.gatewayURL),
                         context
                     )
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -546,9 +546,8 @@ struct AppModel: ModelProtocol {
     /// This property is updated at `.start` with the corresponding value
     /// stored in `AppDefaults`.
     var sphereIdentity: String?
-    var sphereDid: Did? {
-        Did(sphereIdentity ?? "")
-    }
+    var sphereDid: Did?
+    
     /// Default sphere version, if any.
     var sphereVersion: String?
     /// State for rendering mnemonic/recovery phrase UI.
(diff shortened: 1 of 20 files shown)


xcode/Subconscious/Shared/Components/Common/Story/StoryEntryView.swift
```diff
@@ -76,8 +76,8 @@ struct StoryEntryView: View {
                         onViewTransclude: { slashlink in
                             onRequestDetail(slashlink, story.entry.excerpt.description)
                         },
-                        onTranscludeLink: { address, link in
-                            onLink(address, link)
+                        onTranscludeLink: { entry, link in
+                            onLink(entry, link)
```
bfollington, line 80: Call it `context.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -306,14 +308,14 @@ struct DetailStackModel: Hashable, ModelProtocol {
     static func findAndPushLinkDetail(
         state: Self,
         environment: Environment,
+        did: Did,
```
bfollington, line 311: Call this `owner`

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -86,7 +92,8 @@ struct MemoEditorDetailView: View {
                             .padding(.bottom, AppTheme.unit4)
                         BacklinksView(
                             backlinks: store.state.backlinks,
-                            onRequestDetail: { link in
+                            onRequestDetail: {
+                                link in
```
bfollington, line 96: Prev line

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -96,8 +103,16 @@ struct MemoEditorDetailView: View {
                                     )
                                 )
                             },
-                            onLink: { address, link in
-                                notify(.requestFindLinkDetail(address, link: link))
+                            onLink: {
+                                context,
+                                link in
```
bfollington, line 108: prev line

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -102,7 +102,12 @@ struct MemoViewerDetailView: View {
 
 /// View for the "not found" state of content
 struct MemoViewerDetailNotFoundView: View {
-    var backlinks: [EntryStub]
+    @ObservedObject var store: Store<MemoViewerDetailModel>
+    
+    var backlinks: [EntryStub] {
+        store.state.backlinks
+    }
+
```
bfollington, line 110: Revert?

xcode/Subconscious/Shared/Components/Feed/FeedListView.swift
```diff
@@ -33,11 +33,12 @@ struct FeedListView: View {
                                 )
                             )
                         },
-                        onLink: { address, link in
+                        onLink: { entry, link in
```
bfollington, line 36: call it `context`

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -540,6 +540,9 @@ struct AppModel: ModelProtocol {
     /// This property is updated at `.start` with the corresponding value
     /// stored in `AppDefaults`.
     var sphereIdentity: String?
+    var sphereDid: Did? {
+        Did(sphereIdentity ?? "")
+    }
```
bfollington, line 545: We were doing this all over the place, may as well centralize it.

gordonbrander: Optional: we could consider doing this when assigning the value to the model.

xcode/Subconscious/Shared/Components/BacklinksView.swift
```diff
@@ -10,7 +10,7 @@ import SwiftUI
 struct BacklinksView: View {
     var backlinks: [EntryStub]
     var onRequestDetail: (EntryLink) -> Void
-    var onLink: (_ context: Slashlink, SubSlashlinkLink) -> Void
+    var onLink: (_ context: EntryStub, SubSlashlinkLink) -> Void
```
gordonbrander, line 13: I'm generally in favor of passing as little information as possible to callbacks. A few open-ended thoughts...

IIRC there are a few places now where we need to pass a did with the slashlink. This feels like a hint that `slashlink` is not giving us everything we need to do addressing in every case.

We have the `Link` type which is an identifier made up of did + slug. If this is an option here, I would rather use it. However, this does lose petname information, which may be useful.

It pains me to say it but we may want to introduce an `Address` type that contains a DID plus slashlink, or perhaps a `Did`, `Petname?` and `Slug`.

Thoughts?

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -169,8 +175,13 @@ struct MemoViewerDetailLoadedView: View {
             return .systemAction
         }
         
+        guard let did = user?.did else {
+            return .systemAction
```
gordonbrander, line 179: I don't think we want to default to system action here, since we know this will be a `sub://` URL.

Instead, we want to return `.handled` and log what went wrong.

However, I also think that we should be encoding everything we need to follow a URL into the URL. Let's consider adding the necessary DID to the URL.

BTW what role does the DID play in looking up the URL?

gordonbrander (changes requested):
A few questions and asks. Thanks!

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
bfollington: > what role does the DID play in looking up the URL?

The DID is how we detect if we know this user by another (shorter) name. We may tap a link to `@ben.gordon.chris/hello` but if we follow `@ben` already we want to land on `@ben/hello`. If we don't pass the DID through then we have to call `resolve` on `@ben.gordon.chris`, introducing latency.

xcode/Subconscious/Shared/Components/BacklinksView.swift
bfollington: Yep, I agree that the triple of `DID` + `Slashlink` or `DID` + `Petname?` + `Slug` deserves a type. I like the sound of `Address`, I'll explore that path.

xcode/Subconscious/Shared/Components/AppView.swift
bfollington: Done!

xcode/Subconscious/Shared/Components/BacklinksView.swift
bfollington: I ended up going with a more specific name: `ResolvedAddress`. We use the term address all over the codebase to refer to a `Slashlink` and from my POV a Slashlink is the most common kind of address in the app. `ResolvedAddress` is an attempt to carve out a more specific role.

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -546,6 +546,8 @@ struct AppModel: ModelProtocol {
     /// This property is updated at `.start` with the corresponding value
     /// stored in `AppDefaults`.
     var sphereIdentity: String?
```
gordonbrander, line 548: Can we drop `sphereIdentity` given we have `sphereDid`?

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -169,17 +174,22 @@ struct MemoViewerDetailLoadedView: View {
             return .systemAction
         }
         
+        guard let did = user?.did else {
+            MemoViewerDetailModel.logger.log("User profile is not loaded, can't handle link")
```
gordonbrander, line 178: Nit: I wouldn't mind giving the view its own logger.

gordonbrander (approved):
LGTM

PR #981, opened 2023-11-02
Add UIHostingView

gordonbrander:
...a helper view that makes it easier to embed UIHostingControllers in a parent view. Pass a parent controller into the initalizer, along with the UIHostingController. It will add the UIHostingController to the parent controller, and remove the UIHostingController from the parent controller when the UIView is deinitialized.

This should make it easier for us to embed SwiftUI views in the block editor.

Part of epic #214.


bfollington (approved):
Nice, LGTM!

PR #1004, opened 2023-11-29
Notify on sync failure and success

bfollington:
The app never notifies the user that sync failed (or that it succeeded for that matter) which would've made it easier to diagnose an issue @justinabrahms ran into today.

This also provides some feedback that a drag-to-refresh did something.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -1996,7 +1996,12 @@ struct AppModel: ModelProtocol {
         
         return update(
             state: model,
-            action: .indexOurSphere,
+            actions: [
+                .indexOurSphere,
+                .toastStack(
+                    .pushToast(message: "Sync complete")
+                )
+            ],
             environment: environment
         )
     }
@@ -2018,7 +2023,12 @@ struct AppModel: ModelProtocol {
         
         return update(
             state: model,
-            action: .indexOurSphere,
+            actions: [
+                .indexOurSphere,
+                .toastStack(
+                    .pushToast(message: "Failed to sync")
+                )
+            ],
             environment: environment
         )
     }


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -1996,7 +1996,12 @@ struct AppModel: ModelProtocol {
         
         return update(
             state: model,
-            action: .indexOurSphere,
+            actions: [
+                .indexOurSphere,
+                .toastStack(
+                    .pushToast(message: "Sync complete")
+                )
+            ],
```
gordonbrander, line 2004: I don't think we should notify on sync complete. We should simply show the new content.

gordonbrander (changes requested):

xcode/Subconscious/Shared/Components/AppView.swift
bfollington: I've heard from Justin many times (and I agree based on my experience) that it's a broken interaction loop when you refresh and there's no feedback to let you know you're up-to-date & should expect to see the latest.

Most apps seem to represent this as a small notification pill or a persistent indicator, I agree that a full toast notification isn't ideal... just trying not to scope creep it.

gordonbrander: @bfollington the typical interaction pattern on iOS is:

Pull-to-refresh -> loading indicator -- loading indicator displayed until load finished --> display new content if any

The temptation to add a success pill is, I think, a byproduct of sync being unreliable for us in the past, and training us to expect failure. Instead of notifying success, we should make refresh reliable, and also make the loading indicator stay on-screen the reload request is finished.

gordonbrander: After discussion: land this, follow up with this as a short-term solution, then follow up with systemic redesign.

gordonbrander (approved):
After discussion: land this as a short-term solution, then follow up with systemic redesign.

PR #1013, opened 2023-12-04
Fix action logging in XCode 15

### LINKED ISSUE #1012: iOS 15 redacting all actions in logs
> The unified logging system uses privacy options to hide or show interpolated variables in a message. By default, the system doesn’t redact integer, floating-point and Boolean values, but it does redact the contents of dynamic strings and complex dynamic objects. To make a private value public again, configure the privacy of the variable using appropriate modifiers in your message string or interpolated variable. For example, the following code shows how to make a dynamic string visible again using the public modifier: https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code#:~:text=The%20unified%20logging%20system%20uses,again%20using%20the%20public%20modifier%3A

Supposedly this has been the case since 2016. However, we're running into it only now, and it seems others are as well https://developer.apple.com/forums/thread/738648. It seems the behavior of XCode has changed.

- https://developer.apple.com/forums/thread/738648
- https://stackoverflow.com/questions/38344674/why-does-xcode-print-logmessagelogging-private-in-console

gordonbrander:
Fixes #1012.

Requires https://github.com/subconsciousnetwork/ObservableStore/pull/44 to land. Will leave this PR as a draft until we review + cut a new release for ObservableStore including this change.

It seems that logging behavior changed in XCode 15. Now strings and objects are masked private by default.

This change updates ObservableStore to a new version that logs actions at the public level by default. It also updates all of our stores to use ObservableStore's logger instead of logging on `onReceive`.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -16,7 +16,12 @@ struct AppView: View {
     @StateObject private var store = Store(
         state: AppModel(),
         action: .start,
-        environment: AppEnvironment.default
+        environment: AppEnvironment.default,
+        loggingEnabled: true,
+        logger: Logger(
+            subsystem: Config.default.rdns,
+            category: "AppStore"
+        )
     )
     @Environment(\.scenePhase) private var scenePhase: ScenePhase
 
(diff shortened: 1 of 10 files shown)


bfollington (approved):
LGTM, pending the ObservableStore update ofc

Issue #1002, opened 2023-11-22
Tapping on a link in the viewer can load the wrong address


bfollington:
When viewing one of Gordon's notes, he had linked to `@ben/knowledge-structures` and when I tapped that link I went to `@gordon/knowledge-structures` instead of `@ben.gordon/knowledge-structures`. 

Something is wrong in the address translation logic, we should make sure to add a test to cover whatever case is incorrectly handled here.

PR #1014, opened 2023-12-05
Fix multiplayer links

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1002

I've extracted all the logic for link mapping to one function, reworked the approach and added tests and plenty of comments. The bug was due to passing the wrong DID to an action at one of the view call sites so I've tried to remove that footgun from the equation going forward.


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -2896,9 +2896,19 @@ struct AppModel: ModelProtocol {
...
 
+protocol TestableEnvironment {
+    var noosphere: NoosphereService { get }
+    var database: DatabaseService { get }
+    var data: DataService { get }
+    var transclude: TranscludeService { get }
+    
+    var addressBook: AddressBookService { get }
+    var userProfile: UserProfileService { get }
+}
```
bfollington, line 2907: Introduced this so that we can have static functions that work in both testing and the app itself. 

We could alternatively move `DetailStackModel.findBestAddressForLink` into one a service to avoid this change.

xcode/Subconscious/Shared/Components/BacklinksView.swift
```diff
@@ -10,7 +10,7 @@ import SwiftUI
 struct BacklinksView: View {
     var backlinks: [EntryStub]
     var onRequestDetail: (EntryLink) -> Void
-    var onLink: (_ context: ResolvedAddress, SubSlashlinkLink) -> Void
+    var onLink: (_ context: Peer, SubSlashlinkLink) -> Void
```
bfollington, line 13: I dropped the `ResolvedAddress` type in this PR. I think `Peer` is the right balance of flexible and legible.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -104,15 +104,13 @@ enum DetailStackAction: Hashable {
     /// If slashlink does not have a peer part, this will
     /// request an editor detail.
     case findAndPushDetail(
-        address: Slashlink,
-        link: SubSlashlinkLink? = nil
+        address: Slashlink
     )
```
bfollington: I'm slightly firming up the difference between these cases.

`findAndPushDetail` is for taking an address and pushing the correct navigation item onto the stack for it. It takes an optional `fallback` parameter now for the editor case.

`findAndPushLinkDetail` will find the best address for a link, then call through to `findAndPushDetail`. It should populate `fallback` if possible.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -105,14 +105,13 @@ enum DetailStackAction: Hashable {
...
         address: Slashlink,
-        link: SubSlashlinkLink? = nil
+        fallback: String?
     )
     
     case findAndPushLinkDetail(
-        context: ResolvedAddress,
+        context: Peer?,
         link: SubSlashlinkLink
     )
```
bfollington, line 114: I'm slightly firming up the difference between these cases.

`findAndPushDetail` is for taking an address and pushing the correct navigation item onto the stack for it. It takes an optional `fallback` parameter now for the editor case.

`findAndPushLinkDetail` will find the best address for a link, then call through to `findAndPushDetail`. It should populate `fallback` if possible.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -185,15 +184,15 @@ struct DetailStackModel: Hashable, ModelProtocol {
             return findAndPushLinkDetail(
                 state: state,
                 environment: environment,
-                context: context,
-                link: link
+                link: link,
+                context: context
```
bfollington, line 188: flip back

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -288,61 +287,98 @@ struct DetailStackModel: Hashable, ModelProtocol {
...
-        //
-        // This is needed in the viewer but address will always based
-        // on our sphere in the editor case.
-        guard case let .petname(basePetname) = address.peer else {
-            return link.slashlink
+    public static func findBestAddressForLink(
+        link: SubSlashlinkLink,
+        context: Peer?,
+        environment: TestableEnvironment
+    ) async throws -> Slashlink {
```
bfollington, line 294: All the logic for mapping a link to a navigable address is now contained in this function. There are tests for this, but not to full coverage since traversal is impossible in testing.

Did my best to document the thinking, please flag any confusion or tweaks to make.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -288,61 +287,85 @@ struct DetailStackModel: Hashable, ModelProtocol {
...
-        // bare slashlinks relative to the sphere they belong to.
-        //
-        // This is needed in the viewer but address will always based
-        // on our sphere in the editor case.
-        guard case let .petname(basePetname) = address.peer else {
-            return link.slashlink
+    public static func findBestAddressForLink(
+        context: Peer?,
+        link: SubSlashlinkLink,
+        environment: TestableEnvironment
```
bfollington, line 293: All the logic for mapping a link to a navigable address is now contained in this function. There are tests for this, but not to full coverage since traversal is impossible in testing atm.

Did my best to document the thinking, please flag any confusion or tweaks to make.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -288,61 +287,85 @@ struct DetailStackModel: Hashable, ModelProtocol {
...
+        var did: Did? = nil
+        if let petname = slashlink.petname {
+            do {
+                // Any errors during traversal mean we should give up
+                let sphere = try await environment.noosphere.traverse(petname: petname)
+                did = try await sphere.identity()
+            } catch {
+                did = nil
+            }
+        }
```
gordonbrander, line 320: Nit: this could be rewritten in expression form as

```swift
let did = try? await Func.run {
    let sphere = try await environment.noosphere.traverse(petname: petname)
    return try await sphere.identity()
}
```

I find expression-form code easier to follow when possible.

No action required.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
gordonbrander: This looks good. A few questions:

- I see 3x calls to Noosphere. Is this more than we were making before, or the same?
- AddressBookService uses Noosphere under the hood. Could this method could become an extension of NoosphereService or Sphere instead of an environment function?

xcode/Subconscious/Shared/Components/Detail/MemoViewerDetailView.swift
```diff
@@ -42,13 +42,8 @@ struct MemoViewerDetailView: View {
                 )
             case .loaded:
                 MemoViewerDetailLoadedView(
-                    title: store.state.title,
-                    user: store.state.owner,
-                    dom: store.state.dom,
-                    transcludePreviews: store.state.transcludePreviews,
+                    store: store,
```
gordonbrander, line 45: 👍

gordonbrander (approved):
LGTM, with a few questions and requests.

xcode/Subconscious/Shared/Components/BacklinksView.swift
bfollington: > Does passing Peer mean we will have to resolve peernames to DIDs more frequently in the code?

Technically yes but in practice using `Peer` allows us to provide a `DID` if we have it or fall back to a `Petname` when we do not. The issue is: when viewing a given note, any links in that note must be resolved to a DID at _some stage_ for us to visit it. For anything we've indexed (i.e. transcludes) we know the DID from the SQLite records but for any other link we either look it up now or later. The perf is the same.

gordonbrander:
Tested, working.

Issue #1008, opened 2023-11-30
IO error: lock hold by current process


bfollington:
Encountered by @justinabrahms after trying to sync (with a period of inactivity since last sync).

![image](https://github.com/subconsciousnetwork/subconscious/assets/5009316/e3b4cefd-bb09-40aa-9933-d36fe3fc6106)

Subsequent attempts failed with different messages, seemingly intermittently:

![image](https://github.com/subconsciousnetwork/subconscious/assets/5009316/5a485185-832c-426d-8c26-7b557be9c74a)

jsantell:
WRT the LOCK file issue: it could be possible on backgrounding, rocks was closing its background threads (which fcntl will release the LOCK file upon completing), and on reopening Subconscious, the database (lock) was still in use -- https://github.com/subconsciousnetwork/noosphere/pull/734 may fix this issue in this case

PR #1016, opened 2023-12-06
Allow user-configuration of Noosphere log detail

bfollington:
1. Allow user configuration of Noosphere log detail (persisted using AppDefaults)
2. Pipe `stdout` to a logger to capture Noosphere logs

This is per-process on iOS and you can see the result via Console.app

Intended to help with debugging https://github.com/subconsciousnetwork/subconscious/issues/1008


gordonbrander (approved):
LGTM

PR #1018, opened 2023-12-11
Prefer basic Slug constructor for DB queries

### LINKED ISSUE #980: Slug in transclude incorrect for /ttrpg/favorites
I have a note at /ttrpg/favorites. When displayed as a related note, the transclude has the incorrect slug even though it appears to show the expected content. Clicking on the transclude takes me to the incorrect slug.
IMG_0030

bfollington:
Fixes #980 

`formatting:` drops deep slashlinks which is unsuitable for reading backlinks, we were not relying on this behaviour in any calls to `toSlug()` so I've updated the constructor within that helper.

### DIFF
--- a/xcode/Subconscious/Shared/Models/Slug.swift
+++ b/xcode/Subconscious/Shared/Models/Slug.swift
@@ -181,6 +181,6 @@ extension URL {
 
 extension String {
     func toSlug() -> Slug? {
-        Slug(formatting: self)
+        Slug(self)
     }
 }


gordonbrander (commented):
@bfollington I just realized this is probably why Wikilinks eat deep slashlinks.

What if we updated `Slug(formatting:)` to treat deep slashlinks nicely? Does this light anything on fire? If not, I would like to do that as well.

Issue #992, opened 2023-11-10
Forward slashes are stripped when editing a link


cdata:
I can create a note with forward slash in the slug by clicking on a link to it from another note.

However, if I try to use a forward slash when editing a slug, it gets stripped.

I expect the forward slash to be allowed or disallowed in both places (preferably allowed).

PR #1028, opened 2023-12-15
Fix edit link menu

bfollington:
It turns out when I refactored our form fields to use `ViewStore` I accidentally routed around the cursors which means some actions (like populating rename results) were not being dispatched.

I've restored cursors at every use site of `ValidatedFormField`.

Fixes https://github.com/subconsciousnetwork/subconscious/issues/992
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1021

### DIFF
--- a/xcode/Subconscious/Shared/Services/DatabaseService.swift
+++ b/xcode/Subconscious/Shared/Services/DatabaseService.swift
@@ -893,7 +893,9 @@ final class DatabaseService {
         // audience as current.
         let queryAddress = Func.run({
             let audience = current.toAudience()
-            return Slug(formatting: query)?.toSlashlink(audience: audience)
+            // Try literal slug first to preserve slashes
+            // If that fails, try best effort formatting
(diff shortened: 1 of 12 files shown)


xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -893,7 +893,9 @@ final class DatabaseService {
         // audience as current.
         let queryAddress = Func.run({
             let audience = current.toAudience()
-            return Slug(formatting: query)?.toSlashlink(audience: audience)
+            // Try literal slug first to preserve slashes
+            // If that fails, try best effort formatting
+            return (Slug(query) ?? Slug(formatting: query))?.toSlashlink(audience: audience)
```
bfollington, line 898: This fixes https://github.com/subconsciousnetwork/subconscious/issues/992

gordonbrander: 👍 

Separately, I think we should make formatting logic more lenient on slashes.

gordonbrander (approved):
LGTM

Issue #1035, opened 2023-12-18
Drafts should derive slugs after first newline, not after timeout


gordonbrander:
This puts users in more control over when slug is derived. The current timeout is 15s, which is long, but can still feel like a race sometimes.

PR #1036, opened 2023-12-18
Do less work when block editor is disabled

gordonbrander:
This PR re-works block editor setup and polling logic to do nothing when block editor is disabled.

Previously this logic was harmless, but was still running in the background.

The "pull" polling approach implemented here will give us more flexibility in future to do things like #1035 

Relies on https://github.com/subconsciousnetwork/ObservableStore/pull/45


bfollington (approved):
Approved, make sure to update ObservableStore version before merging

gordonbrander:
ObservableStore release landed https://github.com/subconsciousnetwork/ObservableStore/releases/tag/v0.6.1

Updated PR to require ObservableStore@v0.6.1.

PR #1006, opened 2023-11-30
Block Editor Transcludes

gordonbrander:
Picking up where #1000 left off.

PR #1040, opened 2023-12-20
Fix transclude link rebasing

bfollington:
Discovered a bug we introduced in #1006 while I was playtesting: tapping transcludes resulted in double rebasing. For example:

1. viewing `@gordon/test`
2. tap a transclude block for `@gordon/hello`
3. I go to `@gordon.gordon/hello`

We need to pass more context down to `SubtextView` so it can appropriately handle link taps (need rebasing) vs. transclude taps (do not need rebasing).

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/EntryRow.swift
+++ b/xcode/Subconscious/Shared/Components/Common/EntryRow.swift
@@ -18,11 +18,9 @@ struct EntryRow: View {
     var body: some View {
         VStack(alignment: .leading, spacing: AppTheme.unit) {
             SubtextView(
+                peer: entry.toPeer(),
                 subtext: entry.excerpt,
-                onLink: { link in
-                    let rebasedLink = link.rebaseIfNeeded(peer: entry.toPeer())
-                    onLink(rebasedLink)
-                }
+                onLink: onLink
             )
             .font(.callout)
             .multilineTextAlignment(.leading)
--- a/xcode/Subconscious/Shared/Components/Common/SubtextView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/SubtextView.swift
@@ -14,6 +14,7 @@ struct RenderableBlock: Hashable {
 
 struct SubtextView: View {
     private static var renderer = SubtextAttributedStringRenderer()
+    var peer: Peer? = nil
     var subtext: Subtext
     var transcludePreviews: [Slashlink: EntryStub] = [:]
     var onLink: (EntryLink) -> Void
@@ -91,7 +92,8 @@ struct SubtextView: View {
             guard let link = url.toSubSlashlinkLink()?.toEntryLink() else {
                 return .systemAction
             }
-            onLink(link)
+            
+            onLink(link.rebaseIfNeeded(peer: peer))
             return .handled
         })
     }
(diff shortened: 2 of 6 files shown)


xcode/Subconscious/Shared/Components/Common/EntryRow.swift
```diff
@@ -18,11 +18,9 @@ struct EntryRow: View {
     var body: some View {
         VStack(alignment: .leading, spacing: AppTheme.unit) {
             SubtextView(
+                peer: entry.toPeer(),
```
bfollington, line 21: Not sure what the right name for this parameter is. We want the peer so we can selectively rebase depending on whether we tapped a link, a transclude or a link _within_ a transclude and provide a simple interface to the caller.

bfollington: fwiw this is why there were two callbacks passed before, to handle these two cases, but I think this approach is better.

gordonbrander: @bfollington question about conceptual approach... Since transcludes are always embedded within a sphere, and links are always embedded within a document or a transclude, shouldn't we be able to rebase at each level and present the same interface to consumers?

E.g. Alice transcludes Bob's note which has a link to Carol. The rebased Carol link, as exposed via `onLink` is `@carol.bob.alice/foo`.

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
```diff
@@ -91,7 +92,8 @@ struct SubtextView: View {
             guard let link = url.toSubSlashlinkLink()?.toEntryLink() else {
                 return .systemAction
             }
-            onLink(link)
+            
+            onLink(link.rebaseIfNeeded(peer: peer))
```
gordonbrander, line 96: Rebasing link taps but not embedded transclusions seems conceptually surprising (see https://github.com/subconsciousnetwork/subconscious/pull/1040/files#r1432707403). Can you help walk me through what is going on here?

xcode/Subconscious/Shared/Components/Common/EntryRow.swift
gordonbrander: Separately, adding a `peer` property to SubtextView seems good! Subtext is always going to belong to some sphere, and because petnames are relative, linking is always relative. We should probably get links out of Subtext that point to the correct sphere, by default, and passing in a peer in forces correctness.

gordonbrander (commented):
A few questions @bfollington

xcode/Subconscious/Shared/Components/Common/SubtextView.swift
bfollington: The short version is that when we use a `TranscludeView` we pass it an `EntryStub` which comes from the local SQLite index (as a result of fetching transcludes). The address for this `EntryStub` is already fully resolved, or else we wouldn't present the transclude in the first place (because it wouldn't be in the index).

We _could_ work around this by intentionally ignoring the address on the `EntryStub` and deriving it another way, but that is just wasted work ultimately.

Links are different, until we tap a link we do not always know what's on the other end (or else we would typically present a transclude).

bfollington: ^ not to say any of this _has to_ or _must_ be this way, just describing the status quo and why transcludes work differently.

gordonbrander (approved):
LGTM

PR #1044, opened 2023-12-20
Block editor: save when editor is backgrounded or un-presented

### LINKED ISSUE #1041: Block editor should save when exiting editor
In addition to saving on timeout and saving when keyboard is dismissed, we should be saving when exiting an editing session by swiping back or hitting the back button, or loading a new note.

Part of #214

gordonbrander:
Fixes #1041

## Testing

Setup:

- Go to Settings > Developer > Block Editor, and turn on block editor.

Scenario 1:

- Create a new note or edit an existing note
- Type some text
- Swipe back

What should happen: note should be saved.

Scenario 2:

- Create a new note or edit an existing note
- Type some text
- Background the app

What should happen: note should be saved.


bfollington (approved):
LGTM, works in testing too :)

PR #1047, opened 2023-12-21
Block editor: change block type

### LINKED ISSUE #1042: Block editor should allow you to change block types

gordonbrander:
Fixes #1042

### DIFF
--- a/xcode/Subconscious/Shared/AppTheme.swift
+++ b/xcode/Subconscious/Shared/AppTheme.swift
@@ -185,3 +185,16 @@ extension UIColor {
         UIColor(Color.accentColor)
     }
 }
+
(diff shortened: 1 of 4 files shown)


xcode/Subconscious/Shared/AppTheme.swift
```diff
@@ -185,3 +185,16 @@ extension UIColor {
         UIColor(Color.accentColor)
     }
 }
+
+extension ShadowStyle {
+    static func brandShadowLg(
```
gordonbrander, line 190: Introduce the concept of shadow presets. In future we will want to systematize our shadows so as to have a coherent sense of distance and lighting.

Issue #1061, opened 2024-01-08
I would like to customize the color of my notes.


bfollington:
We can store the color of a note as a header and let user's customize it via the meta sheet. 

Proof of concept: https://github.com/subconsciousnetwork/subconscious/tree/2023-12-20-prompts

PR #1064, opened 2024-01-08
User customizable note colors

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1061

- Store color as header
- Allow updating color from metasheet
- Read headers from DB
- Refresh all views when color changes
- Add color to `WellKnownHeaders`

This change dials back the use of color in the editor/viewer in anticipation of the next chunk of work: designing the mini-editor & full editor using card styles.

https://github.com/subconsciousnetwork/subconscious/assets/5009316/84505a31-78bb-47e1-bfbe-924cfa59d60d

### DIFF
--- a/xcode/Subconscious/Shared/AppTheme.swift
+++ b/xcode/Subconscious/Shared/AppTheme.swift
@@ -109,7 +109,7 @@ extension Color {
         .opacity(0.5)
     static let primaryButtonBackgroundDisabled = primaryButtonBackground
         .opacity(0.3)
-    static let primaryButtonText = SwiftUI.Color.accentColor
+    static var primaryButtonText: Color { SwiftUI.Color.accentColor }
     static let primaryButtonTextPressed = primaryButtonText.opacity(0.5)
     static let primaryButtonTextDisabled = primaryButtonText.opacity(0.3)
     static let fabBackground = primaryButtonBackground
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -302,7 +302,7 @@ enum AppAction: Hashable {
     case mergeEntry(parent: Slashlink, child: Slashlink)
     case moveEntry(from: Slashlink, to: Slashlink)
     case updateAudience(address: Slashlink, audience: Audience)
-    case assignColor(addess: Slashlink, color: NoteColor)
+    case assignColor(addess: Slashlink, color: ThemeColor)
     
     // These notifications will be passe down to child stores to update themselves accordingly.
     case succeedSaveEntry(address: Slashlink, modified: Date)
@@ -310,7 +310,7 @@ enum AppAction: Hashable {
     case succeedMoveEntry(from: Slashlink, to: Slashlink)
     case succeedMergeEntry(parent: Slashlink, child: Slashlink)
     case succeedUpdateAudience(MoveReceipt)
-    case succeedAssignNoteColor(address: Slashlink, color: NoteColor)
+    case succeedAssignNoteColor(address: Slashlink, color: ThemeColor)
     case failSaveEntry(address: Slashlink, error: String)
     case failDeleteMemo(String)
     case failMoveEntry(from: Slashlink, to: Slashlink, error: String)
@@ -3121,7 +3121,7 @@ struct AppModel: ModelProtocol {
         state: Self,
         environment: Environment,
         address: Slashlink,
-        color: NoteColor
+        color: ThemeColor
     ) -> Update<Self> {
         let fx: Fx<Action> = Future.detached {
             try await environment.data.assignNoteColor(
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -302,7 +302,7 @@ enum AppAction: Hashable {
     case mergeEntry(parent: Slashlink, child: Slashlink)
     case moveEntry(from: Slashlink, to: Slashlink)
     case updateAudience(address: Slashlink, audience: Audience)
-    case assignColor(addess: Slashlink, color: ThemeColor)
+    case assignColor(address: Slashlink, color: ThemeColor)
     
     // These notifications will be passe down to child stores to update themselves accordingly.
     case succeedSaveEntry(address: Slashlink, modified: Date)
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -302,6 +302,7 @@ enum AppAction: Hashable {
     case mergeEntry(parent: Slashlink, child: Slashlink)
     case moveEntry(from: Slashlink, to: Slashlink)
     case updateAudience(address: Slashlink, audience: Audience)
+    case assignColor(addess: Slashlink, color: NoteColor)
     
     // These notifications will be passe down to child stores to update themselves accordingly.
     case succeedSaveEntry(address: Slashlink, modified: Date)
@@ -309,11 +310,13 @@ enum AppAction: Hashable {
     case succeedMoveEntry(from: Slashlink, to: Slashlink)
     case succeedMergeEntry(parent: Slashlink, child: Slashlink)
     case succeedUpdateAudience(MoveReceipt)
+    case succeedAssignNoteColor(address: Slashlink, color: NoteColor)
     case failSaveEntry(address: Slashlink, error: String)
     case failDeleteMemo(String)
     case failMoveEntry(from: Slashlink, to: Slashlink, error: String)
     case failMergeEntry(parent: Slashlink, child: Slashlink, error: String)
     case failUpdateAudience(address: Slashlink, audience: Audience, error: String)
+    case failAssignNoteColor(address: Slashlink, error: String)
     
     case succeedLogActivity
     case failLogActivity(_ error: String)
@@ -1220,7 +1223,8 @@ struct AppModel: ModelProtocol {
                 state: state,
                 environment: environment
             )
-        case .succeedMoveEntry, .succeedMergeEntry, .succeedLogActivity, .succeedUpdateAudience:
+        case .succeedMoveEntry, .succeedMergeEntry, .succeedLogActivity, .succeedUpdateAudience,
+                .succeedAssignNoteColor:
             return Update(state: state)
         case .succeedSaveEntry(address: let address, modified: let modified):
             return succeedSaveEntry(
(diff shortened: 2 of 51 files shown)


xcode/Subconscious/Shared/Components/Deck/DeckView.swift
```diff
@@ -524,7 +543,7 @@ struct DeckModel: ModelProtocol {
             var model = state
             model.deck = deck
             model.seen = Set(deck.compactMap { card in card.entry })
-            model.pointer = startIndex.clamp(min: 0, max: deck.count - 1)
+            model.pointer = startIndex.clamp(min: 0, max: max(0, deck.count - 1))
```
bfollington, line 546: Caught an error here during testing.

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -282,19 +283,12 @@ enum MemoEditorDetailNotification: Hashable {
     /// Request detail from any audience scope
     case requestFindLinkDetail(EntryLink)
     
+    case requestSaveEntry(MemoEntry)
     case requestDelete(Slashlink?)
-    
     case requestMoveEntry(from: Slashlink, to: Slashlink)
-    case succeedMoveEntry(from: Slashlink, to: Slashlink)
```
bfollington, line 288: None of these success notifications were in use anymore.

EditorDetailNotifications only flow up the tree, but these success events flow down the tree from app since my refactor.

xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -602,19 +610,13 @@ extension DetailStackAction {
...
         case let .requestFindLinkDetail(link):
             return .findAndPushLinkDetail(link)
-        case let .succeedMoveEntry(from, to):
-            return .succeedMoveEntry(from: from, to: to)
-        case let .succeedMergeEntry(parent, child):
-            return .succeedMergeEntry(parent: parent, child: child)
-        case let .succeedSaveEntry(address, modified):
-            return .succeedSaveEntry(address, modified)
-        case let .succeedUpdateAudience(receipt):
-            return .succeedUpdateAudience(receipt)
```
bfollington, line 617: Again, these were unused.

xcode/Subconscious/Shared/AppTheme.swift
```diff
@@ -109,7 +109,7 @@ extension Color {
         .opacity(0.5)
     static let primaryButtonBackgroundDisabled = primaryButtonBackground
         .opacity(0.3)
-    static let primaryButtonText = SwiftUI.Color.accentColor
+    static var primaryButtonText: Color { SwiftUI.Color.accentColor }
```
gordonbrander, line 112: Nice!

xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -302,18 +302,21 @@ enum AppAction: Hashable {
     case mergeEntry(parent: Slashlink, child: Slashlink)
     case moveEntry(from: Slashlink, to: Slashlink)
     case updateAudience(address: Slashlink, audience: Audience)
+    case assignColor(addess: Slashlink, color: NoteColor)
```
gordonbrander, line 305: I think I would recommend just passing `Color` at the level of actions. `NoteColor` can act as a constant we use when assigning colors.

Rationale: the plumbing then supports any color. We enforce defaults at the io boundary. This would allow users to hack the color headers, for example.

If the intent is to support dynamically adaptive colors for light/dark, I believe SwiftUI supports this through the asset bundle (specify light/dark variants of color constants).

Alternatively, if we wanted to keep this in code, we might create a more general type like `AdaptiveColor`:

```swift
struct AdaptiveColor: Hashable {
    var light: Color
    var dark: Color

    func color(for scheme: ColorScheme) -> Color {
        switch scheme {
        case .light:
            return light
        case .dark:
            return dark
        }
    } 
}
```

xcode/Subconscious/Shared/Components/Deck/DeckView.swift
```diff
@@ -119,6 +119,8 @@ enum DeckAction: Hashable {
     case succeedMergeEntry(parent: Slashlink, child: Slashlink)
     case requestUpdateAudience(_ address: Slashlink, _ audience: Audience)
     case succeedUpdateAudience(_ receipt: MoveReceipt)
+    case requestAssignNoteColor(_ address: Slashlink, _ color: NoteColor)
+    case succeedAssignNoteColor(_ address: Slashlink, _ color: NoteColor)
```
gordonbrander, line 123: I think I would recommend just passing Color at the level of actions. NoteColor can act as a constant we use when assigning colors.

xcode/Subconscious/Shared/Components/Deck/DeckView.swift
gordonbrander: Bounds checking. Nice.

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -282,19 +291,12 @@ enum MemoEditorDetailNotification: Hashable {
...
-    
     case requestMergeEntry(parent: Slashlink, child: Slashlink)
-    case succeedMergeEntry(parent: Slashlink, child: Slashlink)
-    
-    case requestSaveEntry(MemoEntry)
-    case succeedSaveEntry(address: Slashlink, modified: Date)
-    
     case requestUpdateAudience(address: Slashlink, audience: Audience)
-    case succeedUpdateAudience(_ receipt: MoveReceipt)
+    case requestAssignNoteColor(_ address: Slashlink, _ color: NoteColor)
```
gordonbrander, line 299: Recommend passing Color in actions

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -610,8 +626,11 @@ struct MemoEditorDetailModel: ModelProtocol {
         modified: Date.distantPast,
         fileExtension: ContentType.subtext.fileExtension
     )
+    var color: NoteColor? = nil
```
gordonbrander, line 629: Recommend storing `Color`

xcode/Subconscious/Shared/Parsers/Header.swift
```diff
@@ -14,6 +14,7 @@ enum HeaderName: String {
     case created = "Created"
     case modified = "Modified"
     case fileExtension = "File-Extension"
+    case color = "Color"
```
gordonbrander, line 17: IMO the header name should be `Theme-Color` and the case name `themeColor` 

Prior art: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

gordonbrander: I don't think the header should store constant names. Instead, we should consider storing hex values.

Should we include support for color variants?

```
Theme-Color: #abc123
Theme-Color-Dark: #abc-123
```

bfollington:
RE: https://github.com/subconsciousnetwork/subconscious/pull/1064#discussion_r1446306648

@gordonbrander I think it's a bit more complicated than this. For each of the five note variants we have _four_ colours:
- Light BG
- Light Highlight
- Dark BG
- Dark Highlight

I can see several issues with storing literal color values:
1. schema lock-in: storing 4 headers is quite heavy but and also locks us in to how color presents itself going forward
2. color drift, if we change the defaults slightly the old colors will persist 
3. injecting malicious colors into other user's deck
4. color space, on web vs iOS we are in different color spaces and may want to tailor colors based on that
5. accessibility, it makes high contrast and color-blind friendly modes 

I personally think of `NoteColor` here more like a simple bucket A, B, C, D or E that I can sort a note into, and perhaps we should use actually generic letter labels instead of colors as the cases of `NoteColor`? The concern of how that bucket is translated to onscreen pixels feels like a view-level responsibility to me. This would allow us to "theme" the whole app and re-define the colors for A...E at will, users could even customize the colors for themselves.

If we want to support user-defined custom colors per-note right now I think we should consider a `NoteColor.custom(String)` case that the client can choose to support to varying degrees.

xcode/Subconscious/Shared/Parsers/Header.swift
bfollington: Updated the header name, see https://github.com/subconsciousnetwork/subconscious/pull/1064#issuecomment-1884141408 RE: storing hex.

xcode/Subconscious/Shared/Components/AppView.swift
bfollington: I moved the colors into the asset bundle, see response here https://github.com/subconsciousnetwork/subconscious/pull/1064#issuecomment-1884141408

xcode/Subconscious/Shared/Components/Deck/DeckTheme.swift
```diff
@@ -52,82 +52,104 @@ extension DeckTheme {
...
+    case a
+    case b
+    case c
+    case d
+    case e
+}
+
+public extension Color {
+    static var cardThemeColorA: Color {
+        Color("CardThemeColorA")
```
bfollington, line 68: There is a new syntax `Color(.cardThemeColorA)` on XCode 15+ but we need these string names for XCode 14.

gordonbrander (changes requested):
> RE: [#1064 (comment)](https://github.com/subconsciousnetwork/subconscious/pull/1064#discussion_r1446306648)
> 
> @gordonbrander I think it's a bit more complicated than this. For each of the five note variants we have _four_ colours:
> 
> * Light BG
> * Light Highlight
> * Dark BG
> * Dark Highlight
> 
> I can see several issues with storing literal color values:
> 
> 1. schema lock-in: storing 4 values seems hefty but also locks us in to how color presents itself going forward
> 2. color drift, if we change the defaults slightly the old colors will persist in the remaining notes
> 3. injecting malicious colors into another user's deck
> 4. color space, on web vs iOS we are in different color spaces and may want to tailor colors based on that
> 5. accessibility, it makes high contrast and color-blind friendly modes difficult to implement/make guarantees around
> 
> I personally think of `NoteColor` here more like a simple bucket A, B, C, D or E that I can sort a note into, and perhaps we should use actually generic letter labels instead of colors as the cases of `NoteColor`? The concern of how that bucket is translated to onscreen pixels feels like a view-level responsibility to me. This would allow us to "theme" the whole app and re-define the colors for A...E at will, users could even customize the colors for themselves.
> 
> If we want to support user-defined custom colors per-note right now I think we should consider a `NoteColor.custom(String)` case that the client can choose to support to varying degrees.
> 
> I moved all the card colours into the asset bundle, renamed `NoteColor` to `ThemeColor` and changed the labels to generic letters as-per the above.

Ok, this makes sense for now, although I suspect we may want to consider how to handle custom color values over the long term.

Since these themes are an app-specific concern, I have one more request: let's change the header key to `Subconscious-Theme`. The value can be a unique key that corresponds to the theme (`a`, `b`, `c`, ...).

We may want to use numbers rather than letters for the theme keys to avoid uppercase/lowercase ambiguity.

Otherwise LGTM

gordonbrander (approved):
LGTM 🥳

Issue #1060, opened 2024-01-08
Authorization list should display the name of each authorization


bfollington:
<img width="363" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/79aa5680-c2ea-4f2c-8c4d-051097ce3af4">

We have the human readable name available, we should just render it.

PR #1066, opened 2024-01-09
Render names in authorization list

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1060

- Fetch name of authorizations
- Render in list view
- Move to `.contextMenu` for revocation and sharing

https://github.com/subconsciousnetwork/subconscious/assets/5009316/ddb7206d-45b1-40bb-9cec-d891a41a55a8

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
@@ -12,6 +12,10 @@ import os
 
 public typealias Cid = String
 public typealias Authorization = String
+public struct NamedAuthorization: Equatable, Hashable {
+    public let name: String
+    public let authorization: Authorization
+}
 
 /// Describes a Sphere.
 /// See `Sphere` for a concrete implementation.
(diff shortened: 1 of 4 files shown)


xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
```diff
@@ -949,6 +968,34 @@ public actor Sphere: SphereProtocol, SpherePublisherProtocol {
         }
     }
     
+    public func authorizationName(authorization: Authorization) async throws -> String {
```
gordonbrander, line 971: Let's add a test for this method

gordonbrander (changes requested):
Looks good, just recommend adding one test

xcode/Subconscious/Shared/Services/Noosphere/Sphere.swift
bfollington: Good call, done!

gordonbrander (approved):
LGTM

Issue #1023, opened 2023-12-14
"Session" for Deck View


bfollington:
Track affinity towards different authors, tags and links with a rolling buffer to augment the behaviour of the deck.

bfollington:
Record swipes, indexing complete events, following a peer etc. in table
  - New table
  - type, category, date, message, metadata

PR #1067, opened 2024-01-10
Log key activities in SQLite

bfollington:
Trying not to over engineer anything here.

- Create a new `activity` SQLite table
- Log events we may be interested in responding to / surfacing:
    - Swipe choices in Deck view (step towards https://github.com/subconsciousnetwork/subconscious/issues/1023)
    - Follow / unfollow user actions
    - Index complete + how many changes

As mentioned above, this is a step towards offering more sophisticated behaviour in the Deck view over multiple sessions but it could also let us offer an "activity" view in the app where you can see updates like "`@gordon` published 4 notes".

https://github.com/subconsciousnetwork/subconscious/assets/5009316/c05dbb91-d3fa-47b1-bdff-60081f8761a0

### DIFF
--- a/xcode/Subconscious/Shared/Services/DatabaseService.swift
+++ b/xcode/Subconscious/Shared/Services/DatabaseService.swift
@@ -1391,39 +1391,6 @@ final class DatabaseService {
             ]
         )
     }
-    
-    func writeActivity(category: String, event: String, message: String, metadata: String?) throws {
-        guard self.state == .ready else {
-            return
-        }
-        
-        try database.execute(
-            sql: """
-            INSERT INTO activity (category, event, message, metadata)
-            VALUES (?, ?, ?, ?)
-            """,
-            parameters: [
-                .text(category),
-                .text(event),
-                .text(message),
-                .text(metadata)
-            ]
-        )
-    }
-}
-
-struct ActivityEvent<Data: Codable>: Codable {
-    let category: ActivityEventCategory
-    let event: String
-    let message: String
-    let metadata: Data?
-}
-
-enum ActivityEventCategory: String, Codable {
-    case system = "system"
-    case deck = "deck"
-    case note = "note"
-    case addressBook = "addressBook"
 }
 
 // MARK: Migrations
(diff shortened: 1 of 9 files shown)


xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1371,9 +1372,27 @@ final class DatabaseService {
         })
         .first
     }
+    
+    func writeActivity<Data: Codable>(event: ActivityEvent<Data>) throws {
```
gordonbrander, line 1376: Let's add a test for this method using in-memory SQL

gordonbrander, line 1376: Let's add a test for this method using in-memory SQL

xcode/Subconscious/Shared/Services/DatabaseService.swift
```diff
@@ -1555,6 +1574,21 @@ extension Config {
...
+            CREATE TABLE activity
+            (
+                id       INTEGER PRIMARY KEY AUTOINCREMENT,
+                category TEXT                           NOT NULL,
+                event    TEXT                           NOT NULL,
+                message  TEXT DEFAULT ''                NOT NULL,
+                metadata TEXT DEFAULT '{}'              NOT NULL,
+                created  TEXT DEFAULT CURRENT_TIMESTAMP NOT NULL
+            );
+            """
```
gordonbrander, line 1591: Is it possible to also add a test for these migrations using a fresh in-memory SQL DB? This is a question. I don't recall if I put together the code to make that work.

gordonbrander (changes requested):
One test request. Otherwise LGTM.

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Done!

xcode/Subconscious/Shared/Services/DatabaseService.swift
bfollington: Yep 👍

gordonbrander (approved):
LGTM!

Issue #1031, opened 2023-12-18
Tapping sync status indicator should reveal gateway settings


bfollington:
![image](https://github.com/subconsciousnetwork/subconscious/assets/5009316/26961614-3d43-4901-9010-8e00cf87aa33)

Raised by @justinabrahms, if it looks tappable it probably should be tappable.

PR #1068, opened 2024-01-11
Show settings when tapping sync indicator

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1031

- tap sync indicator to show settings
- surface sync failure in root settings view


gordonbrander (approved):
LGTM

Issue #991, opened 2023-11-10
Tapping tab while on tab should scroll to top


gordonbrander:
If your already on an app tab and you tap the same tab, it should scroll to top.

PR #1070, opened 2024-01-15
Tap tab button again to scroll to top

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/991

Pretty simple, using `ScrollReaderView` for iOS 16 support but we could clean this up a little using the newer iOS 17 APIs down the line.

- Tap tab button again to scroll to top of lists
- Move away from .enumerated()

https://github.com/subconsciousnetwork/subconscious/assets/5009316/ef31c136-a8f5-4465-ac4f-6c0c5b392140

https://github.com/subconsciousnetwork/subconscious/assets/5009316/0443cf7b-693e-4b6f-bd8d-d071e1f50c3e

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
+++ b/xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
@@ -21,7 +21,9 @@ struct EntryListView: View {
         if let entries = entries {
             if entries.count > 0 {
                 List {
-                    ForEach(entries) { entry in
+                    ForEach(entries.indices, id: \.self) { idx in
+                        let entry = entries[idx]
+                        
                         Button(
                             action: {
                                 onEntryPress(entry)
@@ -35,6 +37,7 @@ struct EntryListView: View {
                                 onLink: onLink
                             )
                         }
+                        .id(idx)
                         .buttonStyle(
                             EntryListRowButtonStyle(
                                 color: entry.color(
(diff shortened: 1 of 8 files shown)


xcode/Subconscious/Shared/Components/Deck/CardStackView.swift
```diff
@@ -287,8 +283,9 @@ struct CardStack: View {
             Spacer()
             GeometryReader { geo in
                 ZStack {
-                    ForEach(enumeratedDeck, id: \.element.id) {
-                        index, card in
+                    ForEach(deck.indices, id: \.self) { index in
+                        let card = deck[index]
+
```
bfollington, line 288: Apparently there are problems with using `enumerated()` this way that can result in `BAD_ACCESS` errors, see: https://stackoverflow.com/a/63145650

Using `indices` is apparently safer and seems equally readable.

xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
```diff
@@ -35,6 +37,7 @@ struct EntryListView: View {
                                 onLink: onLink
                             )
                         }
+                        .id(idx)
```
gordonbrander, line 40: This is likely to cause issues, as ids are meant to be stable identifiers tied to records. If the record changed but had same index, or if the same record changed index, we might see view invalidation bugs.

Should replace with a stable record ID of some sort.

gordonbrander (changes requested):
One note, otherwise LGTM

xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
bfollington: Good note, I've ditched this approach and inserted an `EmptyView` with a stable `.id()` at the top instead. This is easier to read as well imo, so win-win.

gordonbrander (approved):
LGTM

Issue #871, opened 2023-08-14
"Link to this" action


gordonbrander:
Others' notes and transcludes often create surprising sparks of insight. However there is currently no easy flow to create a note or add a link to an existing note.

Solutions:

- "Link to this" action
- Share menu actions

PR #1071, opened 2024-01-15
"Quote in new note" action

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/871

- MVP quoting feature

https://github.com/subconsciousnetwork/subconscious/assets/5009316/a0b1f522-0cf0-476c-8a3d-959e382d3fca

### DIFF
--- a/xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
+++ b/xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
@@ -612,6 +612,14 @@ extension DetailStackAction {
             return .requestUpdateAudience(address, audience)
         case let .requestAssignNoteColor(address, color):
             return .requestAssignNoteColor(address, color)
+        case let .requestQuoteInNewDetail(address):
+            return .pushDetail(
+                .editor(
+                    MemoEditorDetailDescription(
+                        fallback: "\(address.markup)\n\n"
+                    )
+                )
+            )
             
         case let .requestDetail(detail):
             return .pushDetail(detail)
@@ -637,7 +645,13 @@ extension DetailStackAction {
                 )
             )
         case let .requestQuoteInNewDetail(address):
-            return .pushDetail(.editor(MemoEditorDetailDescription(fallback: "\(address.markup)\n\n")))
+            return .pushDetail(
+                .editor(
+                    MemoEditorDetailDescription(
+                        fallback: "\(address.markup)\n\n"
+                    )
+                )
+            )
         }
     }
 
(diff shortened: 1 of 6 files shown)


xcode/Subconscious/Shared/Components/Detail/DetailStackView.swift
```diff
@@ -612,6 +612,14 @@ extension DetailStackAction {
...
         case let .requestAssignNoteColor(address, color):
             return .requestAssignNoteColor(address, color)
+        case let .requestQuoteInNewDetail(address):
+            return .pushDetail(
+                .editor(
+                    MemoEditorDetailDescription(
+                        fallback: "\(address.markup)\n\n"
+                    )
+                )
+            )
```
bfollington, line 622: This results in an untitled editor, preloaded with the link and the cursor autofocused in a new block. 

There is one UX rough edge, if you back out of the new editor this opens, then the app will automatically save the note... which creates a "garbage" note. Perhaps we should have some clever logic that never saves a note unless it contains something other than / in addition to a single slashlink?

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetailMetaSheetView.swift
```diff
@@ -54,7 +54,26 @@ struct MemoEditorDetailMetaSheetView: View {
...
+                        
+                        if let address = store.state.address {
+                            Button(
+                                action: {
+                                    store.send(.requestQuoteInNewNote(address))
+                                },
+                                label: {
+                                    Label(
+                                        "Quote in new note",
+                                        systemImage: "quote.opening"
```
bfollington, line 68: I welcome discussion on the best wording and icon for this action.

To me "linking" is inserting a slashlink into an existing note and "quoting" is creating a new note that references / reflects on another note (using linking).

bmann:
Clipping might be another word. I’m clipping it into my own notebook.

(This looks good!)

gordonbrander (approved):
LGTM.

My one UX feedback is that I think it might be nicer to place the cursor above the link, with the link on it's own "paragraph" below? I'm imagining what this will look like with transcludes support, and by analogy to quote tweet, you probably want your thoughts above the transcluded content, not below it.

bfollington:
> LGTM.
> 
> My one UX feedback is that I think it might be nicer to place the cursor above the link, with the link on it's own "paragraph" below? I'm imagining what this will look like with transcludes support, and by analogy to quote tweet, you probably want your thoughts above the transcluded content, not below it.

I implemented this and I think it is better, going to run with putting the cursor ahead of the quoted content.

PR #1079, opened 2024-01-26
Fix: tapping links to drafts could (erroneously) open the viewer instead of the editor

bfollington:
Our `findBestLink` function was not considering `Did.local` as a possibility. Updated the code + tests to cover these cases.

I found this in-app in two places:
1. tapping a draft in the deck view opens in the viewer (uneditable)
2. tapping a backlink to a draft (in the related notes section) opened the viewer

### DIFF
--- a/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
+++ b/xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
@@ -49,7 +49,7 @@ extension NoosphereService {
         
         // We _could_ also choose to simply bail out and navigate to the address without
         // traversing, at the expense of the clever redirect.
-        var did: Did? = nil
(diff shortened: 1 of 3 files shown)


xcode/Subconscious/Shared/Services/Noosphere/NoosphereService.swift
```diff
@@ -62,8 +62,13 @@ extension NoosphereService {
             return slashlink
         }
         
+        // Preserve local DID
+        guard !did.isLocal else {
+            return slashlink
+        }
+        
         // Is this address ours? Trim off the peer
-        guard did != ourIdentity else {
```
bfollington, line 66: I thought this was clearer but obviously not an essential change

gordonbrander (approved):
Good catch!

PR #1082, opened 2024-01-26
Allow background to show through in empty view

bfollington:


gordonbrander (approved):
LGTM!

Request a follow-up where we vertically center these views if possible.

PR #1089, opened 2024-01-30
Enable sentry performance tracing integration

bfollington:
I've followed the documentation (https://docs.sentry.io/platforms/apple/performance/) and it seems to be working when I examine the Sentry SDK logs, but no data in the dashboard yet. Maybe there's latency, I'll wait and see.


gordonbrander (approved):
LGTM

PR #1090, opened 2024-01-30
Re-implement basic prompts with Tracery

gordonbrander:
This give us a bit more flexibility in how we generate these prompts, creating opportunities to generate prompts that are cheap yet feel more organic. Also, it will likely come in useful when crafting AI prompts in the future.

- Ported prompts over to Tracery Grammar
- Introduced PromptService.
  - PromptService is an actor. This is currently not important, but matches our other services and could allow us to introduce thread-safe state in future. It also forces us to interact with the methods over an async boundary, meaning we could introduce things like HTTP API calls to AI in future.

About the Tracery microlibrary:

- Tracery will stop flattening after exceeding a hard-coded max recursion depth constant (arbitrarily set at 500 which is small but big enough)
  - Test included
- Tracery init allows passing a dictionary of postprocessing modifiers for tags (e.g. `#start.ing#`), but I haven't implemented modifiers-eng (https://github.com/gordonbrander/spandrel/blob/main/modifiers-eng.js)


gordonbrander:
@bfollington did our swift lint settings change recently? I'm getting failures from SwiftLint. However, they seem to be coming from files I did not change.

bfollington (approved):
LGTM!

PR #1094, opened 2024-02-04
Re-check gateway provisioning status if sync fails but we DO have a gateway ID

bfollington:
We ran into an issue where some gateways failed to come online in production and we timed out waiting. We've fixed the backend issue since. The scenario is recoverable by a user opening their gateway settings and hitting the "Check Provisioning Status" button but we can try to auto-detect and resolve the issue too.

If a sync fails but we _do_ have a gateway ID then it means provisioning must have timed out in the past. 

In the MTG future the provisioning time should drop to near-instant hopefully making this even less of an issue.

@gordonbrander can you think of a better condition to detect this situation? The user has redeemed an invite code for a gateway but they _haven't_ received the gateway URL yet. It's still going to be set to `127.0.0.1:4443` for anyone hit by this.

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -2130,17 +2130,27 @@ struct AppModel: ModelProtocol {
         var model = state
         model.lastGatewaySyncStatus = .failed(error)
         
+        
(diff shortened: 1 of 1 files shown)


gordonbrander (approved):
LGTM

PR #1096, opened 2024-02-04
Use LazyVStack for all unknown-length lists

bfollington:
Greatly improves performance in the profile view for users with many notes.


gordonbrander (approved):
Nice!

PR #1098, opened 2024-02-05
Experiment: TraceryDisambiguator

gordonbrander:
This gives us a basic way to customize prompts based on inputs.

- TraceryDisambiguator is essentially a router
  - Tries a succession of routes
  - Collects up to `n` matches
- A route is a protocol that implements `match(:) async -> TraceryContext?`
  - `RegexRoute` is a concrete implementation of the protocol and allows regular-expression-based matching.
- TraceryContext contains a grammar and a start string
  - The start string is used to begin flattening
  - The grammar is used in PromptService to overwrite keys on our base grammar. This allows us to save "state" extracted from the input to the grammar.

RegexRoute should allow us to do simple keyword-based dispatch of prompts.

Future improvements:

- Could be expanded to tokenize, lemmatize, or otherwise preprocess text before handing to routes
- Could be expanded support other kinds of router, including on-device ML classifier, or even LLM.

### DIFF
--- a/xcode/Subconscious/Shared/Library/NaturalLanguage/Tracery.swift
+++ b/xcode/Subconscious/Shared/Library/NaturalLanguage/Tracery.swift
@@ -79,3 +79,32 @@ public struct Tracery {
         flatten(depth: 0, grammar: grammar, start: start)
     }
 }
+
+extension Grammar {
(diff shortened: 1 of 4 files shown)


gordonbrander:
Picking up work here https://github.com/subconsciousnetwork/subconscious/pull/1101

Issue #1102, opened 2024-02-06
Drafts without a slug have no background


bfollington:
<img width="436" alt="image" src="https://github.com/subconsciousnetwork/subconscious/assets/5009316/b7c8401e-dbce-4f1e-a8af-af3cfa1ad71c">

Oversight from https://github.com/subconsciousnetwork/subconscious/pull/1081, we should just pick a color to default to.

PR #1103, opened 2024-02-06
Fall back to `ThemeColor.a` in editor if there's no slug

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1102

image


gordonbrander (approved):
LGTM

PR #1107, opened 2024-02-07
Experiment: score-based classification and routing for prompts #3

gordonbrander:
Yet another riff on https://github.com/subconsciousnetwork/subconscious/pull/1104

## Design

This PR changes the design by making removing the concept of an orchestrator. Instead, we have

- Classifiers
  - May be composed together into a single classifier
- Routes
  - Receives a request struct which contains the input and classifications, and a `process(:)` function.
  - Returns a string or nil.
  - Routes may safely recursively call the parent router using the `process(:)` function.
- Router
  - Has a classifier, which it runs for each request. Classifier may be composed of multiple classifiers.
  - Has an array of routes. Routes are run top-to-bottom for each request.
  - Router exits on the first route match.

This gives routes a high degree of expressivity, since a route may rewrite the input using information from the classifications, and then recurse back into the router. Routes may also call out to specialized sub-routers, allowing us to construct trees of routing, with each router able to recurse on itself. This is similar in principle to many rule-based NLP systems such as AIML or ChatScript that use hierarchy and recursion to pick apart an input and dispatch parts of it to different subsystems, before returning a result.

The previous orchestrator model only had weights to work with to make a choice between results. This approach has much more nuanced control over the result, since the result can be returned from specific and specialized branches within the  tree of routers.

## Concepts

- `PromptClassification`: an individual classification with weight
- `PromptClassifierProtocol`: given an input, produces classifications
- `PromptClassifier`: composes classifiers together
- `PromptRouteRequest`: a request struct containing the context for a given route request, including original input, classifications, and a function to recursively call into the routes router.
- `PromptRouteProtocol`: a route
  - `PromptRoute`: a concrete implementation of `PromptRouteProtocol` taking a closure as a definition
- `PromptRouter`: given an array of routes and a classifier, produces a result with `process(:) -> String?`
  - `PromptRouterRequest`: an ephemeral actor that exists for the duration of a `PromptRouter.process(:)` request. This actor tracks the recursion depth and makes sure requests don't recurse too deeply.
    - Aside: I was pleased to learn that actors do not cost any more than classes
- `PromptService`: configures classifiers, routes, and routers.


bfollington (approved):
LGTM!

Issue #1108, opened 2024-02-07
Selection autocorrect suggestion from keyboard results in extra space character & misaligned cursor


bfollington:
Repoted by a beta user. As the title says, tapping an autocorrect suggestion seemed to insert a space mid-word and left the cursor off-by-one from the end of the word.

Might be an intermittent issue, need to test to confirm.

![image](https://github.com/subconsciousnetwork/subconscious/assets/5009316/e1c66a81-8f83-4a0e-8ef7-a3bfc93d75f0)

Issue #566, opened 2023-04-29
Cursor jumps around while editing a note


cdata:
After entering a local slashlink in a note, I started to type another link two lines down.

As I was typing, the cursor started to jump up a line with each input. Demo video attached.

https://user-images.githubusercontent.com/240083/235327629-5955a638-a13c-43fa-b622-83e1844957d9.MP4

bfollington:
Added context: I have seen this happen but I cannot reproduce it (even using identical markup to the video) on the iOS 16.4 simulator.

We should check for differences between versions as TK2 is still being updated, but also for differences on-device vs simulator.

PR #1109, opened 2024-02-08
Editor performance improvements

bfollington:
While investigating https://github.com/subconsciousnetwork/subconscious/issues/1108 I profiled the `plainEditor()` and found it was extremely expensive on each keystroke. It's incredibly expensive to set the `.text` property on the underlying view and doing so can result in cursor desync. 

I suspect this is also responsible for https://github.com/subconsciousnetwork/subconscious/issues/566.

Setting `.text` frequently causes `updateUIView()` to take a LONG time (up to 400ms in debug mode) which means `isUIViewUpdating` stays true for a long time and we start dropping input events if the user is typing rapidly. This is what is responsible for the autocorrect cursor desync issue linked above and can be replicated by typing very fast into the simulator. It's harder to trigger in release, on-device, but it happens there too.

In brief: there's a pathological update cycle happening that results in setting `.text` twice when we can set it 0 times. I've documented the specifics in the code.

I made two changes:

1. enhance the change detection logic so we can trust `view.text` more often instead of resetting it
1. split the rendering of the new gradient BG into another layer from the editor itself (minor, but helpful) 

The difference is more exaggerated on simulator, but it's night and day.

# Before

https://github.com/subconsciousnetwork/subconscious/assets/5009316/1eec6b50-8abf-4ddb-ab9e-ce0504597587

# After

https://github.com/subconsciousnetwork/subconscious/assets/5009316/39af615e-3802-48da-82a0-af04265b8df2

### DIFF
--- a/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
+++ b/xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
@@ -128,45 +128,45 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
                 )
             )
         }
-        
-        // This allows us to intercept touches on embedded transclude blocks
-        override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
-            guard let touch = touches.first else {
-                super.touchesBegan(touches, with: event)
-                return
-            }
-
-            let tapPoint = touch.location(in: self)
-            
-            guard let textLayoutManager = self.textLayoutManager else {
-                SubtextTextViewRepresentable.logger.warning("Could not access textLayoutManager")
-                super.touchesBegan(touches, with: event)
-                return
-            }
-            
-            guard let textContentStorage = textLayoutManager
-                .textContentManager as? NSTextContentStorage
-            else {
-                SubtextTextViewRepresentable.logger.warning("Could not access textContentStorage")
-                super.touchesBegan(touches, with: event)
-                return
-            }
-            
-            // Did tap a text element?
-            if let textElement = textLayoutManager
-                .textLayoutFragment(for: tapPoint)?
-                .textElement
-            {
-                let content = textContentStorage.attributedString(
-                    for: textElement
-                )
-                
-                // TODO: check for whether this tap should navigate to a link
-                SubtextTextViewRepresentable.logger.debug("Tapped: \(String(describing: content?.string))")
-                // Calling super preserves default behaviour
-                super.touchesBegan(touches, with: event)
-            }
-        }
+//        
+//        // This allows us to intercept touches on embedded transclude blocks
+//        override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+//            guard let touch = touches.first else {
+//                super.touchesBegan(touches, with: event)
+//                return
+//            }
+//
+//            let tapPoint = touch.location(in: self)
+//            
+//            guard let textLayoutManager = self.textLayoutManager else {
+//                SubtextTextViewRepresentable.logger.warning("Could not access textLayoutManager")
+//                super.touchesBegan(touches, with: event)
+//                return
+//            }
+//            
+//            guard let textContentStorage = textLayoutManager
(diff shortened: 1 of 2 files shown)


xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -402,10 +405,30 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
...
+        // updateUIView() called
+        // we set view.text = state.text
+        // now, view.text = "abc"
+        // THEN state.text updates to "abcd"
+        // updateUIVIew() called again
+        // we set view.text = state.text
+        // view.text == "abcd"
+        
+        // so we set view.text TWICE when we could set it zero times
+        // caching the last known rendered value allows us to detect and sidestep this scenario
```
bfollington, line 426: In debugging I can see that action for `.setText` is actually dispatched before `updateUIView()` is called BUT the value stored in `state` lags behind the value passed to `.setText`. 

It's possible there's a solution to this problem at a layer below?

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -196,7 +201,7 @@ struct MemoEditorDetailView: View {
     var highlight: Color? {
         store.state.themeColor?.toHighlightColor()
             ?? store.state.address?.themeColor.toHighlightColor()
-            ?? ThemeColor.a.toColor()
+            ?? ThemeColor.a.toHighlightColor()
```
bfollington, line 204: Small oversight on cursor color here.

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -287,44 +250,6 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
                 )
             )
         }
-        
-        // MARK: - NSTextLayoutManagerDelegate
-                                
-        func textLayoutManager(
```
gordonbrander, line 293: Given we're remove this function, how are we handling what it did before?

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
```diff
@@ -402,10 +324,30 @@ struct SubtextTextViewRepresentable: UIViewRepresentable {
...
+        // updateUIView() called
+        // we set view.text = state.text
+        // now, view.text = "abc"
+        // THEN state.text updates to "abcd"
+        // updateUIVIew() called again
+        // we set view.text = state.text
+        // view.text = "abcd"
+        
+        // so we set view.text TWICE when we could set it zero times
+        // caching the last known rendered value allows us to detect and sidestep this scenario
```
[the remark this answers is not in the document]
gordonbrander: I'm a bit unclear about how this dirty flag approach might play with bi-directional data flow (from user to state, from state to UITextView).

I think the right course of action may be to land this, and then think about ways we might refactor.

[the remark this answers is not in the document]
gordonbrander: We might approach this like we do the block editor:

- UIViewRepresentable gets its own `@ObservedObject` store
- All UITextView events are sent as actions to store, updating state
- UIViewRepresentable subscribes to store state publisher, driving renders via publisher instead of update method
- Check UIView before setting state to ensure they're actually different

This seems to side-step a number of issues in the block editor, including the purple warning, since it guarantees view updates are downstream of state updates, and that those view updates always happen on the main thread.

gordonbrander (approved):

xcode/Subconscious/Shared/Components/Common/SubtextTextViewRepresentable.swift
bfollington: None of this is actually being used, it's all obsolete residue from implementing the TextKit 2 delegate methods. The entire method is a no-op in the current build.

PR #1110, opened 2024-02-09
Narrow trigger condition to re-check provisioning status

bfollington:
The fix I implemented in https://github.com/subconsciousnetwork/subconscious/pull/1094 works but it's triggered in scenarios it shouldn't be. There's no ill effect of running this too often but I'd rather be precise.


gordonbrander (approved):
LGTM

PR #1118, opened 2024-02-19
Skip unloadable likes and unresolvable peers when reading user likes

bfollington:
Liking a note and then deleting it triggered a "Not Found" error on the profile view.

We should skip over unreadable likes rather than throwing.


gordonbrander (approved):
LGTM

PR #1119, opened 2024-02-19
Reduce app startup time

bfollington:
While testing 0.0.18 I noticed the app seemed slow to boot & laggy at times. I profiled it using the App Startup profiling, as well as SwiftUI and good old fashioned CPU profiling, found the main bottlenecks and removed them:

- Limit the number of rows read from the DB to populate the deck
- Remove Sentry profiling integration
    - Turns out we have to pay more money for this and we can profile the app locally just fine for now
- Defer SentryIntegration.start()

For me this seems to cut down launch time by nearly a second in the simulator. Hopefully that carries over to a real device!

### DIFF
--- a/xcode/Subconscious/Shared/Components/AppView.swift
+++ b/xcode/Subconscious/Shared/Components/AppView.swift
@@ -77,6 +77,7 @@ struct AppView: View {
             )
         }
         .onAppear {
+            SentryIntegration.start()
             store.send(.appear)
         }
         // Track changes to scene phase so we know when app gets
--- a/xcode/Subconscious/Shared/Components/Deck/DeckView.swift
+++ b/xcode/Subconscious/Shared/Components/Deck/DeckView.swift
@@ -518,20 +518,15 @@ struct DeckModel: ModelProtocol {
                 try? await Task.sleep(for: .seconds(Duration.loading))
                 
                 let us = try await environment.noosphere.identity()
-                let recent = try environment.database.listFeed(owner: us)
+                let recent = try environment.database.listAll(owner: us, limit: 5)
                 let likes = try await environment.userLikes.readOurLikes()
                 
                 var initialDraw = Array(recent
-                    .prefix(10) // take the 10 most recent posts
+                    .prefix(5) // take the 10 most recent posts
                     .shuffled() // shuffle
-                    .prefix(3)) // take 3
-                
-                // Draw 2 random cards to keep it surprising
-                for _ in 0..<2 {
-                    guard let entry = environment.database.readRandomEntry(owner: us) else {
-                        continue
-                    }
-                    
+                    .prefix(2))
+            
+                if let entry = environment.database.readRandomEntry(owner: us) {
                     initialDraw.append(entry)
                 }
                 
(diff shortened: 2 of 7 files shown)


xcode/Subconscious/Shared/Components/AppView.swift
```diff
@@ -77,6 +77,7 @@ struct AppView: View {
             )
         }
         .onAppear {
+            SentryIntegration.start()
```
bfollington, line 80: We can reduce time to first paint by defering the SentryIntegration call, but we still want to catch any errors that are thrown via `AppAction.appear`

xcode/Subconscious/Shared/Library/Sentry.swift
```diff
@@ -48,9 +48,6 @@ extension SentryIntegration {
             // per https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization this is fine to be public, unless it's abused (e.g. someone sending us /extra/ errors.
             options.dsn = "https://72ea1a54aeb04f60880d75fcffe705ed@o4505393671569408.ingest.sentry.io/4505393756438528"
             options.environment = Config.default.debug ? "development" : "production"
-            options.tracesSampleRate = 1.0
-            options.profilesSampleRate = 1.0
-            options.attachViewHierarchy = true
```
bfollington, line 53: Turns out this has a significant impact on startup & performance and we're not even getting the benefit without paying more money (see PR description).

xcode/Subconscious/Shared/Components/Deck/DeckView.swift
```diff
@@ -518,20 +518,15 @@ struct DeckModel: ModelProtocol {
                 try? await Task.sleep(for: .seconds(Duration.loading))
                 
                 let us = try await environment.noosphere.identity()
-                let recent = try environment.database.listFeed(owner: us)
+                let recent = try environment.database.listAll(owner: us, limit: 5)
                 let likes = try await environment.userLikes.readOurLikes()
                 
                 var initialDraw = Array(recent
-                    .prefix(10) // take the 10 most recent posts
+                    .prefix(5) // take the 10 most recent posts
```
gordonbrander, line 525: Fix comment

gordonbrander: Does this really make a large perf difference?

xcode/Subconscious/Shared/Library/Sentry.swift
gordonbrander: Good to know. If we bring it back, let's make sure to disable it unless in development env.

xcode/Subconscious/Shared/Components/AppView.swift
gordonbrander: We could defer everything that's happening in appear. For example, we could queue an fx in appear for an action called "start", and move everything downstream of that *except* the Sentry setup. This should cause the downstream actions to run on the next tick.

Noting. Would want to defer to a future PR. Worth profiling if there's a difference as well.

gordonbrander (approved):
One nit, one question.

xcode/Subconscious/Shared/Components/Deck/DeckView.swift
bfollington: No, the main improvement came from reducing the number of entries returned by `listAll`. Before it could `compactMap` up to 1000 rows (which could take hundreds of ms!!?) to only use `10` of them. I deemed that `5` was sufficient for the very limited use we make of this data.

PR #1120, opened 2024-02-19
Visual tweaks after playtesting

bfollington:
- Reduce card height to account for prompt header
    - Cards grew in height when we introduced prompts and they feel too tall to me now
- Make whitespace around like button consistent between viewer and editor
- Add user profile header to the viewer
    - Optional addition, I felt myself craving it while browsing
- Comments inherit the note color
    - Reduces noise + creates a sense of visual relatedness


gordonbrander (approved):
LGTM

PR #1121, opened 2024-02-19
Suspend re-rendering to UITextView.text if it causing the change in the first place

bfollington:
I ran into some issues testing on-device even with the previous fix (https://github.com/subconsciousnetwork/subconscious/pull/1109) and decided to change approach to something far simpler.

In my testing this behaves correctly on-device with system autocomplete, swipe typing, inserting markup, copy, paste, cursor manipulation etc.


gordonbrander (approved):
Smart. LGTM.

PR #1114, opened 2024-02-13
Implement "comments" under notes

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1085

- Implement deterministic hash for color across executions
- Add comments to editor and viewer
- Tapping a comment will open a new editor instance preloaded with the comment + address

PR #1122, opened 2024-02-20
Fix unstable derived note colors

bfollington:
Note colors can change between app restarts. Apparently I messed this up again even after trying to fix it in https://github.com/subconsciousnetwork/subconscious/pull/1114

The mistake was subtle, so I'll give myself a pass.

### DIFF
--- a/xcode/Subconscious/Shared/Components/BacklinksView.swift
+++ b/xcode/Subconscious/Shared/Components/BacklinksView.swift
@@ -30,7 +30,7 @@ struct BacklinksView: View {
                             .buttonStyle(RelatedNoteButtonStyle(color: entry.color))
                         }
                         .tint(
-                            entry.headers.themeColor?.toHighlightColor()
+                            entry.highlightColor
                         )
                     }
                 }
--- a/xcode/Subconscious/Shared/Components/Deck/DeckTheme.swift
+++ b/xcode/Subconscious/Shared/Components/Deck/DeckTheme.swift
@@ -137,7 +137,7 @@ extension String {
         // execution
         let hash = self.utf8.reduce(0) { $0 + Int($1) }
         
-        return colors[abs(hash.hashValue) % colors.count]
+        return colors[abs(hash) % colors.count]
     }
 }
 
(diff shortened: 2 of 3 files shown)


xcode/Subconscious/Shared/Components/Deck/DeckTheme.swift
```diff
@@ -137,7 +137,7 @@ extension String {
         // execution
         let hash = self.utf8.reduce(0) { $0 + Int($1) }
         
-        return colors[abs(hash.hashValue) % colors.count]
+        return colors[abs(hash) % colors.count]
```
bfollington, line 140: Calling `.hashValue` resulted in Swift hashing the `Int` using non-deterministic hashing 🙃 

We can just use my "cheap deterministic hash" (summing char codes) directly.

xcode/Subconscious/Shared/Components/BacklinksView.swift
```diff
@@ -30,7 +30,7 @@ struct BacklinksView: View {
                             .buttonStyle(RelatedNoteButtonStyle(color: entry.color))
                         }
                         .tint(
-                            entry.headers.themeColor?.toHighlightColor()
+                            entry.highlightColor
```
gordonbrander, line 33: I'm fuzzy on the control flow here, but it seems like we're bypassing the header-set color?

bfollington: These computed properties check the headers first before falling back to deriving the color from the address, so the old behaviour is preserved (and extended):

```swift
extension EntryStub {
    var color: Color {
        self.headers.themeColor?.toColor()
            ?? self.address.themeColor.toColor()
    }
    
    var highlightColor: Color {
        self.headers.themeColor?.toHighlightColor()
            ?? self.address.themeColor.toHighlightColor()
    }
}
```

gordonbrander: Nice!

gordonbrander (approved):
LGTM

PR #1123, opened 2024-02-21
Improve indexing performance

bfollington:
- Batch reads and database calls in tight loops
- Reduce re-parsing cost in MemoRecord init
- Index peers in TaskGroup
- Reduce re-parsing again
- Adjust task priority

Speeds up indexing between 4x and 6x.

Before

image

After

image

### DIFF
--- a/xcode/Subconscious/Shared/Parsers/Tape.swift
+++ b/xcode/Subconscious/Shared/Parsers/Tape.swift
@@ -84,10 +84,23 @@ struct Tape {
     /// Peek forward, and consume if substrings match
     mutating func consumeMatch(_ subsequence: Substring) -> Bool {
         if let endIndex = offset(by: subsequence.count) {
-            if rest[currentIndex..<endIndex] == subsequence {
-                self.currentIndex = endIndex
-                return true
-            }
+            var idxA = currentIndex
+            var idxB = subsequence.startIndex
+            
+            // iterating through both indices is MUCH faster than constructing and comparing
+            // a substring slice, especially since we get early exist on the first mismatched
+            // character.
+            repeat {
+                if rest[idxA] != subsequence[idxB] {
+                    return false
+                }
+                
+                idxA = rest.index(after: idxA)
+                idxB = subsequence.index(after: idxB)
+            } while idxB < subsequence.endIndex
+            
+            self.currentIndex = endIndex
+            return true
         }
         return false
     }
--- a/xcode/Subconscious/Shared/Services/DataService.swift
+++ b/xcode/Subconscious/Shared/Services/DataService.swift
@@ -292,48 +292,62 @@ actor DataService {
     }
     
     func indexPeers(petnames: [Petname]) async -> [PeerIndexResult] {
-        var results: [PeerIndexResult] = []
-        
-        for petname in petnames {
-            logger.log(
-                "Indexing peer",
-                metadata: [
-                    "petname": petname.description
-                ]
-            )
+        let results = await withTaskGroup(of: PeerIndexResult.self) { group in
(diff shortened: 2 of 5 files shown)


xcode/Subconscious/Shared/Parsers/Tape.swift
```diff
@@ -84,10 +84,23 @@ struct Tape {
...
+                if rest[idxA] != subsequence[idxB] {
+                    return false
+                }
+                
+                idxA = rest.index(after: idxA)
+                idxB = subsequence.index(after: idxB)
+            } while idxB < subsequence.endIndex
+            
+            self.currentIndex = endIndex
+            return true
```
bfollington, line 103: I found this one speedup to be equally effective vs. storing an array of indices (a la https://maximeremenko.com/string-random-access)

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -154,42 +154,67 @@ actor DataService {
         try database.savepoint(savepoint)
         
         do {
+            var toUpdate: [(Slug, Memo)] = []
+            var toRemove: [Slug] = []
```
bfollington, line 158: First pass: fetch all changes from Noosphere
Second pass: apply all changes to index

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -154,42 +154,67 @@ actor DataService {
...
+                    }
+                }
+                
+                // Collect the results
+                for try await record in group {
+                    records.append(record)
+                }
+                
+                return records
+            }
```
bfollington, line 190: We can parallelize the construction of MemoRecord to massively speed up Subtext parsing

xcode/Subconscious/Shared/Parsers/Tape.swift
gordonbrander: Interesting!

There may still be an advantage to storing indices in an array once at Tape creation *IFF* it has the same performance characteristics.

```swift
this.indices = Array(base.indices)
```

The reason is that `consumeMatch` is just one place where we want to do this kind of operation. I think we use it for most of the parser, but there are other cases where we might want to use `peek` or other helpers.

We also use `Substring` and `Substring.Index` in a bookkeeping role, for example, to keep track of the `rest` position and `startIndex`. I'm wondering if we could get a performance improvement by storing indices in an array, and thus avoiding the manipulation of `Substring` indexes altogether.

- `rest()` could become a `Range<Int>` that indexes into `indices`
- `advance()` coud increment an `Int` counter by one
- Etc

gordonbrander: Let's leave that for a future exploration.

xcode/Subconscious/Shared/Services/DataService.swift
```diff
@@ -154,42 +154,67 @@ actor DataService {
         try database.savepoint(savepoint)
         
         do {
+            var toUpdate: [(Slug, Memo)] = []
+            var toRemove: [Slug] = []
             // Memos
             for change in changes {
                 let slashlink = Slashlink(slug: change)
                 // If memo does exist, write it to database.
```
gordonbrander, line 162: Comment out of date

xcode/Subconscious/Shared/Services/DataService.swift
gordonbrander: Let's leave a comment to this effect for our future selves.

gordonbrander (approved):
Just one small request to update comment. Otherwise LGTM!

Great work.

Issue #1100, opened 2024-02-05
Intermittent "Could not save note" error when navigating back


bfollington:
Stems from trying to save with no changes, which may be indicative of two attempts to save a note.

e.g.
```
Failed to save entry: /fungi
Foreign Error: No changes to save
```

Issue #1065, opened 2024-01-09
`autosave` polling action is never cancelled after editor is disposed


bfollington:
After opening and closing several editors and then navigating to root on all tabs I see (many) `autosave` actions being generated indefinitely.

```
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
Action: autosave
...100s more
```

```swift
/// `MemoEditorDetail.swift`
static func start(
        state: MemoEditorDetailModel,
        environment: AppEnvironment
    ) -> Update<MemoEditorDetailModel> {
        let pollFx = AppEnvironment.poll(every: Config.default.pollingInterval)
            .map({ _ in MemoEditorDetailAction.autosave })
            .eraseToAnyPublisher()
        return Update(state: state, fx: pollFx)
    }
```

This seems likely the culprit.

gordonbrander:
@bfollington can you confirm the version of ObservableStore you see this happening on?


I fixed a similar issue in https://github.com/subconsciousnetwork/ObservableStore/pull/45

bfollington:
@gordonbrander Confirmed that we're using 0.6.1 (the latest release) on main and I can still reproduce it on simulator.

Trying to identify the simplest reproduction steps, I can get this to happen by opening and closing the editor multiple times but changing the note's link or audience seems to trigger it more reliably.

PR #1128, opened 2024-03-04
Prevent runaway autosave polling + concurrent save attempts

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1065
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1100 (I think)

- Do not actually attempt to save a note unless it has been modified
- Cancel polling action once editor closes

### DIFF
--- a/xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
+++ b/xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
@@ -106,6 +106,9 @@ struct MemoEditorDetailView: View {
             store.send(MemoEditorDetailAction.appear(description))
             blockEditorStore.send(.appear(description))
         }
+        .onDisappear {
+            store.send(MemoEditorDetailAction.disappear)
+        }
         // Track changes to scene phase so we know when app gets
         // foregrounded/backgrounded.
         // See https://developer.apple.com/documentation/swiftui/scenephase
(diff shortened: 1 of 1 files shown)


xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -1187,10 +1204,47 @@ struct MemoEditorDetailModel: ModelProtocol {
         state: MemoEditorDetailModel,
         environment: AppEnvironment
     ) -> Update<MemoEditorDetailModel> {
-        let pollFx = AppEnvironment.poll(every: Config.default.pollingInterval)
-            .map({ _ in MemoEditorDetailAction.autosave })
```
bfollington, line 1191: This task would outlive the view that created it. This _may_ point to an underlying issue in Observable store as I would expect this `fx` to be cancelled.

xcode/Subconscious/Shared/Components/Detail/MemoEditorDetail.swift
```diff
@@ -1806,8 +1862,8 @@ struct MemoEditorDetailModel: ModelProtocol {
         environment: AppEnvironment,
         entry: MemoEntry
     ) -> Update<MemoEditorDetailModel> {
-        // If editor dom is already saved, noop
-        guard state.saveState != .saved else {
+        // If editor dom is already saved or saving, noop
+        guard state.saveState == .unsaved else {
```
bfollington, line 1866: Guard against concurrent saves.

gordonbrander (approved):
LGTM. I prefer this stepwise polling approach, since it allows us to fold into the state at each step.

PR #1130, opened 2024-03-12
Introduce macOS 13 runner with iOS 17

bfollington:
Run tests on iOS 17 in CI.


gordonbrander (approved):
LGTM

gordonbrander:
@bfollington LGTM. One question: you mentioned this test runner seems to take much longer than the v16 one. Is this something that should cause concern, or just inconvenient?

bfollington:
@gordonbrander both.... I don't think there's anything we can do but it's going to be frustrating for development. It's tolerable though.

PR #1137, opened 2024-03-19
Revert to a direct store binding for AppTab

bfollington:


gordonbrander (approved):
LGTM! Great catch @bfollington

Issue #1135, opened 2024-03-19
Tapping draft in deck opens public link 


bfollington:
I thought I fixed this but it's still happening on Testflight.

Reproduction:

1. Select Deck tab
2. Swipe through cards until you see a _draft_ note from yourself
3. Tap the note, notice a long load time followed by "Not Found"

We're somehow trying to load the public version of the address when launching from the card.

PR #1140, opened 2024-03-19
Fix handling of draft addresses when loading from SQLite

bfollington:
Fixes https://github.com/subconsciousnetwork/subconscious/issues/1135

Drafts are only very rarely being drawn in the deck view but when it _does_ happen we were calculating the address incorrectly.

I relaxed the (old) constraint that preventing drafts from showing up in the deck view most of the time (I think you should be able to see these like any other note, otherwise they will never resurface) and clarified the address resolution when loading from SQLite.


gordonbrander (approved):
LGTM, but can we add a test?

PR #1142, opened 2024-04-03
Configurable sort order for notebook

bfollington:
Sort / filter the notebook view. I added a few new options that occurred to me as easy and useful.

https://github.com/subconsciousnetwork/subconscious/assets/5009316/09d69caf-ff63-44ae-a75d-5fdb9598ee9e


xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
```diff
@@ -40,9 +40,7 @@ struct EntryListView: View {
                     // invisible marker to scroll back to
                     EmptyView().id(Self.resetScrollTargetId)
                     
-                    ForEach(entries.indices, id: \.self) { idx in
-                        let entry = entries[idx]
-                        
+                    ForEach(entries, id: \.self) { entry in
```
bfollington, line 43: We didn't need the index anymore, apparently

xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
```diff
@@ -134,6 +132,7 @@ struct EntryListView: View {
                                 )
                             }
                         }
+                        .id(entry.address)
```
bfollington, line 135: This fixes the animations

xcode/Subconscious/Shared/Components/Common/CountChip.swift
```diff
@@ -22,6 +22,7 @@ struct CountChip: View, Equatable {
 
     var body: some View {
         Text(Self.asText(count: count))
+            .contentTransition(.numericText())
```
gordonbrander, line 25: TIL

xcode/Subconscious/Shared/Components/Common/EntryList/EntryListView.swift
gordonbrander: Entries should be Identifiable, so we shouldn't need the second argument at all. That would be preferable, since `\.self` will consider value equality, and we would prefer to compare ID equality.

gordonbrander (approved):
LGTM. One comment.

